PR #64, opened 2025-05-05
Update stats-view.css

CRIMSONHydra:
Better displays original url component when the url size is big.
**Before:**
-Vertical scroll
-less width
IMG-20250505-WA0012
IMG-20250505-WA0026
Notice how you dont even realise that it can be scrolled down to view the remaining url

**Now:**
-Horizontal scroll
-takes up more width of the container
IMG-20250505-WA0015
IMG-20250505-WA0013

## Summary by Sourcery

Improve the display of long URLs in the stats view by modifying the CSS to provide better visibility and scrolling

Enhancements:
- Modify URL display to use horizontal scrolling instead of vertical scrolling
- Increase the width of the URL container to 70% of available space
- Allow more flexible height for long URLs

### DIFF
--- a/static/css/stats-view.css
+++ b/static/css/stats-view.css
@@ -136,10 +136,11 @@ select:hover {
 }
 
 #url_long_code {
-    max-width: 220px;
+    max-width: 70%;
     display: inline-flex;
-    overflow-x: scroll;
-    height: 28px;
+    overflow-x: auto;
+    min-height: 28px;
+    max-height: 50px;
 }
 
 #url_long_code::-webkit-scrollbar {
--- a/static/css/stats-view.css
+++ b/static/css/stats-view.css
@@ -138,9 +138,11 @@ select:hover {
 #url_long_code {
     max-width: 70%;
     display: inline-flex;
-    overflow-x: auto;
-    min-height: 28px;
-    max-height: 50px;
+    overflow-x: hidden;
+    overflow-y: auto;
+    max-height: 30px;
+    white-space: normal;
+    word-break: break-all;
 }
 
 #url_long_code::-webkit-scrollbar {
@@ -469,6 +471,10 @@ h3 {
     #analysisJson,
     #botsJson {
         height: 250px;
+    } */
+
+    /* #url_long_code {
+        max-width: 50%;
     } */
 
     #url_long_input {
--- a/templates/stats_view.html
+++ b/templates/stats_view.html
@@ -33,7 +33,7 @@
 
     <link rel="stylesheet" href="{{ url_for('static', filename='css/base.css') }}">
     <link rel="stylesheet" href="{{ url_for('static', filename='css/header.css') }}?v=3">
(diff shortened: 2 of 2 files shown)


CRIMSONHydra:
min-height/max-height is not required for horizontal scrolling but I have added it incase you want to retain vertical scrolling behavior, 50px max height doesnt perfectly align with the text and gives you a feel that it can be scrolled further if required

> * Consider if the added `min-height`/`max-height` is necessary when using `overflow-x: auto` primarily for horizontal scrolling.

Zingzy:
Hello @CRIMSONHydra,

Thanks for the PR! While reviewing this it, I found a couple of edge cases issues:

1. When we have this super long URL without breaks (e.g,  `http://www.reallylong.link/rll/7u6g/k_HLxGpgPJW6aPZFkqSRRc/1Iqzoqi0jqID10hsF21yxWwT/xzTekmdAn4X3JE0dPi97g5ZhogAwUDO/lUf/...[1850chars]...6gPMr8AZkbNAGhGRBDFCl`), the UI breaks 

image

2. When the URL is super long but have multiple breaks like `-`, `+`, etc. the long URL gets wrapped and the box height expands upto `50px` which should not happen

image

Would love your thoughts on how we can handle these cases better.

**P.S**. This has been a small yet annoying bug for quite some time - great job spotting and working on it!

CRIMSONHydra:
>2. When the URL is super long but have multiple breaks like -, +, etc. the long URL gets wrapped and the box height expands upto 50px which should not happen

Adding 
`
white-space: nowrap;
`
can fix this issue. Additionally `max-height` is no longer neccassary.

as for
>1. When we have this super long URL without breaks (e.g, http://www.reallylong.link/rll/7u6g/k_HLxGpgPJW6aPZFkqSRRc/1Iqzoqi0jqID10hsF21yxWwT/xzTekmdAn4X3JE0dPi97g5ZhogAwUDO/lUf/...[1850chars]...6gPMr8AZkbNAGhGRBDFCl), the UI breaks


The UI actually works on touchpad on laptop, you can scroll _horizontally_. No idea how to add a visual representation of a scrollbar here in this case or make it work with a mouse. Havent tested it on mobile touch devices either.
Maybe should just implement a vertical scroll instead? Wont look as clean though.

The main point how ever is that most domain names atleast completely fit in the text field right now and the overflown part might be sub domains which arent that relevant in this context I think.

Would love to hear your thoughts on how to proceed. This is a learning experience for me as I am not so experienced with web dev but was just using your site and found this bug

Zingzy:
> Adding
> `white-space: nowrap;`
> can fix this issue. Additionally `max-height` is no longer neccassary

That makes sense. For the first issue, I think we could also try using `width: 30dvw` on larger screens and `60dvw` on smaller ones instead of a fixed percentage like `70%`, which should help avoid the box stretching too wide across the page.

For now, allowing vertical scrolling seems like a practical workaround — it's not ideal visually, but it keeps things functional :)

Zingzy:
ok things look fine to me, I might make some minor adjustments here and there but for now, I am merging this PR

Issue #133, opened 2026-03-29
anonymous users can't reclaim their links


revanthlol:
**Description:**
A clear and concise description of the problem:

Currently, anonymous users who shorten URLs cannot claim ownership of those URLs after signing up or logging in. This leads to confusion when they try to manage or delete links they created anonymously.

---

**Steps to Reproduce:**

1. Shorten a URL while not logged in.
2. Check `recentURLs` in localStorage.
3. Sign up or log in.
4. The previously shortened URL remains unclaimed and cannot be managed.

---

**Expected Behavior:**

* Anonymous users should be able to claim ownership of all URLs they created before signing up.
* The system should handle this automatically after login and optionally allow manual claiming.

---

**Proposed Solution:**

* Generate a one-time `manage_token` when an anonymous user shortens a URL.
* Store the token hashed in the database, raw token returned once in the API response.
* Frontend saves token in `recentURLs`.
* After login/registration, sweep all stored tokens and claim corresponding URLs via a POST `/api/v1/claim` endpoint.
* Show “Unclaimed” badges and a “Claim All” button for unclaimed URLs.

**Security considerations:**

* Tokens are hashed before storage (SHA-256).
* Atomic claim operation to prevent race conditions.
* Tokens cannot be reused or retrieved from API after claim.

---

**Alternatives Considered:**

* Ignoring unclaimed URLs (bad UX).
* Manual claim via backend (less user-friendly).

---

**Additional Context / Screenshots:**

* Optional: screenshots of current unclaimed URLs in UI.
* Optional: localStorage structure for `recentURLs`.

Zingzy:
Picked up in #277, which ships the claim flow end to end: anonymous creates return a one-time claim token, and an authenticated endpoint attaches those links to an account later, stats included. The bearer-token approach you proposed in #134 was the right call, and the oracle-resistance note from it lives on in the implementation, credited to you.

PR #134, opened 2026-03-29
Add anonymous URL ownership claim via manage_token

revanthlol:
## What does this PR do?

Fixes the UX issue where anonymous users couldn't manage URLs they created 
before signing up. Closes #<open an issue first and put the number here>

## How it works

1. When an anonymous user shortens a URL, the backend generates a cryptographically 
   random `manage_token`, stores its SHA-256 hash in the URL document, and returns 
   the raw token **once** in the API response.

2. The frontend stores the token inside the existing `recentURLs` localStorage 
   structure as `{ alias, manage_token }`.

3. On the result page, a banner shows the token with a copy button and a 
   "Claim this link" option that opens the auth modal.

4. On the homepage, unclaimed URLs show an **Unclaimed** badge and a 
   **"Sign in to claim all your links"** button.

5. After successful login or registration, `sweepAndClaimTokens()` automatically 
   claims all stored tokens in one pass before redirecting.

6. The `/api/v1/claim` endpoint is atomic — a single `update_one` with all 
   conditions in the filter (owner is anonymous + token hash matches). Token is 
   nulled immediately after claim, making replay impossible.

## Security

- Token is never stored in plaintext — only SHA-256 hash persisted in DB
- Claim is a single atomic MongoDB operation, no read-then-write race
- Wrong token, already claimed, and not found all return 403 (no oracle)
- `manage_token` is excluded from all other API responses

## Files changed

**Backend**
- `schemas/models/url.py` — added `manage_token: Optional[str]` to `UrlV2Doc`
- `repositories/url_repository.py` — added `claim_by_manage_token()`
- `services/url_service.py` — token generation in `create()`, new `claim_url()`
- `routes/api_v1/claim.py` — new `POST /api/v1/claim` endpoint
- `routes/api_v1/shorten.py` — unpack tuple return, expose token in response
- `schemas/dto/responses/url.py` — added `manage_token` field to `UrlResponse`

**Frontend**
- `static/js/index-script.js` — store token in `recentURLs` on shorten
- `static/js/index-recent-links.js` — Unclaimed badge + Claim All button
- `static/js/result-script.js` — token banner, claimNow() → auth modal flow
- `static/js/auth.js` — `sweepAndClaimTokens()` on login/register success
- `static/css/result.css` — token banner styles
- `static/css/index.css` — badge + claim button styles

**Tests**
- `tests/integration/api_v1/test_claim.py` — 6 new integration tests
- `tests/integration/api_v1/test_shorten.py` — updated mocks for tuple return
closes issue #133

## Summary by Sourcery

Add support for anonymous URL ownership claims using one-time manage tokens surfaced in shorten responses and claimable after authentication.

New Features:
- Expose a one-time manage_token in shorten API responses for anonymously created URLs so they can later be claimed.
- Introduce a POST /api/v1/claim endpoint to transfer anonymous URLs to the authenticated user using a manage token.
- Add frontend flows and UI to surface manage tokens, mark unclaimed recent links, and trigger a bulk claim after sign-in.

Bug Fixes:
- Resolve the inability for users to manage URLs they created anonymously before signing up by enabling post-hoc ownership claims.

Enhancements:
- Extend URL persistence and service layers to store hashed manage tokens and support atomic claims in the repository.
- Improve recent-links local storage format to carry alias metadata and support backward compatibility with the previous string-only format.

Tests:
- Add integration tests covering the claim endpoint behavior and the presence or absence of manage tokens in shorten responses.
- Update shorten integration tests to reflect the new create return signature including the manage token.

## Summary by CodeRabbit

* **New Features**
  * Claim anonymous URLs to your account using one-time management tokens
  * Anonymous shorten responses now return a one-time manage token
  * One-click batch claiming of recent unclaimed links

* **Style**
  * Added UI components and banners for token display, copy, and claim flows

* **Tests**
  * New and updated integration/unit tests covering claiming and token behavior

* **Chores**
  * Added Docker containerization support


Zingzy:
@revanthlol will come back to this soon!

repositories/url_repository.py
```diff
@@ -234,3 +235,39 @@ async def check_stats_privacy(self, alias: str) -> dict:
...
+        """
+        try:
+            result = await self._col.update_one(
+                {
+                    "alias": alias,
+                    "owner_id": ANONYMOUS_OWNER_ID,
+                    "manage_token": token_hash,
+                },
+                {
+                    "$set": {
```
Zingzy, line 258: 🟡 **[MEDIUM] Bug**

Claiming an anonymous URL updates the owner and clears the token, but it leaves `private_stats` untouched. In this PR, anonymous URLs are created with `private_stats=None`, so a claimed URL can remain in that invalid/uninitialized state after ownership transfer.

**Why this matters in production:** A user may successfully claim a link but still have inconsistent privacy settings on that URL. In production this can lead to claimed links behaving as public when the rest of the system expects owned links to have an explicit boolean privacy value.

**Suggested fix:**
Include `private_stats` normalization in the same atomic update used for claiming, e.g. set it to `False` when it is `None` as part of the ownership transfer so claimed documents satisfy the expected boolean invariant.

services/url_service.py
```diff
@@ -250,6 +252,13 @@ async def create(
             "last_click": None,
         }
 
+        raw_token: Optional[str] = None
```
Zingzy, line 255: 🟡 **[MEDIUM] Bug**

Anonymous URLs are now created with `manage_token`, but the service still returns `UrlV2Doc.from_mongo(doc)` where `doc` contains the hashed token. That means the internal model returned from `create()` now carries sensitive token-hash state that previously did not exist, increasing the chance of accidental logging/serialization by future callers.

**Why this matters in production:** If another caller later serializes or logs the returned `UrlV2Doc`, the hashed claim token could leak into logs or APIs. Even though it's hashed, exposing credential material broadens the attack surface and violates the PR's goal of keeping the token out of normal responses.

**Suggested fix:**
Before constructing `UrlV2Doc`, remove `manage_token` from the in-memory `doc` or explicitly set it to `None` in the returned model, while still persisting the hashed value to MongoDB. For example, build the DB payload separately from the response/model payload.

static/js/auth.js
```diff
@@ -461,4 +498,31 @@ function handlePasswordInput(password) {
...
+async function sweepAndClaimTokens() {
+    let list = [];
+    try { list = JSON.parse(localStorage.getItem('recentURLs')) || []; } catch (_) { return; }
+
+    const unclaimed = list.filter(item => typeof item === 'object' && item.manage_token);
+    if (unclaimed.length === 0) return;
+
+    for (const item of unclaimed) {
+        try {
+            const res = await fetch('/api/v1/claim', {
```
Zingzy, line 510: 🟠 **[HIGH] Bug**

`sweepAndClaimTokens()` uses a plain `fetch()` call instead of `authFetch()`, so if the freshly issued access token is expired or missing but a refresh cookie is still valid, every claim request will fail with 401 and the stored manage tokens will never be claimed during this login flow.

**Why this matters in production:** Users can successfully sign in but still lose the automatic claim behavior for their anonymous links when their access token needs a refresh. In production this will look like claiming is flaky or broken, leaving links unowned until the user retries manually.

**Suggested fix:**
Use `authFetch('/api/v1/claim', ...)` here instead of `fetch(...)`, or add equivalent 401-refresh-retry handling before deciding whether to clear or retain each token.

static/js/index-recent-links.js
```diff
@@ -70,32 +70,99 @@ function renderRecentURLs() {
...
 
 document.addEventListener('DOMContentLoaded', renderRecentURLs);
 
 // Re-render when authentication state changes
-document.addEventListener('auth:init', renderRecentURLs);
\ No newline at end of file
+document.addEventListener('auth:init', renderRecentURLs);
+
+function claimAllAnonymousURLs() {
+    sessionStorage.setItem('spoo_claim_redirect', window.location.href);
```
Zingzy, line 162: 🟡 **[MEDIUM] Bug**

When `openAuthModal` is not available, this fallback redirects to `/` without preserving any signal that the auth modal should open there. On pages where `auth.js` is not loaded, clicking the new claim button just lands the user on the homepage with no claim flow started.

**Why this matters in production:** Users can click "Sign in to claim all your links" and end up on the homepage with nothing happening. That makes the new claim flow appear broken and can leave anonymous links permanently unclaimed unless the user figures out the next step manually.

**Suggested fix:**
In the fallback branch, redirect to a URL/state that the homepage can use to auto-open the auth modal (for example, set a dedicated sessionStorage flag like `spoo_open_auth_modal` before navigating, or redirect with a query parameter that homepage JS consumes).

static/js/result-script.js
```diff
@@ -54,4 +54,52 @@ var copyButton = document.querySelector(".copy-button");
...
     copyShortUrl();
-});
\ No newline at end of file
+});
+
+// ── Manage Token (anonymous URL claim) ────────────────────────────────────
+
+(function initManageTokenBanner() {
+    // Extract short code from current URL: /result/<alias>
+    const alias = window.location.pathname.split('/').pop();
```
Zingzy, line 63: 🟡 **[MEDIUM] Bug**

The alias is read directly from `window.location.pathname`, but the app navigates to the result page using `encodeURIComponent(data.alias)`. For aliases containing spaces, emoji, or other encoded characters, this value will not match the raw `alias` stored in `recentURLs`, so the manage-token banner never appears.

**Why this matters in production:** Users with non-URL-safe aliases will not see their one-time claim token on the result page. In production, that means some anonymous links cannot be claimed later even though the backend generated a valid token.

**Suggested fix:**
Decode the path segment before comparing it to localStorage entries, e.g. `const alias = decodeURIComponent(window.location.pathname.split('/').pop() || '');`, and apply the same normalization anywhere else this page derives the alias from the URL.

static/js/result-script.js
```diff
@@ -54,4 +54,52 @@ var copyButton = document.querySelector(".copy-button");
...
+
+async function claimNow() {
+    // Redirect to home with a flag so the auth modal opens.
+    // sweepAndClaimTokens() in auth.js handles the actual claim after login.
+    const alias = window.location.pathname.split('/').pop();
+    if (alias) {
+        // Store current result page so we can redirect back after login
+        sessionStorage.setItem('spoo_claim_redirect', window.location.href);
+    }
+    if (typeof openAuthModal === 'function') {
```
Zingzy, line 100: 🟠 **[HIGH] Bug**

`claimNow()` tries to open the auth modal only if `openAuthModal` exists on the current page; otherwise it redirects to `/` without setting any flag that tells the homepage to actually open the auth modal. On the result page, where `auth.js` is typically not loaded, clicking "Claim this link" will just send the user home and stop the intended claim flow.

**Why this matters in production:** In production, users can click the claim CTA and end up on the homepage with no login prompt, making the feature appear broken. Many anonymous users will abandon the claim flow and lose the chance to attach their links to their account.

**Suggested fix:**
When `openAuthModal` is unavailable, redirect to `/` with an explicit signal the homepage can consume, such as a query param or `sessionStorage` flag (for example `sessionStorage.setItem('spoo_open_auth_modal', '1')`), and update the homepage/auth bootstrap to automatically open the modal when that flag is present.

routes/api_v1/claim.py
```diff
@@ -0,0 +1,65 @@
...
+    summary="Claim Anonymous URL",
+)
+@limiter.limit(_claim_limit, key_func=_claim_key)
+async def claim_url(
+    request: Request,
+    body: ClaimUrlRequest,
+    user: CurrentUser = Depends(require_auth),
+    url_service: UrlService = Depends(get_url_service),
+) -> ClaimUrlResponse:
+    """
```
Zingzy, line 47: 🔵 **[LOW] Maintainability**

The endpoint docstring documents 409 and 404 responses, but the implementation intentionally collapses all claim failures into 403. This makes the generated API documentation misleading for clients integrating with the new endpoint.

**Why this matters in production:** API consumers may build incorrect error handling based on the published docs and misinterpret normal claim failures as unexpected behavior. That can lead to broken frontend flows or unnecessary support/debugging work in production.

**Suggested fix:**
Update the docstring to state that any failed claim attempt returns 403, including wrong token, already-claimed URLs, and missing URLs.

static/js/index-script.js
```diff
@@ -83,10 +83,17 @@ document.addEventListener('DOMContentLoaded', () => {
             const key = 'recentURLs';
             let list = [];
             try { list = JSON.parse(localStorage.getItem(key)) || []; } catch (_) { list = []; }
-            list.unshift(data.alias);
-            list = Array.from(new Set(list)).slice(0, 3);
+
+            // Remove any existing entry for this alias then prepend
+            list = list.filter(item => (typeof item === 'string' ? item : item.alias) !== data.alias);
+            list.unshift({
```
Zingzy, line 89: 🟡 **[MEDIUM] Bug**

The shorten flow now always rewrites `recentURLs` entries into the new object shape and then truncates to the latest 3 items. If a user already has 3 legacy string entries in localStorage and shortens one more URL, all older entries are converted out by slicing, so any unclaimed tokens that were previously stored only on matching object entries can be dropped during mixed-format transitions. This makes the claim flow fragile across upgrades because localStorage migration is implicit and lossy.

**Why this matters in production:** Users upgrading from an older frontend state can lose claimability metadata for recent anonymous links, preventing them from claiming URLs after signing in.

**Suggested fix:**
Perform an explicit migration step before truncation: normalize all existing entries to a consistent object shape, preserve any existing `manage_token` values for matching aliases, then deduplicate and slice.

Zingzy (changes requested):
## 🤖 AI Code Review Summary

**AI analysis complete. Found 8 issues across 7 files.**

### Issue Breakdown

| Severity | Count |
|---|---|
| 🔴 Critical | 0 |
| 🟠 High     | 2 |
| 🟡 Medium   | 5 |
| 🔵 Low      | 1 |
| **Total**   | **8** |

| Issue Type | Count |
|---|---|
| Bug | 7 |
| Maintainability | 1 |

*8 inline comment(s) posted directly on changed lines.*

---
*Generated by [Code Review Copilot](https://github.com)*

static/js/auth.js
Zingzy, line 510: 🟠 **[HIGH] Bug**

`sweepAndClaimTokens()` sends claim requests with plain `fetch()` instead of `authFetch()`. If the access token is expired immediately after login/register but the refresh cookie is still valid, these claim calls will fail with 401 and the automatic claim sweep will silently skip claiming the user's anonymous links.

**Why this matters in production:** Users can successfully sign in but still not have their anonymous links claimed, making the new feature appear unreliable. In production this leaves links unowned until the user retries manually or loses track of the token.

**Suggested fix:**
Use `authFetch('/api/v1/claim', ...)` here, or add equivalent 401 refresh-and-retry handling before deciding whether the claim failed.

static/js/index-recent-links.js
```diff
@@ -70,32 +70,99 @@ function renderRecentURLs() {
...
+    if (typeof openAuthModal === 'function') {
+        openAuthModal();
+    } else {
+        window.location.href = '/';
```
Zingzy, line 166: 🟡 **[MEDIUM] Bug**

When `openAuthModal` is unavailable, the fallback only redirects to `/` and does not preserve any signal that the homepage should actually open the auth modal. This makes the new "Sign in to claim all your links" CTA dead-end on pages where `auth.js` is not loaded.

**Why this matters in production:** Users can click the claim button and land on the homepage with no login prompt, so the claim flow appears broken. In production that will leave anonymous links unclaimed unless users manually discover how to sign in.

**Suggested fix:**
Before redirecting, store an explicit flag such as `sessionStorage.setItem('spoo_open_auth_modal', '1')` or append a query parameter that homepage/auth bootstrap code reads to auto-open the auth modal after navigation.

services/url_service.py
```diff
@@ -267,7 +276,7 @@ async def create(
             private_stats=private_stats,
         )
 
-        return UrlV2Doc.from_mongo(doc)
+        return UrlV2Doc.from_mongo(doc), raw_token
```
Zingzy, line 279: 🟡 **[MEDIUM] Security**

`create()` now returns a `UrlV2Doc` built from the same in-memory `doc` that contains the hashed `manage_token`. That exposes credential material to every internal caller of this service, even though the PR intent says the token should be excluded from normal responses.

**Why this matters in production:** If another code path later logs or serializes the returned model, the hashed claim token could leak into logs or APIs. Even hashed secrets should be minimized because they expand the attack surface and make future mistakes more damaging.

**Suggested fix:**
Build separate payloads for persistence and return value, or clear `doc["manage_token"]` before calling `UrlV2Doc.from_mongo(doc)` so the returned model never carries the stored token hash.

static/js/result-script.js
Zingzy, line 63: 🟡 **[MEDIUM] Bug**

The alias is taken directly from the URL path, but navigation to this page uses `encodeURIComponent(data.alias)`. For aliases containing spaces, emoji, or other encoded characters, the encoded path segment will not match the raw alias stored in `recentURLs`, so the manage-token banner will never render.

**Why this matters in production:** Users with non-URL-safe aliases won't see their one-time claim token on the result page. That can prevent legitimate anonymous links from being claimed later, making the feature appear unreliable for some users.

**Suggested fix:**
Decode the path segment before comparing it to localStorage entries, e.g. `const alias = decodeURIComponent(window.location.pathname.split('/').pop() || '');`, and use the same normalization anywhere else this file derives the alias from the URL.

static/js/result-script.js
Zingzy, line 100: 🟠 **[HIGH] Bug**

When `openAuthModal` is unavailable on the result page, `claimNow()` falls back to `window.location.href = '/'` without preserving any signal that the homepage should actually open the auth modal. This breaks the intended claim flow on pages where `auth.js` is not loaded.

**Why this matters in production:** A user can click "Claim this link" and land on the homepage with no login prompt or claim action started. In production this makes the new claim flow look broken and will cause users to abandon claiming their anonymous links.

**Suggested fix:**
Before redirecting, store an explicit flag such as `sessionStorage.setItem('spoo_open_auth_modal', '1')` or append a query parameter that homepage/auth bootstrap code reads to auto-open the auth modal after navigation.

repositories/url_repository.py
Zingzy, line 258: 🟡 **[MEDIUM] Bug**

The atomic claim update transfers ownership and clears the manage token, but it leaves `private_stats` unchanged. Anonymous URLs are created with `private_stats=None`, so after a successful claim the document can remain in that anonymous-only state even though it now has a real owner.

**Why this matters in production:** Claimed links can keep inconsistent privacy configuration, causing owned URLs to behave like public/anonymous links and breaking assumptions elsewhere in the app about owner-managed stats privacy.

**Suggested fix:**
Include `private_stats` normalization in the same `$set` used for claiming, e.g. set it to `False` when claiming an anonymous URL whose `private_stats` is `None`, so owned URLs always end up with an explicit boolean value.

Zingzy (changes requested):
## 🤖 AI Code Review Summary

**AI analysis complete. Found 6 issues across 5 files.**

### Issue Breakdown

| Severity | Count |
|---|---|
| 🔴 Critical | 0 |
| 🟠 High     | 2 |
| 🟡 Medium   | 4 |
| 🔵 Low      | 0 |
| **Total**   | **6** |

| Issue Type | Count |
|---|---|
| Bug | 5 |
| Security | 1 |

*6 inline comment(s) posted directly on changed lines.*

---
*Generated by [Code Review Copilot](https://github.com)*

PR #157, opened 2026-04-14
refactor: field handler registry, response DTO factories, and password verification on UrlCacheData

Zingzy:
- UrlService.update() field logic extracted into FIELD_HANDLERS registry with _simple_field_handler factory — adding a URL field is one handler + one registry entry
- from_doc() classmethods on UpdateUrlResponse, UrlResponse, UrlListItem — eliminates duplicate inline construction in routes
- UrlCacheData.verify_password() handles schema-specific hashing (argon2 v2 / plaintext v1) — route no longer knows about SchemaVersion for password checks
- is_emoji_alias() added for dispatch — decoupled from max_emoji_alias_length creation policy so lowering the cap can't orphan existing emoji URLs
- Config field validators on AppSettings — rejects invalid limits/timeouts at startup
- Malformed start_date/end_date now rejected with 400 instead of silently falling back to defaults
- $elemMatch fix for OAuth provider lookup — prevents cross-element array matching

## Summary by Sourcery

Refactor URL update, response DTO construction, and password verification to centralize logic and simplify route handlers.

New Features:
- Introduce a registry of field-specific handlers for URL updates to make adding or changing updatable fields extensible.
- Add factory classmethods on URL response DTOs to build responses directly from UrlV2Doc instances.

Enhancements:
- Extract URL auto-reactivation rules into a dedicated helper on UrlService for clearer expiry handling.
- Return raw UrlV2Doc objects from UrlService.list_by_owner and map them to UrlListItem DTOs at the API layer.
- Move password verification logic for redirects into UrlCacheData so schema-specific hashing is encapsulated in cache models.
- Normalize datetime handling and Unix timestamp conversion for URL list and detail responses via shared utilities.

Tests:
- Update redirect and URL lifecycle integration tests to patch the new password verification location and accommodate the refactored response construction.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated API response construction for improved consistency across endpoints.
  * Enhanced password verification mechanism for protected URLs.
  * Optimized URL update operations processing.
  * Improved internal URL listing response handling.

### DIFF
--- a/infrastructure/cache/url_cache.py
+++ b/infrastructure/cache/url_cache.py
@@ -9,6 +9,7 @@ from dataclasses import asdict, dataclass
 
 import redis.asyncio as aioredis
 
+from shared.crypto import verify_password as verify_password_hash
 from shared.logging import get_logger
 
 log = get_logger(__name__)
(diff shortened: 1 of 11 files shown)


Zingzy:
@coderabbitai review

PR #158, opened 2026-04-14
refactor: reorganize authentication routes and services

Zingzy:
- Moved authentication routes into a dedicated package structure for better organization.
- Replaced the old auth routes with new modular routes for device authentication, password management, and email verification.
- Updated service dependencies to reflect the new structure, enhancing clarity and maintainability.
- Removed the legacy AuthService in favor of a more modular approach with specific services for credentials, device auth, password management, and email verification.

## Summary by Sourcery

Modularize authentication into dedicated services and route modules for credentials, device auth, password management, and email verification, replacing the legacy monolithic auth service.

New Features:
- Introduce dedicated FastAPI route modules for core auth, device auth, password management, and email verification, grouped under an auth routes package.
- Add CredentialService, PasswordService, EmailVerificationService, OtpService, and DeviceAuthService to encapsulate distinct authentication and password-reset responsibilities.
- Expose helper dependencies and utilities such as user_repo access and a reusable fetch_user_profile helper for route handlers.

Bug Fixes:
- Ensure URL alias conflict logging uses a consistent short_code field name instead of alias.

Enhancements:
- Refactor service wiring and DI to register granular auth-related services and shared TokenFactory/OtpService instances on the app state.
- Decouple OAuthService from the legacy AuthService by injecting TokenFactory directly for token issuance.
- Simplify logging utilities by removing an unused context-binding helper and a custom timestamp processor in favor of structlog's TimeStamper configuration.

## Summary by CodeRabbit

* **New Features**
  * Full device authorization flow (login, consent, callback, token exchange/refresh, revoke).
  * Password reset endpoints (request & confirm) and email verification flows.
  * Public auth endpoints: login, register, refresh, logout; authenticated endpoints: /auth/me and set-password.

* **Refactor**
  * Authentication routes reorganized into a single auth package with modular sub-routers.
  * Auth layer split into focused services for clearer responsibilities and easier maintenance.


routes/auth/device.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+def _build_callback_redirect(
+    code: str, state: str, redirect_uri: str, app: AppEntry
+) -> RedirectResponse:
+    """Build the redirect to the callback page or a registered redirect_uri."""
+    params = urlencode({"code": code, "state": state})
+    if redirect_uri and redirect_uri in app.redirect_uris:
+        separator = "&" if "?" in redirect_uri else "?"
+        return RedirectResponse(f"{redirect_uri}{separator}{params}", status_code=302)
```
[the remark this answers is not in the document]
Zingzy: FP

routes/auth/device.py
```diff
@@ -0,0 +1,347 @@
...
+    app = device_auth_service.resolve_app(app_id)
+    if not app:
+        return _device_error(request, "Unknown or unsupported application")
+
+    if not device_auth_service.validate_redirect_uri(redirect_uri, app):
+        return _device_error(request, "Invalid redirect URI for this application")
+
+    # Create grant
+    await grant_repo.create_or_reactivate(user.user_id, app_id)
+    log.info("app_consent_granted", user_id=str(user.user_id), app_id=app_id)
```
Zingzy, line 188: 🟠 **[HIGH] Security**

The consent CSRF token is only compared against the cookie value; its 10-minute TTL is never enforced server-side. Because the form field can be replayed as long as the browser keeps the cookie, a stale consent page can still grant app access well after the intended expiry window.

**Why this matters in production:** A user who leaves an old consent page open could unknowingly approve app access later than intended, and any script or extension with access to that page could replay the approval while the cookie survives. This weakens the time limit the route appears to promise.

**Suggested fix:**
Store an issued-at timestamp or signed expiry in the CSRF token and validate it on POST, or use a server-side session/nonce store so expired consent tokens are rejected even if the cookie is still present.

routes/auth/device.py
```diff
@@ -0,0 +1,347 @@
...
+    return DeviceRefreshResponse(
+        access_token=result.access_token,
+        refresh_token=result.refresh_token,
+    )
+
+
+# ── App revocation (dashboard action) ────────────────────────────────────────
+
+
+@router.post("/auth/device/revoke", include_in_schema=False)
```
Zingzy, line 319: 🟡 **[MEDIUM] Security**

`/auth/device/revoke` relies only on `X-Requested-With: fetch` for CSRF protection. That header is not a secret and can be sent by any same-origin script, so this endpoint has no per-request nonce protection like the consent flow does.

**Why this matters in production:** If any same-origin script injection happens on the site, an attacker can silently revoke a user's app grants without needing to steal credentials. That can break connected extensions or CLI sessions unexpectedly for real users.

**Suggested fix:**
Protect revocation with the same synchronizer-token/double-submit CSRF pattern used for consent, or require an Origin/Referer check in addition to a real CSRF token.

routes/auth/routes.py
```diff
@@ -0,0 +1,292 @@
...
+    401 response is returned.
+
+    **Authentication**: Requires a valid ``refresh_token`` cookie
+
+    **Rate Limits**: 20/min
+
+    **Notes**: The old refresh token is invalidated after use (rotation).
+    """
+    jwt_cfg = request.app.state.settings.jwt
+    refresh_token_str = request.cookies.get("refresh_token")
```
Zingzy, line 190: 🟠 **[HIGH] Security**

The cookie-based `/auth/refresh` route rotates tokens solely from the `refresh_token` cookie and does not require any CSRF signal. Because browsers automatically attach HttpOnly cookies on cross-site POSTs, a third-party site can silently trigger token rotation for a logged-in user.

**Why this matters in production:** A malicious website could cause users' sessions to be refreshed without their intent just by making their browser send a POST request. That weakens your session-boundary protections and can interfere with logout/session management assumptions in production.

**Suggested fix:**
Require a CSRF defense on this endpoint, such as validating an anti-CSRF header/token pair or enforcing strict same-site behavior that is compatible with your clients. Mirror whatever CSRF protection is already used on other cookie-authenticated state-changing routes.

Zingzy (commented):
## 🤖 AI Code Review Summary

**AI analysis complete. Found 10 issues across 6 files.**

### Issue Breakdown

| Severity | Count |
|---|---|
| 🔴 Critical | 0 |
| 🟠 High     | 4 |
| 🟡 Medium   | 6 |
| 🔵 Low      | 0 |
| **Total**   | **10** |

| Issue Type | Count |
|---|---|
| Bug | 7 |
| Security | 3 |

*10 inline comment(s) posted directly on changed lines.*

---
*Generated by [Code Review Copilot](https://github.com)*

services/auth/credentials.py
```diff
@@ -0,0 +1,216 @@
...
+            access_token=access_token,
+            refresh_token=refresh_token,
+            verification_sent=verification_sent,
+        )
+
+    async def refresh_token(self, refresh_token_str: str) -> AuthResult:
+        """Perform stateless token rotation.
+
+        Verifies the refresh JWT, re-fetches the user to get the latest
+        ``email_verified`` status, and issues a new token pair.  The original
```
Zingzy, line 183: 🟠 **[HIGH] Bug**

`refresh_token()` converts the JWT `sub` claim directly with `ObjectId(user_id)` without validating it. If a malformed or legacy refresh token contains a non-ObjectId subject, this raises `InvalidId` and escapes the documented `AuthenticationError` path, turning an auth failure into an unhandled 500.

**Why this matters in production:** A bad refresh token should be rejected cleanly, but today it can crash the request instead. In production that means clients may see intermittent 500 errors during refresh, and malformed tokens can bypass normal auth error handling and logging.

**Suggested fix:**
Wrap the `ObjectId(user_id)` conversion in validation and raise `AuthenticationError("invalid or expired refresh token")` when `sub` is missing or not a valid ObjectId. For example, catch `bson.errors.InvalidId` before calling `find_by_id`.

services/auth/device.py
```diff
@@ -0,0 +1,153 @@
...
+    async def exchange_device_code(self, code: str) -> AuthResult:
+        """Exchange a one-time device auth code for JWT tokens.
+
+        Raises:
+            AuthenticationError: Code invalid, expired, or already used.
+        """
+        svc_log = log.bind(op="auth.device_exchange")
+
+        token_hash = hash_token(code)
+        token_doc = await self._token_repo.consume_by_hash(
```
Zingzy, line 113: 🟠 **[HIGH] Bug**

`exchange_device_code()` still consumes the device code via `consume_by_hash()`, but this PR introduced the new atomic `consume_if_unused()` path specifically to prevent double-use races. As written, two concurrent exchanges of the same device code can both pass the lookup/validation window and mint JWTs before the token is marked used.

**Why this matters in production:** In production, the same one-time device code could be redeemed more than once if requests arrive at nearly the same time. That would let multiple clients obtain valid tokens from a code that is supposed to be single-use.

**Suggested fix:**
Change the exchange flow to first look up the token by hash, then atomically consume it with `consume_if_unused(token_doc.id)` before issuing tokens, mirroring the OTP service's new pattern. If the consume step fails, return `AuthenticationError("invalid or expired device auth code")`.

services/auth/otp.py
```diff
@@ -0,0 +1,167 @@
...
+            )
+            msg = (
+                "Too many password reset attempts. Please try again later."
+                if token_type == TOKEN_TYPE_PASSWORD_RESET
+                else "Too many verification attempts. Please try again later."
+            )
+            raise RateLimitError(msg)
+
+        if delete_existing:
+            await self._token_repo.delete_by_user(user_id, token_type)
```
Zingzy, line 75: 🟡 **[MEDIUM] Bug**

When `delete_existing=True`, `create_otp()` deletes prior OTPs before inserting the new one. If the subsequent repository create or email-send path fails, the user is left with no still-valid code, even if they had one moments earlier.

**Why this matters in production:** A user who retries password reset or verification during a transient DB/email failure can lose their last working code and get stuck until they request another one. In production this shows up as flaky auth flows that are hard for users to recover from.

**Suggested fix:**
Only invalidate older OTPs after the new token has been successfully persisted, or keep the previous token until the caller confirms delivery succeeded. A simple fix is to create the new token first and then delete older tokens excluding the newly created record.

services/auth/otp.py
```diff
@@ -0,0 +1,167 @@
...
+    ) -> None:
+        """Verify an OTP code and mark it as used.
+
+        Raises:
+            ValidationError: On any verification failure (wrong code,
+                expired, max attempts, already used).
+        """
+        svc_log = log.bind(op="otp.verify")
+
+        token_doc = await self._token_repo.find_latest_by_user(user_id, token_type)
```
Zingzy, line 105: 🟡 **[MEDIUM] Bug**

`verify_otp()` always verifies only the latest unused token for the user/type. Because `create_otp()` does not delete existing verification tokens by default, requesting a second email-verification code silently invalidates the first code even though it remains unused and unexpired in storage.

**Why this matters in production:** Users who enter a still-unexpired code from an earlier email can be told it is invalid after requesting a newer one, which feels random and increases support load. It also leaves stale valid tokens in the database that can never succeed, making behavior harder to reason about.

**Suggested fix:**
Either delete older tokens when issuing every OTP, or change verification to look up by hashed code plus user/type instead of always selecting the latest token. That keeps any still-valid issued code verifiable until expiry.

services/auth/otp.py
```diff
@@ -0,0 +1,167 @@
...
+                token_type=token_type,
+            )
+            raise ValidationError(
+                "Too many failed attempts. Please request a new code."
+            )
+
+        # Compare hash — increment attempts on mismatch
+        otp_hash = hash_token(otp_code)
+        if token_doc.token_hash != otp_hash:
+            await self._token_repo.increment_attempts(token_doc.id)
```
Zingzy, line 143: 🟡 **[MEDIUM] Bug**

Failed-attempt counting is not atomic with the max-attempts check. Two concurrent wrong-code submissions can both pass the `attempts >= MAX_VERIFICATION_ATTEMPTS` guard and each increment afterward, allowing more than the intended five guesses.

**Why this matters in production:** An attacker or buggy client can squeeze in extra OTP guesses by racing requests, weakening the brute-force protection you intended. In production that reduces the effectiveness of the verification-attempt cap under load or abuse.

**Suggested fix:**
Move the attempt limit into the repository update condition, e.g. increment only when `used_at is None`, `expires_at > now`, and `attempts < MAX_VERIFICATION_ATTEMPTS`, and treat a failed update as a locked/expired token.

services/auth/verification.py
```diff
@@ -0,0 +1,145 @@
...
+        svc_log = log.bind(op="auth.send_verification")
+
+        user_oid = ObjectId(user_id)
+        user = await self._user_repo.find_by_id(user_oid)
+        if not user:
+            raise NotFoundError("user not found")
+
+        if user.email_verified:
+            raise ValidationError("email already verified")
+
```
Zingzy, line 133: 🟡 **[MEDIUM] Bug**

If `send_verification_email()` returns false, the freshly created OTP remains valid in storage even though no email was delivered. Repeated retries can quickly consume the per-hour OTP quota and leave behind usable codes that the user never received.

**Why this matters in production:** Users can get locked out of requesting another verification code after transient email-provider failures because failed sends still count against the OTP rate limit. It also leaves active verification codes in the database that were never actually delivered, which makes the flow harder to reason about and support.

**Suggested fix:**
Only persist the OTP after the email send succeeds, or delete/invalidate the just-created OTP when `send_verification_email()` fails. If the provider can raise exceptions, wrap the send in try/except and perform the same cleanup there.

services/auth/verification.py
```diff
@@ -0,0 +1,145 @@
...
+        Raises:
+            NotFoundError:   User not found.
+            ValidationError: Email already verified.
+            RateLimitError:  Too many recent verification requests.
+            AppError:        Failed to send email.
+        """
+        svc_log = log.bind(op="auth.send_verification")
+
+        user_oid = ObjectId(user_id)
```
Zingzy, line 126: 🟡 **[MEDIUM] Bug**

`ObjectId(user_id)` is called without handling invalid IDs. Unlike the explicit `NotFoundError` path, a malformed `user_id` will raise BSON's `InvalidId` and escape as an unexpected 500-level error from this service.

**Why this matters in production:** A bad or corrupted user ID in the auth context can crash this endpoint instead of returning a controlled client error. In production that means noisy error logs and inconsistent API behavior for malformed requests or stale tokens.

**Suggested fix:**
Validate/convert `user_id` defensively and translate BSON conversion failures into an application-level validation/not-found error, e.g. catch `InvalidId` around `ObjectId(user_id)` and raise `ValidationError` or `NotFoundError` consistently.

PR #230, opened 2026-07-08
feat: per-country geo-targeting rules on short links

### LINKED ISSUE #13: Feature: Allow Users to Geo-Target Their URLs
## Description
Implement a feature that allows users to geo-target their URLs. This will enable users to redirect visitors to different URLs based on their geographic location, enhancing the relevance and effectiveness of their content.

## Tasks
- [ ] **Detect User Location**
    - Implement functionality to detect the geographic location of users based on their IP address.
    - Ensure accurate detection across a wide range of locations.

- [ ] **Store Geo-Targeted URLs**
    - Update the database schema to store URLs specific to different geographic locations.
    - Ensure backward compatibility with existing URLs.

- [ ] **Redirect Based on Location**
    - Implement logic to redirect users to the appropriate URL based on their geographic location.
    - Ensure the redirection is seamless and does not introduce significant latency.

- [ ] **User Interface Updates**
    - Update the URL creation and management interface to allow users to specify geo-targeted URLs.
    - Ensure the interface is user-friendly and intuitive.

- [ ] **Testing**
    - Write unit tests to verify the accuracy of location detection and redirection logic.
    - Perform integration tests to ensure the feature works seamlessly with the existing system.

- [ ] **Documentation**
    - Update the documentation to include details about the new geo-targeting feature.
    - Provide instructions on how to create and manage geo-targeted URLs.

Zingzy:
## What

Geo-targeting v1 (API-only): a link can now carry per-country destination overrides — visitors from a listed country get their URL, everyone else gets the default `long_url`.

```jsonc
POST /api/v1/shorten
{
  "url": "https://example.com/global",
  "geo_rules": { "IN": "https://example.in/", "US": "https://example.com/us" }
}
```

- Flat map shape (ISO 3166-1 alpha-2 → URL), max 50 entries. Keys case-normalised, case-collisions rejected, codes validated via pycountry. Errors use field paths like `geo_rules.IN`.
- `PATCH /urls/{id}`: omit = unchanged, `null`/`{}` = clear all, map = full replace. Echoed in create/update/list responses.
- No UI — the new frontend consumes this API later. The backend-rendered `+` preview page is updated as part of the security posture (below).

## How it works

**Redirect hot path** (`routes/redirect_routes.py`): decision runs between the password check and bot check — `CF-IPCountry` header first (free, CF-injected), fallback to a new `GeoIPService.get_country_code()` mmdb lookup that only fires for geo links when the header is absent (self-host). O(1) dict lookup; `Cache-Control: no-store` on geo-varied responses; decision stamped on `ClickEvent` (`resolved_country`, `geo_matched`) and the sampled redirect log for auditability. Unknown countries (`XX`, `T1`, unresolvable) fall back to the default.

**Caching**: `geo_rules` added to the `UrlCacheData` Redis projection (pre-deploy cache entries deserialise to `None` — no version bump).

**Edge cache**: hot geo links are served **from the CF Worker** via a new `geo_redirect` KV entry type — `url` is the default destination plus a `rules` map, and the Worker serves `rules[request.cf.country] ?? url` with `Cache-Control: no-store`. This is the same CF geodata origin reads from `CF-IPCountry`, so edge-served and origin-served decisions can never disagree. Lockstep contract change in one commit (Python models + JSON schema + fixtures + Worker + both test suites); rules-less geo entries and unknown types still passthrough (fail-open pinned by the malformed fixtures), so deploy order between app and Worker doesn't matter. Rule edits ride the same TTL-staleness window (~≤6 min) as plain URL edits.

**Feature gating**: `geo_targeting` flag via the existing flag service, **write-side only** — setting rules on create/edit returns 403 when off, anonymous callers get 401 before the flag is consulted, and clearing rules always works so de-allowlisted owners can roll back. The redirect path never checks the flag.

## Security

- Every geo destination gets the same two-stage validation as `long_url` (`validate_url` + DB blocklist) on **both** create and edit, with patterns fetched once per request.
- Fixes a pre-existing gap: editing `long_url` skipped the blocklist check that create runs — edits were a side door for blocked destinations. Now closed (with a regression test).
- Anti-cloaking: the `spoo.me/abc+` preview page shows a variance banner and **every** destination (default + each rule, grouped by URL) — no geo rule is ever hidden from someone previewing the link. Password-protected links still hide everything.

## Testing

- 1863 Python tests + 37 Worker tests (vitest in workerd) pass, ruff + tsc clean.
- New coverage: validation field paths, update-handler set/replace/clear/no-op semantics, cache-projection survival on the cache-hit path, ClickEvent stream backward-compat (pre-deploy payloads still decode), geo promotion as `geo_redirect` entries, Worker per-country serving (match / unmatched / XX / T1 / missing `cf.country` / lowercase normalisation / no-store), per-country redirect matrix at origin, flag gating, preview transparency incl. password precedence.
- Live smoke against real Mongo: header match, lowercase normalisation, unmatched fallback, mmdb fallback, HEAD, and preview page all verified end-to-end.
- k6 load+stress (local scenario, not committed): 125k requests at up to 150 VUs across both resolution paths, **zero wrong-country redirects** (hard `count<1` threshold), p95 within the repo's 500ms smoke bar.

## Ops note

The flag service is default-deny: nothing is enabled until a `geo_targeting` flag doc is seeded in Mongo (`rollout_type: ALLOWLIST` to start, same as custom_domains).

## Deferred (by design)

`EU`/continent aliases, clicks-by-destination stats dimension, dashboard UI.

## Summary by CodeRabbit

* **New Features**
  * Added geo-targeted URL redirects, letting links send visitors to different destinations by country.
  * URL creation, updates, previews, and listings now support country-based destination rules.
  * Preview pages now show a clearer geo-targeting banner and destination breakdown.

* **Bug Fixes**
  * Redirects now avoid caching when destination depends on country.
  * Anonymous users can no longer set geo-targeted destinations.
  * Existing cached and legacy records continue to load safely.


tests/integration/test_legacy_url_routes.py
```diff
@@ -590,3 +590,92 @@
...
+
+        app = build_test_app(legacy_url_router, overrides={get_db: lambda: db})
+        with TestClient(app) as client:
+            resp = client.get("/abc1234+")
+
+    assert resp.status_code == 200
+    html = resp.text
+    assert "different destinations depending on the visitor" in html
+    assert "Default destination:" in html
+    assert "example.com" in html  # default
```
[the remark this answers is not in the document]
Zingzy: False positive — this is a test assertion checking that the rendered preview HTML contains an expected destination row (`assert "example.in" in html`). It's not sanitizing or authorizing a URL; the substring check is exactly what a template-rendering assertion should do. Dismissing.

tests/integration/test_legacy_url_routes.py
```diff
@@ -590,3 +590,92 @@
...
+    assert "example.in" in html
+    assert "us.example.com" in html
```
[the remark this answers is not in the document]
Zingzy: False positive — same as above: a template-rendering test assertion on preview HTML output, not URL sanitization logic. Dismissing.

schemas/dto/requests/url.py
```diff
@@ -25,6 +25,32 @@
...
+def _normalise_geo_rules(v: dict | None) -> dict | None:
+    """Normalise geo_rules keys to uppercase ISO codes.
+
+    JSON parsing guarantees key uniqueness, but normalisation could merge
+    keys like `"in"` and `"IN"` — reject that instead of silently picking one.
+    Semantic validation (real ISO codes, URL safety) lives in the service layer.
+    """
+    if v is None:
+        return None
+    normalised: dict[str, str] = {}
```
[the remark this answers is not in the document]
Zingzy: Good catch — fixed in 26ff723. Non-dict values now pass through the before-validator untouched so Pydantic rejects them with its normal dict type error (422), and there's a regression test covering list/str/int payloads.

services/url_service.py
```diff
@@ -64,6 +66,43 @@
 
 AliasCheckResult = Literal["available", "length", "format", "taken"]
 
+
+def _validate_geo_rules(
+    rules: dict[str, str],
+    *,
+    blocked_self_domains: tuple[str, ...],
+    patterns: list,
+    timeout: float,
```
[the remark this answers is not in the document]
Zingzy: Fixed in 26ff723 — annotated as `Sequence[str]`, matching `validate_blocked_url`'s signature (patterns are regex strings from the blocklist collection).

routes/legacy/url_shortener.py
```diff
@@ -517,16 +533,20 @@ async def preview_url(
...
+    # by destination URL (display only; storage stays a flat code→url map).
+    geo_destinations = None
+    if url_data.get("geo_rules"):
+        grouped: dict[str, list[str]] = {}
+        for code, dest in url_data["geo_rules"].items():
+            grouped.setdefault(dest, []).append(code)
+        geo_destinations = [
+            {"countries": codes, **_split_destination(dest)}
+            for dest, codes in grouped.items()
+        ]
```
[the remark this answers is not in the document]
Zingzy: Done in 26ff723 — codes are sorted when building the preview rows. Display-only; storage and routing keep the map as-is.

schemas/dto/requests/url.py
```diff
@@ -25,6 +25,32 @@
...
+            raise ValueError(f"duplicate country code after normalisation: '{code}'")
+        if not isinstance(url, str) or not url.strip():
+            raise ValueError(f"geo_rules['{code}'] must be a non-empty URL string")
+        if len(url) > _GEO_URL_MAX_LENGTH:
+            raise ValueError(
+                f"geo_rules['{code}'] URL exceeds {_GEO_URL_MAX_LENGTH} characters"
+            )
+        normalised[code] = url.strip()
+    return normalised
+
```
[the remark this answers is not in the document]
Zingzy: Fixed in 26ff723 (same issue flagged by Sourcery): `isinstance(v, dict)` guard passes non-dicts through to Pydantic's own type validation → clean 422 instead of an AttributeError. Went with pass-through rather than raising ValueError so the error message comes from Pydantic's standard dict-type error. Regression test added.

services/url_service.py
```diff
@@ -64,6 +66,43 @@
...
+                field=f"geo_rules.{code}",
+            )
+        if not validate_url(url, blocked_self_domains=blocked_self_domains):
+            raise ValidationError(
+                "URL is not allowed or invalid", field=f"geo_rules.{code}"
+            )
+        if not validate_blocked_url(url, patterns, timeout=timeout):
+            raise ValidationError("URL is blocked", field=f"geo_rules.{code}")
+
+
```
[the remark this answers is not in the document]
Zingzy: Declining this one deliberately. The worst case is bounded and admin-controlled: patterns come from the operator-curated blocklist collection (not user input), each pattern runs under the regex library's 0.2s ReDoS timeout, and this check runs only on rate-limited authenticated write endpoints (create/edit) — never on the redirect hot path. It's also byte-for-byte the same synchronous pattern the existing long_url validation has always used in create; moving only geo rules to a thread would leave that inconsistency. If blocklist regex cost ever becomes a real issue, offloading belongs in validate_blocked_url itself as a codebase-wide change, not in this feature PR.

routes/api_v1/feature_gates.py
Zingzy, line 1: are we sure `routes/api_v1/` is the right place to put this?

routes/redirect_routes.py
```diff
@@ -144,6 +145,23 @@ async def redirect_url(
                 status_code=401,
             )
 
+    # 2b. Geo-targeting decision — must run before the click event is built
```
Zingzy: what are these comments? what is `2b`?

schemas/models/url.py
```diff
@@ -19,6 +19,10 @@
 
 from schemas.models.base import ANONYMOUS_OWNER_ID, MongoBaseModel, PyObjectId
 
+# Max country entries in a geo_rules map. Bounds doc size, cache entry size
+# and per-request validation work.
+GEO_MAX_COUNTRIES = 50
```
Zingzy: should this not go to the constants settings?

Zingzy: there was some setting where all constants like these are stored

services/url_service.py
```diff
@@ -64,6 +66,43 @@
 
 AliasCheckResult = Literal["available", "length", "format", "taken"]
 
+
+def _validate_geo_rules(
```
Zingzy, line 70: are these helper functions stored in `services`? or are they moved to some utils? what's the general pattern in the codebase?

routes/api_v1/feature_gates.py
Zingzy: Considered the alternatives and I'd keep it here — reasoning: `dependencies/` holds DI providers and `Depends`-style guards (`require_scopes` etc.), but this check is conditional on request body content (only fires when `geo_rules` is present), so it can't be expressed as a dependency. `shared/` is framework-free pure functions, and this raises HTTP-semantic errors. The closest precedent for 'helper shared across route modules' is `routes/cookie_helpers.py` / `routes/legacy/helpers.py` — a route-layer helper module next to its consumers. Since both consumers are api_v1 endpoints, `routes/api_v1/feature_gates.py` is the narrowest home. Happy to move it if you have a different convention in mind.

routes/redirect_routes.py
Zingzy: The route body is organised as numbered step comments (1. resolve, 2. password, 3. bot block, …) and I'd wedged the geo decision in as `2b` to avoid renumbering. Lazy — fixed in 7fd17e1: steps are now cleanly 1–6 with geo as step 3.

schemas/models/url.py
Zingzy: You're right, and found the place you meant — the "Validator constraints (overridable by self-hosters via env vars)" section in `AppSettings` (`blocked_url_regex_timeout`, `max_emoji_alias_length`, `url_password_min_length`…). Moved in 7fd17e1: `geo_rules_max_countries: int = 50` lives there now, added to the positive-int validator, threaded through the `UrlService` ctor via wiring exactly like `max_emoji_alias_length`. Self-hosters can override via env; the DTO no longer hardcodes the number (it keeps only shape normalisation — same DTO/service split as the emoji alias cap).

services/url_service.py
Zingzy: The codebase has a two-tier pattern and this follows it: (1) `shared/validators.py` = pure, framework-free, bool-returning validators (`validate_url`, `validate_blocked_url`, `validate_alias`…) — reusable anywhere, no app-error coupling. (2) Service modules keep module-level helpers that ORCHESTRATE those primitives and raise `errors.ValidationError` with field paths — precedents in this same file: `_raise_for_status`, `_v2_doc_to_cache`, `_legacy_doc_to_cache`, and all the `_handle_*` FIELD_HANDLERS. `_validate_geo_rules` composes shared validators + app errors + a settings-driven cap, so it belongs in tier 2, module-level next to its only consumers. If it ever grows a second consumer outside UrlService, promoting the pure parts into `shared/validators.py` would be the move.

PR #231, opened 2026-07-08
feat: custom meta-tags (social link previews)

### LINKED ISSUE #234: [FEAT] Custom meta tags: per-link social previews (og:title / og:image / theme color)
## Explanation

Let link owners customize how their short links look when shared: the title, description, image, and accent color that render as the preview card on WhatsApp, Discord, X, Telegram, Slack, iMessage, LinkedIn, etc.

## Is your feature request related to a problem? Please describe.

Today a shared spoo.me link always previews with whatever meta tags the *destination* page has. That means no control over how a link presents itself: campaign links can't carry their own framing, links to pages with poor or missing OG tags produce broken-looking cards, and the preview can't be updated without touching the destination site.

## Describe the solution you'd like

A `meta_tags` object on the existing API (`POST /api/v1/shorten` and `PATCH /api/v1/urls/{id}`):

- `title` (required), `description`, `image` (https URL **or** direct upload), `color` (embed accent, e.g. Discord's border)
- Link-preview crawlers get a small prerendered HTML page carrying the full OG + Twitter Card tag set
- Humans, search engines, and everything else keep today's redirect, byte-for-byte
- A companion `GET /api/v1/metadata?url=` endpoint that fetches and parses a destination's *existing* tags, so clients can prefill before customizing

## Describe alternatives you've considered

- **Injecting tags client-side.** Impossible; preview crawlers don't execute JS.
- **Rendering the preview page on every request.** Wasteful; the card only changes when the owner edits it, so it can be prerendered at write time and served statically.

---

## Tasks

- [x] `LinkMetaTags` data model + `meta_tags` on shorten/PATCH APIs
- [x] Preview-crawler classifier (vendored positive allowlist, shared by origin and edge)
- [x] Prerendered preview page with the full OG + Twitter tag set
- [x] Edge-first serving (KV write-through, worker serves crawlers with zero origin round-trips)
- [x] Transparency card on the `spoo.me/{alias}+` preview page
- [x] Direct image upload (data URI to object storage) with synchronous validation
- [x] Async validation of external image URLs (SSRF-guarded, dimensions recorded)
- [x] Advisory `warnings` for platform image cliffs (size, dimensions, aspect ratio)
- [x] `GET /api/v1/metadata` prefill endpoint with caching and rate limits
- [ ] Verification on real platforms (Facebook debugger, Telegram, Discord, X) + edge deploy
- [ ] Gradual rollout behind the feature flag
- [ ] Dashboard UI (API-first for now, UI tracked separately)

## Code Example

```bash
curl -X POST https://spoo.me/api/v1/shorten \
  -H "Authorization: Bearer <api-key>" \
  -H "Content-Type: application/json" \
  -d '{
    "url": "https://example.com/launch",
    "meta_tags": {
      "title": "We just launched 🎉",
      "description": "The all-new example.com. Faster, cleaner, open source.",
      "image": "https://cdn.example.com/launch-card.png",
      "color": "#FF5733"
    }
  }'
```

## Additional context

Implementation is underway in #231. Design and architecture notes are in the comments below. v2 links only.

Zingzy:
Per-link custom social previews: link owners set `title`, `description`, `image`, and `color` (Discord embed accent), and link-preview crawlers (WhatsApp, iMessage, Discord, Slack, Telegram, facebookexternalhit, Twitterbot, …) receive a prerendered HTML page carrying the full OG + Twitter tag set. Humans, search engines, AI crawlers, and generic scrapers keep today's 302 byte-for-byte.

## Serving architecture (edge-first)

- On every meta-tags write, origin renders `meta_preview.html` once and eagerly writes an `og_only` entry to Workers KV. The edge worker answers preview bots straight from KV (~0ms origin); humans always pass through so origin keeps click tracking. Hot og-links get `redirect` entries with embedded `og_html` via the existing hotness promotion.
- Classification is a **positive allowlist**, not generic bot detection: `wants_preview()` matches ~65 source-verified preview-crawler tokens (`data/preview_bots.json`, byte-identical copy in the edge contract dir, pinned by tests on both sides) plus HEAD and `?bot=1` (debug view). Search/AI crawlers simply don't match — no cloaking exposure, destination keeps its SEO/link equity. In-app-browser tokens (`Line/`, bare Pinterest/Viber/Snapchat) are regression-tested as humans.
- `request.cf.verifiedBotCategory` is wired as an additive edge signal but ships **log-only** (`cf_categories: []`) until recon on the real zone records the runtime strings.
- Preview serves are never clicks, at edge or origin.
- Old workers pass through the unknown `og_only` type and ignore `og_html` on redirect entries — no lockstep deploy needed.

## API (feature is write-gated)

- `meta_tags` object on `POST /api/v1/shorten` and `PATCH /api/v1/urls/{id}` (whole-object replace, `null` clears; clearing is never gated). Requires a verified account with the `custom_meta_tags` flag (403 otherwise). Blocklist regexes run over all text fields plus a destination re-check on every meta write.
- `image` accepts an https URL or a `data:image/png|jpeg|webp;base64` URI — decoded, magic-byte matched against the declared type, capped at 512KB, stored content-addressed in R2 under `og/{owner_id}/` (self-hosted bytes are scannable and takedown-able; no TOCTOU). R2 client is ~50 lines of stdlib SigV4 over the existing httpx client — **zero new dependencies** (`imghdr` is gone in 3.13, so there's a stdlib magic-byte sniffer too).
- External https images are validated out-of-band: own stream (`events:meta-image`) + `meta-image` consumer group in the click worker, SSRF-guarded fetch (DNS resolved first, private/link-local/IPv4-mapped-IPv6 rejected, connection pinned to the validated IP, per-hop re-checks), CAS-filtered writes so a racing user edit is never clobbered. Success records dims → the served page gains `og:image:width/height`.
- `GET /api/v1/metadata?url=` — prefill companion: fetches the destination through the same guarded fetcher, parses tags with a head-only stdlib parser, Redis-cached 1h (5m negative), auth-required, 20/min.
- `CurrentUser.tier` now carries `UserDoc.plan`, so flipping the flag ALLOWLIST→TIER at paid launch is a pure data change.
- The `/{alias}+` preview page shows the custom card next to the real destination (anti-phishing transparency), and `block_bots` og-links never leak their destination into the bot page.

## Testing

- 1965 Python + 55 edge worker (vitest) tests, including the cross-language classification matrix and contract fixtures.
- Full local e2e: gated API flow (register→verify→API key→flag), edge serving via `wrangler dev` + explorer-API KV write-through, async image validation against a live remote image, data-URI upload to MinIO via the SigV4 client, click recording through the stream for edge passthroughs (and none for preview serves), SSRF rejection of metadata-endpoint probes.

## Rollout notes

- Seed `feature_flags`: `{name: "custom_meta_tags", enabled: true, rollout_type: "allowlist", allowlist_user_ids: [...]}` (values are lowercase).
- Edge: deploy to beta, run the `verifiedBotCategory` recon (strings land in `cf_categories` in BOTH `preview_bots.json` copies). Empty = UA-regex-only, fully functional.
- Set the five `R2_*` env vars to enable uploads; without them https image URLs work and uploads reject with a clear error. Self-hosts without CF/queue Redis degrade gracefully (origin serves previews, validation skips).

## Summary by CodeRabbit

* **New Features**
  * Added custom social preview metadata, including titles, descriptions, images, colors, and image-quality warnings.
  * Link-preview crawlers now receive dedicated Open Graph/Twitter preview pages, while human visitors continue to redirect normally.
  * Added an authenticated metadata endpoint for retrieving and parsing destination-page tags.
  * Added support for HTTPS and uploaded image assets, with validation and secure handling.
  * Added edge-cached previews for faster link-sharing experiences.

* **Bug Fixes**
  * Improved preview behavior for HEAD requests, password-protected links, geo redirects, and cache bypasses.
  * Added SSRF protections for outbound metadata and image fetching.


shared/sigv4.py
Zingzy, line 1: is this a storage implementation or a helper function? is it supposed to be here in `infrastructure/storage`?

infrastructure/safe_fetch.py
```diff
@@ -0,0 +1,196 @@
...
+
+import dns.asyncresolver
+import dns.exception
+import httpx
+
+from infrastructure.logging import get_logger
+
+log = get_logger(__name__)
+
+_USER_AGENT = "spoo.me-og-validator/1.0 (+https://spoo.me)"
```
Zingzy: should we put these values here? or should this go to the constants?

Zingzy: will this user agent be trusted by most of the sites?

infrastructure/queue_redis.py
```diff
@@ -35,12 +35,15 @@ def parse_redis_version(version: str) -> tuple[int, int] | None:
...
-    None (→ inline fallback in wiring) when: stream mode isn't requested,
-    the server is unreachable, or the server predates ACKED trimming.
+    None when: no URI, the server is unreachable, or the server predates
+    ACKED trimming. Each consumer of the connection decides for itself
+    (click-sink wiring checks ``sink == "stream"``; the meta-image sink
+    checks its own toggle) — inline-click deployments can still run async
+    meta-image validation.
     """
-    if settings.sink != "stream" or not settings.queue_redis_uri:
+    if not settings.queue_redis_uri:
```
Zingzy, line 46: why was this logic changed?

routes/api_v1/_meta_gate.py
Zingzy, line 1: for this check #230 , gates were configured in some other way!

routes/redirect_routes.py
```diff
@@ -132,6 +133,29 @@ async def redirect_url(
         return _error_page(request, "410", "SHORT URL EXPIRED", 410)
     resolve_ms = int((time.perf_counter() - resolve_start) * 1000)
 
+    # 1b. Custom meta-tags: preview crawlers get the owner's OG card instead
```
Zingzy: what is `1b`? make comments more generic

services/url_service.py
```diff
@@ -424,6 +526,16 @@ async def create(
             )
             raise ValidationError("URL is blocked", field="long_url")
 
+        # 2b. Meta-tags: resolve data-URI uploads to R2 URLs, then abuse
```
Zingzy: generic comments?

services/url_service.py
```diff
@@ -380,6 +438,50 @@ async def check_alias(
...
+            ):
+                log.info("meta_tags_rejected", reason="blocked_pattern")
+                raise ValidationError(
+                    "meta_tags content is not allowed", field="meta_tags"
+                )
+        if not validate_blocked_url(
+            long_url, patterns, timeout=self._blocked_url_regex_timeout
+        ):
+            log.info("meta_tags_rejected", reason="blocked_destination")
+            raise ValidationError("URL is blocked", field="long_url")
```
Zingzy, line 576: why is this a concern of custom meta-tags?

services/url_service.py
```diff
@@ -235,6 +279,10 @@ def __init__(
         system_default_domain: str,
         blocked_url_regex_timeout: float = 0.2,
         max_emoji_alias_length: int = 15,
+        og_writethrough: OgEdgeWritethrough | None = None,
+        r2_storage: R2StorageClient | None = None,
+        meta_image_max_bytes: int = 512_000,
```
Zingzy, line 361: should this be a separate config value?

services/meta_tags/events.py
Zingzy, line 1: not sure what's the purpose of this file?

services/meta_tags/parse_html.py
Zingzy, line 1: can using `bs4` make things faster?

services/meta_tags/validator.py
```diff
@@ -0,0 +1,117 @@
...
+        log.info(
+            "meta_image_validated",
+            url_id=event.url_id,
+            bytes=meta["bytes"],
+            width=meta["width"],
+            height=meta["height"],
+            cas_applied=changed,
+        )
+
+    async def _invalidate(self, alias: str, domain: str) -> None:
```
Zingzy, line 126: why is this needed here??

services/meta_preview.py
```diff
@@ -0,0 +1,40 @@
...
+
+from typing import TYPE_CHECKING
+
+from shared.url_utils import extract_hostname
+
+if TYPE_CHECKING:
+    from infrastructure.cache.url_cache import UrlCacheData
+
+
+def build_preview_context(url: UrlCacheData, *, auto_redirect: bool = True) -> dict:
```
Zingzy, line 18: what if it was not cached?

routes/api_v1/metadata.py
```diff
@@ -0,0 +1,122 @@
...
+            timeout=_FETCH_TIMEOUT,
+            max_bytes=_FETCH_MAX_BYTES,
+            # Big pages are fine — tags live in <head>, parse the prefix.
+            truncate_over_cap=True,
+            user_agent=settings.meta_tags.fetch_user_agent,
+        )
+    except FetchTransientError as exc:
+        raise UpstreamTimeoutError("destination did not respond in time") from exc
+    except FetchHardError as exc:
+        payload = {"error": f"destination is not a fetchable HTML page ({exc})"}
```
Zingzy: **Resolution oracle (should-fix).** `exc` here is the raw `FetchHardError` message, and it's returned to the caller. The messages are distinguishable — `"host does not resolve"` vs `"host resolves to a non-public address"` vs `"content-type 'text/html'"` vs `"status 404"` — so an authenticated user (20/min) can use this endpoint to learn whether any hostname resolves and whether it resolves to a private/internal IP. Fetching private-IP *content* is still blocked, so no internal content leaks, but the resolve-level disclosure shouldn't ship in an outbound-fetch feature.

Return one generic client-facing message and keep the specific `exc` in the structured log only.

services/meta_tags/validator.py
```diff
@@ -0,0 +1,135 @@
...
+        await self._url_cache.invalidate(alias, domain)
+        if self._og_writethrough is None:
+            return
+        # Re-render the edge entry from fresh DB state (the cache was just
+        # invalidated, so a cache read would miss).
+        from services.url_service import _v2_doc_to_cache
```
Zingzy: **Private cross-module import, third call site (should-fix).** This function-body `from services.url_service import _v2_doc_to_cache` dodges a circular import by reaching into a url_service private. The `UrlV2Doc → UrlCacheData` projection is now needed by three parties (url_service, the edge write-through, and this validator), which means it has outgrown its home as a private helper.

Promote it to a `UrlCacheData.from_v2_doc(doc)` classmethod on the cache model — kills the deferred import and the circularity, and matches the layering rule that derived facts live on the model. ~20 min.

services/url_service.py
```diff
@@ -458,6 +516,50 @@ async def check_alias(
...
+
+    async def validate_meta_tags(self, meta: MetaTagsRequest, *, long_url: str) -> None:
+        """Abuse checks for a meta_tags write.
+
+        Blocklist regexes run over title/description/image (the same
+        patterns the long_url flows through at create), plus a destination
+        re-check — a link that was fine as a bare redirect deserves a second
+        look the moment someone dresses it up with a custom preview.
+        """
+        patterns = await self._blocked_url_repo.get_patterns()
```
Zingzy, line 553: **Off-loop, for consistency (should-fix).** `d4a18fa` moved geo's blocklist validation to `asyncio.to_thread`; `validate_meta_tags` still runs synchronous `validate_blocked_url` inline over the text fields + destination. Fewer passes than geo's ×50 so the ceiling is lower, but it should ride the same `to_thread` treatment as the fix next door — the blocklist grows as the abuse strategy matures.

services/click/bot_detection.py
```diff
@@ -44,6 +54,34 @@ def _load_bot_user_agents() -> list[str]:
...
+    """Should this request get the custom OG page instead of the 302?
+
+    Positive allowlist of link-preview crawlers only — NOT generic bot
+    detection. Search/AI crawlers, scrapers and humans all fall through to
+    the 302 (search engines keep destination SEO/link equity; a missed
+    preview bot just follows the redirect and shows the destination's own
+    tags — today's behavior). HEAD ⇒ preview: no real user HEADs; email
+    scanners and link expanders do. Only ever called for og-enabled links.
+    """
+    if bot_param or method == "HEAD":
```
Zingzy, line 77: **Confirm intended: HEAD ⇒ 200 text/html.** Any HEAD returns `True` regardless of UA, so a HEAD to an og-enabled link serves the OG page (200, no `Location`) where every other link returns 302. The rationale (link expanders / email scanners HEAD) is sound and no existing link regresses, but uptime monitors and clients that HEAD to discover the redirect target will now see 200-no-Location on og-links. If that bites, gate the HEAD shortcut on an actual preview UA rather than method alone.

schemas/dto/requests/url.py
```diff
@@ -84,6 +84,62 @@ class UrlFilter(RequestBase):
...
+        examples=["We just launched 🎉"],
+    )
+    description: str | None = Field(
+        default=None,
+        max_length=240,
+        description="og:description — roughly 200 chars render on most platforms.",
+    )
+    image: str | None = Field(
+        default=None,
+        max_length=700_000,  # data URIs: ~512KB decoded at 4/3 base64 inflation
```
Zingzy: **Two caps that can silently disagree (nit).** This DTO hardcodes `max_length=700_000` chars (~512KB base64-inflated), while the actual decoded cap is `meta_image_max_bytes` from `R2_UPLOAD_MAX_BYTES` (self-hoster tunable). A self-hoster who raises the byte cap above ~512KB is still clamped here. Either derive this bound from settings, or document that raising one requires raising the other (the `MAX_CONTENT_LENGTH` interplay is already noted, so the awareness is there).

Zingzy (commented):
## Review: custom meta-tags

**Strong work — approve with a few should-fixes, nothing blocking.** This is a much bigger and riskier surface than geo-targeting (outbound fetching, user-image hosting, a stored-HTML page served from origin, a new edge-serving mode), and the security fundamentals are handled with real care. The house patterns are followed almost everywhere, and the flag-gating now matches geo exactly.

I traced the load-bearing security properties directly:
- **SSRF fetcher is solid** — per hop it re-requires https, re-resolves the host, rejects if *any* A/AAAA is private (rebinding + split-horizon), connects to the resolved IP with SNI/Host preserved, and streams under a hard cap. Redirects re-validate from the top.
- **Stored-XSS is covered** — `meta_preview.html` renders user text into `text/html` served from the origin with no CSP, so Jinja autoescape is load-bearing. It's on (Starlette forces `autoescape=True`), `render.py` reuses the same env, no `|safe`, and the one inline script uses `{{ long_url | tojson }}`. Worth a defense-in-depth CSP on that response as a follow-up, since it all rests on autoescape staying on.

### Should-fix (before merge)
1. **`/api/v1/metadata` leaks a resolution oracle** — the error string differentiates "host does not resolve" vs "resolves to a non-public address" vs content-type/status. An authed user can probe whether arbitrary hosts resolve and whether they're internal. Collapse to one generic message; keep detail in logs. (inline)
2. **`_v2_doc_to_cache` imported privately from a third place** — the deferred `from services.url_service import _v2_doc_to_cache` in the validator. Promote to `UrlCacheData.from_v2_doc()`. (inline)
3. **Run `validate_meta_tags` blocklist checks off the event loop**, like geo now does (`d4a18fa`). (inline)

### Confirm-intended
4. **HEAD on an og-link now returns 200 text/html instead of 302** — uptime monitors / link scanners that HEAD will see 200-no-Location. No existing link regresses (meta_tags is new). (inline)
5. **Two image-size caps that can silently disagree** — DTO hardcodes 700_000 chars while the real cap is `R2_UPLOAD_MAX_BYTES`. (inline)

### Product / abuse (roadmap, not this PR)
This is the cloaking primitive squared — stacked with geo, a link can vary destination by country *and* present a custom card. The mitigations here are the strongest part (positive-allowlist classification so search/Safe Browsing keep seeing the real destination, the `+` page showing the card next to the real destination, both-path blocklist re-checks, `updated_ip` forensics, self-hosted scannable image bytes). Two follow-ups:
- Abuse tooling + threat-intel scanning must read `meta_tags` (title/description/image) and geo rules, not just `long_url` — the card is the payload the victim sees.
- spoo now hosts user images on `og.spoo.me` — a content-moderation surface. Magic-byte gate + `delete_object` takedown exist; proactive perceptual-hash scanning is worth a roadmap item.

### Notably good
The SSRF fetcher's transient/hard/denied taxonomy (esp. `FetchDeniedError` = indeterminate, don't clear the user's image), CAS-on-image_url concurrency, omit-not-null wire format for old-worker compat, and the self-host degradation matrix (no R2 → uploads rejected cleanly, https URLs still work; no queue Redis → sync-only; no edge → origin serves). Test coverage is thorough — cross-language classification matrix, per-audience edge fixtures, block_bots-omits-destination on both origin and write-through paths.

Settle #1 and #2 before merge; #3–#5 are fine as fast-follows.

services/meta_preview.py
```diff
@@ -0,0 +1,40 @@
...
+    """Template context for meta_preview.html.
+
+    ``long_url`` is omitted for block_bots links — a bot-blocked destination
+    must never leak to bots. ``auto_redirect`` is disabled for ?bot=1 so
+    developers can inspect the page a crawler sees.
+    """
+    reveal = not url.block_bots
```
Zingzy: **🔴 Blocker — password bypass.** `reveal` gates only on `block_bots`, so for a password-protected link (block_bots False) the context below includes `long_url`, `dest_host`, and the `location.replace(...)` auto-redirect. The origin preview branch runs *before* the password gate and the edge write-through fires for password links, so `curl -A 'Twitterbot/1.0' https://spoo.me/<code>` (or HEAD, or a real unfurl) returns the destination + a JS redirect to it — password protection bypassed. A real browser still gets 401, which is why it's easy to miss.

This is the single choke point for origin + promotion + write-through:
```python
reveal = not url.block_bots and not url.password_hash
```
And add the missing assertion to `test_password_og_link_serves_page_to_bots` (it currently checks only status 200 + no click; the `block_bots` sibling test asserts the destination is absent — the password test should too).

infrastructure/safe_fetch.py
```diff
@@ -0,0 +1,210 @@
...
+    content_type: str
+    final_url: str
+
+
+def _is_public(ip: str) -> bool:
+    addr = ipaddress.ip_address(ip)
+    if isinstance(addr, ipaddress.IPv6Address) and addr.ipv4_mapped:
+        addr = addr.ipv4_mapped
+    return not (
+        addr.is_private
```
Zingzy: **SSRF gap: CGNAT is treated as public.** Verified on the repo venv: `ipaddress.ip_address("100.64.0.1")` → `is_private=False, is_reserved=False, is_global=False`, so this flag-union returns `True` (public) for the whole `100.64.0.0/10` range. EKS/GKE pod IPs and some cloud infra live there, so a hostname resolving to CGNAT space passes the guard (blast radius limited by the cert-pinning, so it's probing not exfil — still a hole in the stated "all addresses public" contract).

Replace the flag union with:
```python
return addr.is_global
```
(after the existing ipv4-mapped unwrap) — verified it rejects CGNAT, 6to4, Teredo, and NAT64 in one stroke while still accepting the real public test IPs, and it makes the transitional-address coverage independent of the CPython version. Add a `100.64.0.1` case to `TestIsPublic`.

infrastructure/safe_fetch.py
```diff
@@ -0,0 +1,210 @@
...
+        ip = await _resolve_public_ip(parsed.host)
+
+        # Pin the connection to the validated IP; keep name-based TLS via
+        # sni_hostname and the Host header.
+        pinned = parsed.copy_with(host=_bracket(ip))
+        async with httpx.AsyncClient(follow_redirects=False, timeout=timeout) as client:
+            request = client.build_request(
+                "GET",
+                pinned,
+                headers={"Host": parsed.host, "User-Agent": user_agent},
```
Zingzy: **gzip decompression bomb + no total deadline.** This request only overrides Host/User-Agent, so it inherits httpx's default `accept-encoding: gzip, deflate` (verified) — `aiter_bytes()` then transparently decompresses, and the `len(buf) > max_bytes` check downstream runs on *decompressed* bytes appended per socket read, so ~60MB can land in a single chunk before the cap trips (×`batch_size` in the validator). Add `"Accept-Encoding": "identity"` here.

Separately: `httpx.AsyncClient(timeout=timeout)` is per-operation, not total — the read timer resets each chunk, so a slow-drip server holds this coroutine open indefinitely under the byte cap. Wrap the streamed read in an overall `asyncio.timeout()`.

services/edge_cache/og_writethrough.py
```diff
@@ -0,0 +1,86 @@
...
+        them before the write — never for plain links, whose promoted
+        redirect entries must not be touched.
+        """
+        if url.domain != self._system_domain:
+            return
+        key = cache_key(url.domain, url.alias)
+        try:
+            if url.meta_title is not None and url.url_status == UrlStatus.ACTIVE:
+                entry = EdgeCacheEntry(type="og_only", og_html=render_meta_preview(url))
+                ok = await self._kv.put(key, entry.to_kv_json())
```
Zingzy: **No TTL + best-effort delete = a stale entry can persist indefinitely.** `put` here writes with no `expiration_ttl` (persistent), and the delete branch below swallows exceptions. The "worst case bots reach origin" reasoning is asymmetric: a failed *put* is safe, but a failed *delete* on meta-clear / deactivate / (likely) admin-block leaves a persistent entry the edge keeps serving with no TTL backstop to heal it. Combined with the password-leak blocker, a leaking or abuse-blocked link could keep unfurling its destination on social platforms.

Give `og_only` a bounded TTL re-synced on edit, or add a reconciliation/retry for failed deletes — and confirm the admin `BLOCKED` path calls `og_writethrough.remove` (I couldn't find it; admin-block isn't routed through `url_service.update()`).

Zingzy (commented):
## Follow-up: correcting my verdict — there is a blocker

I ran two deeper adversarial passes (SSRF/image pipeline; edge worker + bot classification) after my first review and verified the findings against the repo's own `.venv`. **My earlier "nothing blocking" was wrong** — there's a real password-bypass. Escalating, plus a few verified SSRF/DoS should-fixes.

### 🔴 Blocker — password-protected links leak their destination via the preview page
A link with **both** a password and `meta_tags` (nothing forbids the combination — they're independent optional fields) hands its destination to anyone sending a preview-crawler UA or a HEAD request, at **both** origin and edge:
- `build_preview_context` withholds the destination only for `block_bots`, never `password_hash`, so `long_url` + `dest_host` + the `location.replace(...)` auto-redirect are in the rendered page for a password link. (inline)
- The origin preview branch runs **before** the password gate; the edge `og_only` write-through fires for password links (it checks `meta_title` + ACTIVE only).
- `curl -A 'Twitterbot/1.0' https://spoo.me/<code>` (or `-I`, or any real unfurl) returns the destination + a JS redirect to it; a real browser correctly gets 401.
- `test_password_og_link_serves_page_to_bots` asserts only `status==200` and no click — it never checks the destination is absent, unlike its `block_bots` sibling, so the leak passes CI.

**One-line fix at the choke point:** `reveal = not url.block_bots and not url.password_hash` in `build_preview_context` (fixes origin, promotion, and write-through together), plus the missing "password omits destination" assertions on both sides.

### SSRF / DoS should-fixes (verified in venv)
- **CGNAT `100.64.0.0/10` is classified as public** — `_is_public`'s flag-union misses it (`ipaddress`: `is_private=False, is_reserved=False, is_global=False`). EKS/GKE pod IPs live there. `not addr.is_global` closes it and also covers 6to4/Teredo/NAT64 in one stroke. (inline)
- **gzip decompression bomb** — the fetcher inherits httpx's default `accept-encoding: gzip, deflate` (it only overrides Host/User-Agent), so `aiter_bytes()` decompresses and the byte cap is checked *post*-decompression — ~60MB can land in one chunk before the cap trips, ×`batch_size` in the validator. Send `Accept-Encoding: identity`. (inline)
- **No total request deadline** — httpx `timeout` is per-operation (read timer resets per chunk), so a slow-drip server holds a validator coroutine / metadata handler open indefinitely under the 1MB cap. Wrap the streamed read in an overall `asyncio.timeout()`.

### Edge lifecycle should-fixes
- **`og_only` entries have no TTL and failed deletes don't self-heal** — a failed `put` is safe (bots fall through to origin), but a failed `delete` on clear/deactivate leaves a persistent entry the edge keeps serving, with no TTL backstop. Give `og_only` a bounded TTL (re-synced on edit) or retry failed deletes. (inline)
- **Admin-block likely doesn't clear KV (verify)** — admin `BLOCKED` status isn't set through `url_service.update()`, so I couldn't find an `og_writethrough.remove` on that path. Combined with the no-TTL persistence, a link blocked for abuse could keep unfurling its card + destination on social platforms while humans get 451. Please confirm the admin-block path invalidates KV.
- **Promotion vs write-through race** — detached `promote()` (CF calls stack to ~18s) can land a `redirect` entry carrying stale `og_html` after a concurrent meta edit's write-through; no CAS between the two writers. Bounded by the promotion TTL, narrow trigger.

### Hardening / nits
- **R2 objects** are served from `og.spoo.me` (a spoo.me subdomain) with only `content-type` set — add `X-Content-Type-Options: nosniff` + `Content-Disposition: attachment` (to the SigV4-signed set) and prefer a cookieless bucket domain. Magic-byte gate + SVG rejection already make this defense-in-depth, not open XSS.
- **Upload happens before blocklist validation** — a create that fails `validate_meta_tags` has already written the image to R2 (orphan, no GC). Validate text/destination before spending the R2 write.
- **No CI guard for the bot-classification twin** — `contract-guard.yml` triggers only on `contract.py`; `preview_bots.json` byte-identity rests solely on the pytest job (the vitest side deep-equals *parsed* objects, so whitespace/order drift would pass there). Add the preview_bots paths to the guard.
- **SigV4 test** asserts prefix + determinism, not the published AWS signature — a canonicalization bug that stayed deterministic would pass. Add a real known-answer vector (the construction itself is correct).

### Confirmed good (so you can calibrate)
- **XSS is safe** — verified Starlette forces `autoescape` on `.html` for both the request path and `render.py`'s request-free render; no `|safe`; the one inline script uses `{{ long_url | tojson }}` (htmlsafe). A CSP on that response is worthwhile defense-in-depth but autoescape is the real control.
- **SSRF core is genuinely well-built** — true IP pinning (connect-to-validated-IP, SNI/cert stay on the real hostname, so `verify=True` caps the blast radius of the CGNAT gap to liveness probing, not exfil), all-records-public, per-hop redirect re-validation, CAS on every image write. The findings above are the residue, not the core.
- **Cross-language classifier is functionally equivalent** (same JSON, same substring semantics) and byte-identity is unit-tested; `block_bots` destination withholding is thorough (URL + hostname) and tested on both sides.

Net: fix the password-leak blocker before merge; the SSRF/DoS trio and the edge-lifecycle items are the next tier.

PR #235, opened 2026-07-11
feat: per-user /me API — onboarding resume, layouts, feature availability

Zingzy:
## What

The backend half of the new dashboard frontend, plus local-dev quality of life:

- **Onboarding resume** (`GET/PUT /auth/onboarding`): Dub-style Redis step cache (24h TTL) so the route-per-step wizard survives reloads and device switches; new-OAuth-user routing lands people in the flow.
- **Page layouts** (`GET/PUT/DELETE /api/v1/me/layouts/{page}`): per-user widget-board documents, stored opaquely, client-versioned.
- **Feature availability** (`GET /api/v1/me/features`): mirrors the flag service's default-deny answers as `enabled | locked | hidden` states so the frontend renders only what an account actually has. `locked` is reserved for paid plans — entitlements become a data change, not an API bump. Exposure is an explicit in-code allowlist (`EXPOSED_FEATURES`) so ops flags can never leak through the endpoint.
- **Domains DX**: create rate limit 5→10/hour (slowapi bills failed attempts too — typos and flag-gate 404s were stranding users mid-setup) and a `CUSTOM_DOMAINS_MOCK_DCV` local-dev backend that walks PENDING→ACTIVE without Cloudflare credentials (default off, logs loudly when active).

## Notes

- main merged in; flag service unions cleanly (`require()` from main + `states_for()` from this branch, one constants registry).
- `/me/features` is read-only by design — enforcement stays on the write endpoints; a lying client still hits the real gates.
- Suite: 1938 passing; the one `test_indexes` failure is pre-existing on main-side history (fails identically without this branch's commits).

## Test plan

- [x] Integration tests for all new endpoints (auth required, default-deny → all hidden, allowlist → per-user enabled, full-registry contract)
- [x] Live-verified against the dashboard frontend: allowlisted account sees gated surfaces, fresh accounts see none
- [x] Domain PENDING→ACTIVE walked locally via the DCV mock

## Summary by Sourcery

Introduce per-user onboarding state, dashboard feature availability, and page layout persistence, along with a local-dev mock for custom domain verification and adjusted domain creation limits.

New Features:
- Add GET/PUT /auth/onboarding to persist per-user onboarding progress in a soft 24h Redis cache.
- Expose per-account feature availability via GET /api/v1/me/features using a stable enabled/hidden/locked contract.
- Add GET/PUT/DELETE /api/v1/me/layouts/{page} to store opaque, client-owned dashboard layout documents per user and page.

Enhancements:
- Wire a page layout repository/service and Mongo indexes for per-(user, page) layout storage.
- Redirect brand-new OAuth users to the onboarding flow when no explicit next URL is provided.
- Add OpenAPI grouping and rate limits for the new /me and onboarding endpoints.

Tests:
- Add integration tests covering /api/v1/me/features and per-user page layout endpoints, including auth and validation behavior.

Chores:
- Add a MockDcvBackend and configuration flag to simulate custom domain verification in local development, and increase the custom domain creation rate limit to 10/hour.

## Summary by CodeRabbit

* **New Features**
  * Added authenticated `/api/v1/me/layouts/{page}` endpoints to save, fetch, and reset per-user dashboard layouts (pages limited to Analytics/Overview).
  * Added `/api/v1/me/features` endpoint to surface feature availability.
  * Added onboarding resume/complete endpoints with a durable onboarding timestamp.
  * OAuth sign-in now routes new accounts to onboarding when no explicit destination is provided.
  * Added a local-dev option to mock DCV verification.

* **Improvements**
  * Tightened layout request validation and updated rate limits.
  * Prevented mock DCV usage in production.
  * Expanded OpenAPI tagging with a new “Me” tag.


routes/auth/onboarding.py
```diff
@@ -0,0 +1,120 @@
...
+        return OnboardingStateResponse()
+    try:
+        data = json.loads(raw)
+    except (TypeError, ValueError):
+        return OnboardingStateResponse()
+    step = data.get("step")
+    return OnboardingStateResponse(
+        step=step,
+        path=data.get("path"),
+        completed=step == "completed",
```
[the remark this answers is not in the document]
Zingzy: Intentional: `done` is the recap *page* — a resumable step the user can sit on — while `completed` is the terminal marker written only when they finish the flow. Reaching the recap without clicking finish must not read as completed, or the dashboard gate would stop routing them back in. Documented at the flag site in 830dd9e.

Zingzy (commented):
## Review: per-user `/me` API (onboarding, layouts, feature availability)

**Clean, low-risk, mergeable — a couple of decisions to make consciously, none blocking.** No external fetching, no new attack surface, everything authed and flag-aware, and it follows the house patterns closely (layered repo→service→route, `Annotated` service DIs, flag-registry constants, compound unique index, typed DTOs). CI green across 3.10–3.14. Findings below are two design decisions, one placement nit, and small stuff; the rest is genuinely good.

### Should-decide before merge
1. **Layout `page` is an open namespace with no per-user cap** — arbitrary `^[a-z0-9_-]{1,32}$` and no count limit, while every other per-user resource is quota'd (`max_active_keys`, `max_per_user`). Restrict to a known-page allowlist (caps storage + rejects junk). (inline)
2. **`/api/v1/me/*` skips the scope-gating the rest of `/api/v1` uses** — bare `AuthUser` vs `require_scopes(...)`. Decide JWT-only or scoped. (inline) — note: `/auth/onboarding`'s bare `AuthUser` is *consistent* with `GET /auth/me`, so that half is fine as-is; this is specifically about the `/api/v1/me/*` endpoints.

### Placement nit
3. **Onboarding state is raw Redis + inline JSON in the route** — key scheme, TTL, serialization, None-tolerance and error-swallowing all live in the handler, where every other Redis-backed state is a dedicated `infrastructure/cache/` class (`url_cache`, `feature_flag_cache`, `meta_fetch_cache`). Extract an `OnboardingCache`. (inline)

### Nits
- `states_for` awaits `is_enabled` sequentially in the comprehension — `asyncio.gather` on a dashboard-load path. (inline)
- `PageLayoutRepository.upsert` reaches into `self._col.update_one` with its own error handling while `get`/`delete` use base helpers (base has no upsert helper, so justified — noting the asymmetry).
- Mock-DCV boot-guard is production-only; `staging` with `mock_dcv` would still let anyone claim domains (staging isn't public, so low).
- `AB_TESTING_FLAG` is exposed in `/me/features` with no endpoint enforcing it yet — contract runs slightly ahead of the enforcement surface (intentional/forward-looking).
- Layouts are last-write-wins, no version/ETag — two devices silently clobber (PR notes client-versioned, so acceptable).

### Notably good
- **`EXPOSED_FEATURES` allowlist** so ops flags can't leak through `/me/features`, pinned by the full-registry contract test — the important safety property here.
- **`FeatureState` three-state contract with `LOCKED` reserved** so paid entitlements become a data change, not an API bump — right for where plans are heading.
- **Mock DCV that fails to boot in production** (not a warning) — and it's placed in `AppSettings._populate_sub_configs_and_secret`, which is exactly the right spot: there was no existing precedent for a dev-flag-in-prod guard, `CustomDomainSettings` can't see `env` on its own, and this is the cleaner of the two sane options.
- Ownership-by-`user_id` in every query + compound unique `(user_id, page)` index (no IDOR); soft onboarding degradation (Redis down/expired → empty → client localStorage); documented `DOMAIN_CREATE` 5→10 rationale (slowapi bills failed attempts); OAuth `is_new → /onboarding` routing with the device-consent deep-link carve-out (verified `is_new` is set only on genuine account creation).

### Merge safety
Same as prior merges: deploy is release-gated not push-gated, all three surfaces are net-new endpoints (hot path untouched), and the mock DCV physically can't boot in prod. Onboarding degrades softly, so a Redis outage can't break the flow.

Decide #1 and #2, extract the onboarding cache (#3), and the rest are fast-follows.

routes/api_v1/me.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    return FeaturesResponse(features=await flag_service.states_for(user))
+
+
+PagePath = Annotated[
+    str,
+    Path(
+        min_length=1,
+        max_length=32,
+        pattern=r"^[a-z0-9_-]+$",
```
Zingzy: **Open namespace with no per-user cap (should-decide).** `page` accepts any `^[a-z0-9_-]{1,32}$`, and neither the service nor the repo bounds how many distinct page docs one user can create. Every other per-user resource in the codebase is quota'd — API keys have `max_active_keys`, custom domains have `max_per_user` — but layouts have neither a count cap nor a known-page allowlist. So an authenticated user can accumulate unbounded distinct page docs (rate-limited to 60/min × 32 KiB, but unbounded over time) and pollute the namespace with junk slugs.

The frontend has a fixed set of dashboard pages, so restricting `page` to an allowlist of those slugs is the clean fix — it caps storage *and* rejects garbage in one move. Low urgency (authed + rate-limited), but it's the one missing quota relative to the rest of the codebase.

routes/api_v1/me.py
```diff
@@ -0,0 +1,134 @@
...
+@router.get(
+    "/features",
+    responses=AUTH_RESPONSES,
+    operation_id="getMyFeatures",
+    summary="Get Feature Availability",
+)
+@limiter.limit(Limits.DASHBOARD_READ)
+async def get_my_features(
+    request: Request,
+    user: AuthUser,
```
Zingzy: **`/api/v1/me/*` skips the scope-gating the rest of `/api/v1` uses (should-decide).** `management.py`/`metadata.py`/etc. gate on `require_scopes(URL_MANAGEMENT_SCOPES / URL_READ_SCOPES)`; these `/me` endpoints use bare `AuthUser` (= `require_auth`, no scope). So any valid API key — including a `urls:read`-only key — can read/write dashboard layouts and enumerate feature availability. It's not an IDOR (a key is still scoped to its own user) and the data is low-sensitivity, but it's a real consistency deviation.

These are dashboard-session concerns, so I'd lean JWT-only (`require_jwt`) rather than accepting arbitrary API keys — `keys.py` already refuses API keys on principle ("API keys must not manage API keys"), and "an API key shouldn't manage dashboard layouts" is the same shape. Alternatively give them a read scope so the choice is explicit. (Note: `/auth/onboarding`'s bare `AuthUser` is fine — it matches `GET /auth/me`.)

routes/auth/onboarding.py
```diff
@@ -0,0 +1,124 @@
...
+log = get_logger(__name__)
+
+router = APIRouter()
+
+RedisDep = Annotated[Any, Depends(get_redis)]
+
+ONBOARDING_TTL_SECONDS = 24 * 60 * 60  # Dub's ONBOARDING_WINDOW: soft 24h gate
+
+
+def _key(user_id: str) -> str:
```
Zingzy: **Raw Redis + inline JSON in the route (placement nit).** This handler owns the key scheme (`_key`), the TTL constant, `json.dumps`/`loads`, the None-Redis tolerance, and the error-swallowing — all inline. Every other Redis-backed state in the codebase is a dedicated class in `infrastructure/cache/` (`url_cache`, `feature_flag_cache`, `meta_fetch_cache`), each with exactly this shape: `redis_client: Redis | None`, a `_key()`, get/set, TTL, swallowed errors.

Extracting an `OnboardingCache` (constructed in `wire_services`, stored on `app.state` like `meta_fetch_cache`) would match the pattern and move the soft-degradation logic out of the HTTP layer. Behavior is fine — this is purely the placement that the next Redis feature will copy from.

services/feature_flag_service.py
```diff
@@ -124,6 +138,27 @@ async def is_enabled(self, name: str, user: CurrentUser | None) -> bool:
...
+        Inherits ``is_enabled``'s default-deny: unregistered flags, cache
+        and repo failures all read as HIDDEN.
+        """
+        return {
+            name: (
+                FeatureState.ENABLED
+                if await self.is_enabled(name, user)
+                else FeatureState.HIDDEN
+            )
+            for name in EXPOSED_FEATURES
```
Zingzy: **Sequential awaits (nit).** `is_enabled` is awaited one-at-a-time in the comprehension, so this is N sequential cache lookups (4 today) on a dashboard-load path. `asyncio.gather` over `EXPOSED_FEATURES` would parallelize it:
```python
states = await asyncio.gather(*(self.is_enabled(n, user) for n in EXPOSED_FEATURES))
return {n: FeatureState.ENABLED if ok else FeatureState.HIDDEN for n, ok in zip(EXPOSED_FEATURES, states)}
```
Tiny (they're cache-backed), just noting it's on the hot dashboard path.

routes/api_v1/me.py
Zingzy: Fixed in b8cb58e: `page` is now a closed allowlist (`analytics`, `overview`) via `Literal` — unknown slugs 422, capping both storage and the namespace. Tests cover the rejection and the allowlisted pages.

routes/api_v1/me.py
Zingzy: Fixed in b8cb58e: all `/api/v1/me/*` endpoints are `JwtUser` now — dashboard-session state, API keys refused on the key-management principle. `/auth/onboarding` stays bare `AuthUser` per your note.

routes/auth/onboarding.py
Zingzy: Fixed in b8cb58e: extracted `infrastructure/cache/onboarding_cache.py` (same shape as meta_fetch_cache — None-redis tolerant, \_key, TTL, swallowed errors), constructed in `wire_services`, injected via `OnboardingCacheDep`. The route is thin HTTP now. (The same commit also splits the resume pointer from completion — `onboarded_at` on the user doc — per the follow-up design discussion.)

services/feature_flag_service.py
Zingzy: Fixed in b8cb58e: `asyncio.gather` over EXPOSED_FEATURES with a strict zip.

schemas/models/user.py
```diff
@@ -110,4 +110,10 @@ class UserDoc(MongoBaseModel):
     created_at: datetime | None = None
     updated_at: datetime | None = None
     last_login_at: datetime | None = None
+    # Permanent completion fact — the onboarding wizard's resume pointer is
+    # ephemeral (Redis, 24h), but "this account finished onboarding" must
+    # survive it. Null = never finished.
+    onboarded_at: datetime | None = None
```
Zingzy, line 116: **[Should-decide] Every pre-existing account reads `onboarded_at = null`.** The `None` default is correct for backward-compat (old docs decode fine), but on deploy your entire existing user base looks "never onboarded." The backend won't force them in — the OAuth `is_new → /onboarding` routing only fires for brand-new accounts — but this field exists *so the frontend can gate on it*, and if the frontend routes on `onboarded_at == null`, it'll sweep every existing user into the wizard.

**Blast radius:** potentially the whole existing base, one-time, reversible. **Decide:** either a one-time backfill (`onboarded_at = created_at` for accounts predating onboarding) or confirm the frontend keys on `is_new`, not raw `onboarded_at`. Worth settling before the flow goes live so nobody gets nagged into re-onboarding.

routes/auth/onboarding.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Stamps `onboarded_at` on the user document (first completion wins —
+    repeat calls are idempotent and keep the original timestamp), records
+    the optional HDYHAU answer, and drops the resume pointer.
+
+    **Authentication**: Required (JWT or API key)
+
+    **Rate Limits**: 30/min
+    """
+    user_id = ObjectId(str(user.user_id))
```
Zingzy: **[Nit] Redundant round-trip.** `user.user_id` is already an `ObjectId` (`CurrentUser.user_id: ObjectId`), so `ObjectId(str(user.user_id))` converts it to a string and back. Every other route (e.g. the layout handlers) passes `user.user_id` straight through. Just use `user.user_id`.

routes/api_v1/me.py
```diff
@@ -0,0 +1,133 @@
...
+    enforcement — the write endpoints enforce the same gates server-side.
+
+    **Authentication**: Required.
+    """
+    return FeaturesResponse(features=await flag_service.states_for(user))
+
+
+# Closed set: every dashboard board the frontend actually renders. An
+# allowlist (not just a pattern) caps per-user storage and rejects junk
+# slugs — layouts were the one per-user resource with no quota. Grows in
```
Zingzy: **[Nit] Comment carries fix-narration.** The kernel is a good constraint — closed set of dashboard boards, an allowlist caps per-user storage, grows with the frontend's boards. But "layouts *were* the one per-user resource with no quota" narrates the history of the fix (reviewer/changelog talk), not something the next reader editing this line needs. Trim that clause; keep the why (`Literal` not a pattern → storage cap + no junk slugs).

tests/integration/api_v1/test_me_features.py
```diff
@@ -0,0 +1,83 @@
...
+    return _build_test_app(
+        {
+            require_jwt: lambda: user,
+            get_current_user: lambda: user,
+            get_feature_flag_service: lambda: _flag_svc(enabled_names),
+        }
+    )
+
+
+def test_requires_auth():
```
Zingzy, line 49: **[Nit] Test gap for the JWT-only switch.** `test_requires_auth` covers no-auth → 401, but nothing asserts the actual new behavior: an *API-key*-authed request to `/me/*` now gets 403 (`require_jwt` rejects `api_key_doc is not None`). One case pinning "API key → 403" would lock in the `AuthUser → JwtUser` change so a future revert can't pass silently.

.github/workflows/image.yml
```diff
@@ -13,6 +17,7 @@ on:
       - '**/*.md'
       - '**/*.rst'
       - 'docs/**'
+  workflow_dispatch:
```
Zingzy, line 20: **[Build — info/optional] `workflow_dispatch` now exposes the GHCR PAT to branch builds.** The core change is sound — the `enable=${{ github.event_name == 'push' }}` guard means a branch dispatch can't move `edge`, and deploy is release-gated, so branch images can't reach prod. But this trigger lets any write-access actor build *any ref*, and the job holds `secrets.GHCR_PAT` + `packages: write` while running a **branch-controlled Dockerfile/build**. Solo, that's fine; the day a collaborator exists it's a supply-chain surface (a hostile branch could exfiltrate the PAT or push a poisoned `:<sha>`).

Two low-effort hardenings: prefer the built-in `GITHUB_TOKEN` over a long-lived PAT for GHCR login (with `packages: write` it suffices for the same-repo package), and — if you want defense-in-depth — SHA-pin the `docker/*` actions. Neither blocks merge.

Zingzy (commented):
## Re-review after the `b8cb58e` rework — all prior findings closed

Every item from the last review is fixed at the choke point, and the pointer/completion split is a real improvement (it fixes a latent bug: completion used to live only in the 24h Redis cache, so an expired pointer would silently re-onboard a finished user). Verified: ran the affected suites in a worktree (56 passed), CI green 3.10–3.14, `require_jwt` genuinely 403s API keys, completion uses a CAS filter (`{onboarded_at: None}`) so first-completion-wins is concurrency-safe.

What's left is one decision and a few nits — inline. (Re-checked and *dropped* the `success: bool` "always true" nit: `LogoutResponse` already uses that convention, so `OnboardingCompleteResponse` is consistent, not a smell.)

**Lower-priority, not inlined** (out of this PR's scope, noting so they're captured): no GHCR retention/GC so `:<sha>` tags accumulate unbounded; third-party actions are major-tag-pinned rather than SHA-pinned; and `image.yml`'s comment overstates the cache-warming benefit of branch dispatch (GitHub scopes the Actions cache per-branch, so a branch build warms the branch scope, not main's release cache).

Net: decide the `onboarded_at` backfill before enabling the flow; the merge itself is safe.

PR #237, opened 2026-07-12
feat: serve Next frontend on beta.spoo.me

Zingzy:
Staging rehearsal of the frontend cutover: adds the `next-beta` service to the beta compose slice and the `@next` path matcher to the beta Caddy vhost. Prod blocks are untouched.

## What
- `next-beta`: `ghcr.io/spoo-me/frontend:${NEXT_IMAGE_TAG:-edge}`, `SPOO_API_URL=http://spoo_app_beta:8000` inline (SSR fetches + dormant rewrites stay on docker DNS), healthcheck on `/api/health`, no env file needed.
- Caddy `beta.spoo.me`: page paths (marketing, auth, onboarding, dashboard, `/_next/*`, `/relay/*`) → `spoo_next_beta:3000`; everything else (short codes, `/auth/*`, `/oauth/*`, `/api/v1/*`, `/static/*`) falls through to `spoo_app_beta:8000` as today.
- Caddyfile parse-checked with `caddy fmt`.

## Deploy (manual, beta compose project)
```bash
IMAGE_TAG=0963092 docker compose -p spoo-beta -f docker-compose.beta.yml --env-file .env up -d
docker exec spoo_caddy caddy reload --config /etc/caddy/Caddyfile
```
Note: backend image `0963092` predates the `ONBOARDING_REDIRECT_ENABLED` flag (redirect always on — desired on beta). Images built from the current feat/onboarding tip need the env var in `.env.beta`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Introduce a beta Next.js frontend service and route selected beta.spoo.me paths through it while leaving existing backend routing and production configuration unchanged.

New Features:
- Add next-beta frontend service to the beta Docker Compose stack targeting the beta backend over internal Docker networking.

Deployment:
- Update beta Caddy virtual host routing so marketing and app page paths are served by the Next.js frontend while non-page traffic continues to hit the beta backend directly.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated beta frontend for selected Next.js routes, static/public assets, Next runtime paths, and short preview URLs.
  * Added health monitoring for the new beta frontend service.
* **Improvements**
  * Updated beta request routing to use the dedicated beta frontend for matched requests, while other traffic continues to the main application.
  * Improved HTML error handling for key HTTP responses by returning the dedicated error page content while preserving the original status code.
  * Updated edge short-code eligibility to exclude additional reserved paths and preview-intent `+` codes.
* **Tests**
  * Expanded routing and eligibility coverage for the new edge-cache rules.


caddy/Caddyfile
```diff
@@ -80,9 +110,88 @@ spoo.me, www.spoo.me {
...
+	handle @robots {
+		header Content-Type "text/plain; charset=utf-8"
+		respond "User-agent: *
+Disallow: /" 200
+	}
+
+	# Only paths with a built Next page belong here — a listed path
+	# with no page behind it 404s in Next instead of falling through
+	# to FastAPI.
+	@next {
```
Zingzy, line 127: **Should-verify before relying on beta: pre-reservation aliases.**

The reserved list protects creations going forward, but any link created BEFORE its path was added to `reserved_aliases.py` still exists in Mongo — and this matcher silently shadows it on beta (Caddy hands its path to Next; the link keeps resolving on prod, 404s on beta, and would break for real at prod cutover). Entries like `testimonials`, `icons-3d`, `humans`, or `brand` are recent enough that a nine-year-old legacy link or an early v2 link could plausibly own one.

One query closes the question: count docs across `urlsV2`, legacy `urls`, and `emojis` whose alias (case-insensitive) is in `RESERVED_ALIASES`. Zero means this whole class of problem is empty and the matcher is safe at prod cutover too; nonzero means those links need migration or grandfathering before the flip. Worth a line in the prod-cutover checklist §0 either way, since the same matcher is what gets ported.

caddy/Caddyfile
```diff
@@ -67,6 +67,36 @@
 	}
 }
 
+# Compose one branded error page (beta): when the origin self-describes
+# an error (X-Error-Code) and the client meaningfully asked for HTML,
+# swap the body for the Next error page while KEEPING the origin status
+# (the inner proxy would otherwise answer with Next's 200). JSON clients
+# and Accept:*/* scanner probes get the origin response untouched —
+# composition costs a Next render. args[0] = the status code.
+(compose_beta_err) {
```
Zingzy, line 76: Should-decide, mostly for the prod port: this couples every HTML error response to Next availability. Unknown-short-code 404s are the hottest error path on the site, and with this snippet each one (for HTML clients) takes a synchronous Next render — if `spoo_next_beta` is down or overloaded, branded 404s become 502s from the inner proxy instead of degrading to the origin body. Fine to accept on beta; at prod scale it deserves a conscious decision (accept the coupling, or note that Caddy has no clean circuit-breaker here and the `handle` fallback below only covers non-HTML).

Smaller, same snippet: the composed response replaces origin headers, so a 429's `Retry-After` (and any rate-limit headers) are dropped for HTML clients. Probably nothing depends on it for browser traffic, but worth knowing it is lossy.

Zingzy (commented):
Mergeable, nothing blocking. This is ops config plus edge eligibility, so I reviewed it against a different bar than the API PRs: collision surfaces, prod blast radius of the edge changes, and post-merge staleness (the branch predates the whole emoji/bulk cycle). All three hold up; two things to decide or verify before leaning on beta results, inline.

## What I verified hands-on

- **Every single-segment page path in `@next`/`@next_get` is in `shared/reserved_aliases.py`** — checked entry by entry against main. `EXCLUDED_EXACT` in the worker mirrors the Python list exactly, and its drift comment prices the failure mode correctly (one wasted KV read, never a wrong answer). New short codes cannot collide with the Next page namespace.
- **The edge changes are prod-safe today, not just on beta.** They ship to the production worker at the next edge deploy, so I checked them as prod changes: `EXCLUDED_EXACT` entries are unclaimable (reserved), so skipping KV loses nothing; `+`-suffixed keys never exist in KV, so the preview exclusion converts a guaranteed miss into a skipped read. Both are pure wins on spoo.me regardless of the beta rollout.
- **No rebase hazard from the merged cycle**: only `contract/fixtures.json` moved on main among this PR's files (the #251 skin-tone fixture). I ran the edge suite with this branch's eligibility code against main's fixtures — 76 green, and the emoji fixture that postdates this branch passes, so percent-encoded emoji paths stay eligible under the new exclusions.
- **The compose healthcheck actually works**: the frontend image is `node:22-alpine`, so busybox `wget` is present.
- **Caddy semantics check out**: `handle` ordering beats the proxies for robots.txt, same-type `reverse_proxy` directives match in source order (so `@next` → `@next_get` → `@preview` → fallthrough is the evaluation order), `replace_status` with a named response matcher is valid 2.5+ syntax, and `caddy reload` fails atomically so a bad config never takes the running one down.

## What is genuinely good here

The `@next_get` split is the kind of thing that only exists because someone got burned: `/`, `/contact`, `/report`, `/stats/*` all carry shipped POST contracts, and an unscoped matcher would have Next answering 200 to every legacy API client. The comment says it was caught twice on beta — encoding that as a matcher with the rationale attached is exactly right. Same for the `beta.spoo.me` network alias: the undici-cannot-override-Host constraint is real, and solving tenant scoping by making docker DNS produce the right Host header is the clean fix, documented at both the alias and the consumer.

## Enable-time tail (not merge gates)

Merging ships nothing (Caddy and compose apply on manual deploy; the worker on the next edge release). Before trusting beta as a cutover rehearsal, the routing matrix deserves one scripted pass: a page path, a short-code redirect, `POST /` anonymous shorten, `POST /emoji` legacy, `GET /stats/{code}`, a `/{code}+` preview including percent-encoded emoji, and a composed 404 for HTML vs untouched JSON 404. The PR body verifies `caddy fmt` only, which is a syntax check — `caddy validate` (or just the reload's built-in validation) plus that matrix is the real proof.

PR #239, opened 2026-07-12
feat: X-Error-Code header + edge-composed error bodies

Zingzy:
The backend half of the error-pages system: HTML error responses self-describe so the edge can compose branded Next pages, while self-hosters keep today's behavior byte-for-byte. Companion frontend PR: spoo-me/frontend#5. Caddy wiring lands separately after the beta handle_response spike.

## What

- `X-Error-Code: <error_code slug>` on every HTML error response, on BOTH render paths:
  - `middleware/error_handler.py::_error_html` (AppError → its slug; slowapi 429 → `rate_limit_exceeded`; catch-all → `internal_error`)
  - `routes/redirect_routes.py::_error_page` system-default branch (`not_found` / `gone` / `blocked` / `forbidden`) — the hot path renders its own errors and never reaches the app-level handler, so it needs its own treatment. Tenant branch untouched.
- `EDGE_COMPOSED_ERRORS` (new AppSettings bool, default **false**): when on, intercepted statuses ({404, 410, 429, 451, 500} app-level; {404, 410, 451} hot path) return an empty body + header on GET/HEAD — Caddy discards the body anyway, so the template render is skipped. 403s (bot block, password) always keep their bodies.
- JSON error contract untouched, byte-for-byte. `_wants_json` untouched.
- The 400 password-form path carries no slug today and gets no header (one dict line when needed).

## Tests

75 pass across the four touched files; full integration + smoke sweep 522 pass (includes the tenant and custom-domain error paths). New coverage: header slugs on every HTML error family, flag-on empty bodies for 404/500 and the hot-path 404, 403 body retention, POST body retention, JSON untouched, config default.

## Not this PR

- Caddy `handle_response` blocks (beta spike first — placeholder syntax + status preservation verified empirically before prod).
- Splitting the `blocked` slug into safety-vs-takedown (the 451 page copy keys on the slug, so finer wordings come free later).
- Accept-header gating for `*/*` scanner probes (lives in Caddy's matcher, not here).

## Summary by Sourcery

Introduce machine-readable X-Error-Code headers on HTML error responses and add an edge_composed_errors flag to allow edge infrastructure to compose branded error pages while preserving existing JSON contracts and default behavior.

New Features:
- Expose X-Error-Code slugs on all HTML error responses from both application and redirect error paths.
- Add the EDGE_COMPOSED_ERRORS configuration flag to allow intercepted HTML error statuses to return empty bodies for edge-composed error pages while keeping 403 and non-GET/HEAD responses rendered.

Tests:
- Extend integration, redirect, rate-limiting, and config smoke tests to cover X-Error-Code headers, edge-composed empty-body behavior, JSON error invariants, and the new configuration default.

## Summary by CodeRabbit

- **New Features**
  - Added machine-readable `X-Error-Code` headers to relevant HTML error responses (e.g., not found, forbidden, rate limit exceeded, gone/blocked, and server errors).
  - Introduced an optional edge-composed error mode for selected statuses that returns an empty body on GET/HEAD while still sending `X-Error-Code`.
  - Preserved the existing branded HTML error-page behavior by default, and kept JSON API error responses unchanged.
- **Tests**
  - Expanded integration and smoke test coverage for error handling, redirects, and rate limiting.

### DIFF
--- a/middleware/error_handler.py
+++ b/middleware/error_handler.py
@@ -69,8 +69,13 @@ def _wants_json(request: Request) -> bool:
 
 
 # Statuses the edge intercepts when EDGE_COMPOSED_ERRORS is on — must stay
-# in sync with the Caddy handle_response matcher and routes/redirect_routes.py.
-_EDGE_INTERCEPTED_STATUSES = {404, 410, 429, 451, 500}
+# in sync with the Caddy handle_response matcher. Two intentionally
+# different sets: the app-level handlers below hand every intercepted
+# status to the edge, while the redirect hot path
+# (routes/redirect_routes.py) only intercepts resolve failures — its 403
+# bot-block page always keeps its server-rendered body.
+EDGE_INTERCEPTED_STATUSES = {404, 410, 429, 451, 500}
+REDIRECT_EDGE_INTERCEPTED_STATUSES = {404, 410, 451}
 
 
 def _error_html(
@@ -87,7 +92,7 @@ def _error_html(
     settings = getattr(request.app.state, "settings", None)
     if (
         getattr(settings, "edge_composed_errors", False)
-        and status_code in _EDGE_INTERCEPTED_STATUSES
+        and status_code in EDGE_INTERCEPTED_STATUSES
         and request.method in {"GET", "HEAD"}
     ):
         return Response(status_code=status_code, headers={"X-Error-Code": slug})
--- a/middleware/error_handler.py
+++ b/middleware/error_handler.py
@@ -68,12 +68,10 @@ def _wants_json(request: Request) -> bool:
     )
 
 
-# Statuses the edge intercepts when EDGE_COMPOSED_ERRORS is on — must stay
-# in sync with the Caddy handle_response matcher. Two intentionally
-# different sets: the app-level handlers below hand every intercepted
-# status to the edge, while the redirect hot path
-# (routes/redirect_routes.py) only intercepts resolve failures — its 403
-# bot-block page always keeps its server-rendered body.
+# Statuses that return EMPTY bodies under EDGE_COMPOSED_ERRORS — must stay
+# a subset of what Caddy's handle_response matcher composes, or visitors
+# get blank pages (verify at cutover; redirect ⊆ app-level is test-pinned).
+# The redirect set is smaller on purpose: its 403 bot-block keeps its body.
 EDGE_INTERCEPTED_STATUSES = {404, 410, 429, 451, 500}
 REDIRECT_EDGE_INTERCEPTED_STATUSES = {404, 410, 451}
 
@@ -88,7 +86,8 @@ def _error_html(
     the template entirely — Caddy throws the body away and composes the Next
     error page, so rendering it would be wasted work.
     """
-    # getattr-with-default: tests build bare apps without lifespan state.
+    # settings can be absent (app built without lifespan) — default-off is
+    # the fail-safe.
     settings = getattr(request.app.state, "settings", None)
     if (
         getattr(settings, "edge_composed_errors", False)
--- a/middleware/error_handler.py
+++ b/middleware/error_handler.py
@@ -68,8 +68,30 @@ def _wants_json(request: Request) -> bool:
     )
 
 
-def _error_html(request: Request, status_code: int, message: str) -> Response:
-    """Render error.html template for browser requests."""
+# Statuses the edge intercepts when EDGE_COMPOSED_ERRORS is on — must stay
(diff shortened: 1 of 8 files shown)


middleware/error_handler.py
```diff
@@ -68,8 +68,30 @@ def _wants_json(request: Request) -> bool:
     )
 
 
-def _error_html(request: Request, status_code: int, message: str) -> Response:
-    """Render error.html template for browser requests."""
+# Statuses the edge intercepts when EDGE_COMPOSED_ERRORS is on — must stay
+# in sync with the Caddy handle_response matcher and routes/redirect_routes.py.
+_EDGE_INTERCEPTED_STATUSES = {404, 410, 429, 451, 500}
```
Zingzy: **[Should-decide] Intercepted-status set is a triple source of truth with an unenforced contract.** This set, `_EDGE_INTERCEPTED_CODES` in `routes/redirect_routes.py:58`, and — landing later — Caddy's `handle_response` matcher must all agree. The comment flags it, which is right, but the failure mode is user-facing and worth naming precisely: if the origin returns an **empty body for a status Caddy does NOT compose, the visitor gets a blank page.** So the origin-empty set must be a **subset** of what Caddy composes, and nothing here enforces that (Caddy config lives elsewhere).

**Blast radius:** system-domain 404/410/451/429/500, but only when the flag is on — our-deployment-only, flipped manually at cutover after the beta spike — so the sequencing mitigates it. It becomes real the day someone adds a status to one Python set and forgets Caddy. **Before enabling:** verify empirically that the union of the two Python sets is composed by Caddy, and consider either a shared constant or a test asserting the hot-path codes ⊆ this set so the two *Python* halves can't drift silently (the Python↔Caddy gap stays a runbook item).

middleware/error_handler.py
```diff
@@ -68,8 +68,30 @@ def _wants_json(request: Request) -> bool:
...
+    request: Request, status_code: int, message: str, slug: str
+) -> Response:
+    """Render error.html template for browser requests.
+
+    Every HTML error self-describes via ``X-Error-Code`` so the edge can
+    route on it. With ``edge_composed_errors`` on, intercepted statuses skip
+    the template entirely — Caddy throws the body away and composes the Next
+    error page, so rendering it would be wasted work.
+    """
+    # getattr-with-default: tests build bare apps without lifespan state.
```
Zingzy: **[Nit] Reword to the invariant, not the test.** This same comment is duplicated verbatim in `routes/redirect_routes.py:102`, and both phrase it around the test harness. The constraint it protects is real and worth keeping — don't "simplify" the `getattr` chain to a direct `app.state.settings.edge_composed_errors`, because `settings` can be absent when the app is built without the lifespan, and the default-`False` is the fail-safe that keeps edge-composition *off* if settings is ever missing. Phrase it as that (e.g. `# settings may be unset when the app is built without the lifespan; default-off is the fail-safe`) rather than "tests build bare apps" — the next reader cares that settings can be missing, not that a test exercises it.

Zingzy (commented):
## Review: X-Error-Code + edge-composed error bodies

**Mergeable, no blockers.** Small and disciplined: flag-off it's byte-for-byte today's behavior plus one additive header (JSON path untouched, verified), and flag-on is our-deployment-only behind a Caddy cutover that hasn't shipped. Both render paths handled correctly and separately, 403-keeps-body and tenant-untouched carve-outs are right, placement matches the sibling toggle (`onboarding_redirect_enabled`). CI green 3.10–3.14; ran the four touched suites in a worktree (75 passed); confirmed slugs are static class attrs so there's no `X-Error-Code` injection surface and no new info leak (slug is already implied by the status).

One decision + one wording nit inline. **Merge-safety sequencing:** don't flip `EDGE_COMPOSED_ERRORS=true` until Caddy is actually composing the intercepted set, or the origin hands out blank pages — the default-false + manual-cutover design already routes you through that gate, just naming it.

**Lower-priority (not inlined):** `EMAIL_NOT_VERIFIED` is the lone SCREAMING_CASE slug in an otherwise snake_case taxonomy (`errors.py`, pre-existing, and it's a 403 so never edge-composed) — worth normalizing if the Caddy matcher/Next map ends up keying on it. The 400 password-form path staying slug-less is fine — you scoped it out and it's a POST anyway.

**What's notably good:** the drift-warning comment on the status set is the *right* kind of constraint comment; asserting the schema *default* (not the resolved value) in the config test so a local `.env` can't flip it; and the coverage (empty-body for both an app-level status and the hot-path 404, 403/POST keep-body, JSON-unaffected).

PR #240, opened 2026-07-12
feat: public link preview endpoint

Zingzy:
## Summary

Implements `GET /api/v1/public/preview/{short_code}` — the backend wire for the already-built link preview page (`/{code}+`, spoo-landing PR #2). The frontend contract is frozen: the spoo-landing mock (`app/api/mock/[...path]/public-preview.ts`) IS the contract, consumed via `lib/api/public-preview.ts`, and this endpoint matches it byte-for-byte.

Core safety semantics:

- **Status-agnostic resolution** — expired / inactive / blocked links still answer 200 with their status stated; only truly missing codes 404. Resolution mirrors the redirect's dispatch heuristic (emoji → v1; 6 chars → v1 then v2; otherwise v2 then v1), scoped to the system default domain, exact-match (case-sensitive), read-only.
- **Destination-only-while-active** — `destination` and `geo_destinations` ride the wire only while the link is active AND not password-protected. Password, expiry (including derived expiry the click path hasn't persisted yet), pause and block all withhold them: the preview never reveals more than the redirect would. There is no password unlock on this wire.
- **Anti-cloaking transparency** — geo-targeted links list EVERY rule, grouped by destination URL with countries sorted; `destination` stays the unmatched-country fallback.
- **Owner-set `meta_tags` never appear** — the preview shows our resolved facts only; sender-controlled content stays off the safety surface. No stats payload; `private_stats` plays no role here (preview ≠ stats).

The endpoint ships dark: nothing serves it until the frontend page + Caddy `/{code}+` matcher are wired (gated on the handle_response spike).

## Changes

- `routes/api_v1/public_preview.py` — new route, no auth dependency (the response is identical for everyone, owners included), own rate limit.
- `services/public_preview_service.py` — resolution + derivation. Deliberately does NOT use `UrlService.resolve` (it raises on non-active statuses and its cache shape lacks `created_at`); resolves raw docs per generation instead. v1/emoji docs are fetched as raw dicts (typed `LegacyUrlDoc` drops `creation-date`/`creation-time`, which `created_at` needs). `_split_destination` is lifted from the legacy Jinja preview with a pointer comment.
- `schemas/dto/responses/public_preview.py` — `PreviewDestination`, `PreviewGeoDestination`, `PublicPreviewResponse` (serialize only).
- `middleware/rate_limiter.py` — new `Limits.PUBLIC_PREVIEW = "30 per minute; 2000 per day"`: like the redirect, this endpoint is an existence oracle by design, so it gets its own bounded budget instead of the anon API tier.
- `routes/api_v1/__init__.py`, `dependencies/wiring.py`, `dependencies/services.py` — one-line include + standard singleton wiring/getter.

## Tests

- New `tests/integration/api_v1/test_public_preview.py` (22 tests, all mocked): 404 body; full-body v2 active assertion; destination splitting (bare host → empty path, querystring/fragment preserved, http vs https); geo grouping/sorting; password withholding (v2 hash + v1 plaintext, secrets never leak); persisted EXPIRED/INACTIVE/BLOCKED lowercased + withheld; derived expiry (past `expire_after`, max-clicks reached) for both generations; naive v1 expiration never expires (legacy parity); v1 `created_at` combination + null for ancient rows; percent-encoded emoji alias; `private_stats` irrelevance; `meta_tags` absent from the wire; case-sensitive lookups; v1-vs-v2 resolution order for 6- and 7-char codes.
- Targeted run: 22 passed. Full sweep: 2114 passed.

## Summary by Sourcery

Implement a read-only public link preview surface that resolves short codes across generations and returns a sanitized, rate-limited preview payload suitable for the /{code}+ frontend page.

New Features:
- Add a public link preview API endpoint that exposes status and sanitized destination details for short codes across v1, v2, and emoji aliases.

Enhancements:
- Introduce a dedicated PublicPreviewService and response DTOs to derive preview-safe link metadata without affecting existing redirect logic.
- Configure a specific rate limit bucket for public link preview requests and wire the new service into the FastAPI dependency system.

Tests:
- Add comprehensive integration tests covering status-agnostic resolution, password and expiry-based withholding, geo-targeted destinations, and v1/v2 resolution order for the public preview endpoint.

## Summary by CodeRabbit

* **New Features**
  * Added a public link preview endpoint that works without authentication.
  * Preview responses include link status, creation details, destination information, and geographic routing when applicable.
  * Expired, inactive, and blocked links return preview information while protecting destination details when needed.
  * Added support for both legacy and current short-link formats, including emoji aliases.

* **Security & Reliability**
  * Added dedicated rate limits for public preview requests.
  * Missing short codes return a not-found response.

### DIFF
--- a/services/public_preview_service.py
+++ b/services/public_preview_service.py
@@ -5,14 +5,18 @@ Resolves a short code across both URL generations WITHOUT status gating
 (expired / inactive / blocked links still answer; only truly missing codes
 404) and derives the preview wire shape. The safety semantic the page is
 built on: the destination (and geo rules) ride the wire ONLY while the
-link is active and not password-protected — the preview never reveals
-more than the redirect would.
+link is active and not password-protected — the preview never reveals a
+destination the redirect would refuse to serve. For geo-targeted links
+that means enumerating the full country→destination map while a single
+redirect follows only one rule: deliberate anti-cloaking transparency,
+not a leak.
 
 Deliberately does NOT use ``UrlService.resolve`` — it raises on non-active
 statuses and its cache shape lacks ``created_at``. Raw docs are resolved
-here with the same dispatch heuristic (``UrlService._dispatch``), scoped
-to the system default domain. Custom-tenant aliases never resolve on this
-endpoint (known gap, out of scope).
+here instead, in the order ``shared.alias_dispatch.resolution_order``
+prescribes (the same single source of truth the redirect dispatches on).
+Lookups are scoped to the system default domain: custom-domain aliases
+resolve only via the redirect; this endpoint is system-domain-only.
 
 Everything is derived READ-ONLY — this endpoint never writes.
 """
@@ -21,7 +25,7 @@ from __future__ import annotations
 
 from datetime import datetime, timezone
 from typing import Any
-from urllib.parse import unquote, urlparse
+from urllib.parse import unquote
 
 from errors import NotFoundError
 from infrastructure.logging import get_logger
@@ -33,36 +37,12 @@ from schemas.dto.responses.public_preview import (
     PreviewGeoDestination,
     PublicPreviewResponse,
 )
-from schemas.models.url import UrlStatus, UrlV2Doc
+from schemas.models.url import SchemaVersion, UrlStatus, UrlV2Doc
+from shared.alias_dispatch import resolution_order
 from shared.datetime_utils import convert_to_gmt, parse_datetime
-from shared.validators import is_emoji_alias
+from shared.url_utils import split_destination
 
 log = get_logger(__name__)
-
-
-def _split_destination(url: str) -> dict:
-    """Split a destination URL into display parts.
-
-    Lifted from the legacy Jinja preview's ``_split_destination``
-    (routes/legacy/url_shortener.py) — importing the route module would
-    drag in template/limiter side-effects. Keep the two in lockstep; the
-    spoo-landing mock mirrors this exact logic.
-    """
-    parsed = urlparse(url)
-    domain = parsed.netloc or parsed.path.split("/")[0]
-    path = (
-        parsed.path
-        + ("?" + parsed.query if parsed.query else "")
-        + ("#" + parsed.fragment if parsed.fragment else "")
-    )
-    if path == "/":
-        path = ""
-    return {
-        "url": url,
-        "domain": domain,
-        "path": path,
-        "is_https": parsed.scheme == "https",
-    }
 
 
 class PublicPreviewService:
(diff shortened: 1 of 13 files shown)


services/public_preview_service.py
```diff
@@ -0,0 +1,284 @@
...
+            NotFoundError: the code exists in neither generation.
+        """
+        # Emoji aliases arrive percent-encoded — same as the redirect path.
+        short_code = unquote(short_code)
+
+        # Dispatch mirrors UrlService._dispatch: emoji → emojis collection;
+        # 6 chars → v1 first, v2 fallback; anything else → v2 first, v1
+        # fallback. Lookups are exact matches — case is never normalized,
+        # exactly like the redirect ("/Docs" and "/docs" are different links).
+        if is_emoji_alias(short_code):
```
Zingzy: **[Should-fix] Resolution heuristic is a second copy of `UrlService._dispatch` — drift risk on a public surface.** This len/order dispatch (emoji-only → `len == 6` v1-first → else v2-first) is faithful *today* — I diffed it against `url_service.py:1108`; the redirect's explicit `== 7` branch does the same v2-first as "other", so merging them is correct, and `is_emoji_alias` is already shared. But this is exactly where a silent drift bites: change the redirect's threshold and not this copy, and the preview describes a *different link* than the one that actually redirects for the same code — a public endpoint returning link A's facts while the redirect serves link B.

Extract the length/order skeleton into a shared helper both call, or at minimum add a test that pins the two dispatchers to the same answer over a code table. The "mirrors `UrlService._dispatch`" comment is the current guard — make it structural.

services/public_preview_service.py
```diff
@@ -0,0 +1,284 @@
...
+            destination=destination,
+            geo_destinations=geo_destinations,
+        )
+
+    def _v2_effective_status(self, doc: UrlV2Doc) -> str:
+        """Lowercase wire status, folding derived expiry into ``expired``.
+
+        The persisted status flip happens on the click path
+        (``UrlRepository.expire_if_max_clicks``); a time-based flip has no
+        writer at all. The preview must agree with what the redirect would
```
Zingzy: **[Should-fix] This parity claim is inverted — reword it, because it's a future foot-gun.** The redirect enforces *no* time-based expiry at all: `resolve()` never compares `expire_after`/`expiration-time` to now (v2 gates on persisted status, v1 on live max-clicks only — `url_service.py:491` is `schema == V1`-gated). So the preview deriving time-expiry doesn't "agree with the redirect" — it's **stricter** (safe direction: it withholds a destination the redirect would still serve). The module docstring's "never reveals more than the redirect would" (line 8) is likewise not literally true for geo (`_group_geo_rules` reveals the full country→destination map; a single redirect reveals one).

The behavior is correct — just state the invariant that actually holds: *"the preview never reveals a destination the redirect would refuse"*, and note the derived time-expiry is deliberately stricter than the (expiry-blind) redirect, and geo enumeration is intentional transparency. The risk of the current wording: someone reads "must agree with the redirect," sees the divergence, and "fixes" it by removing the derived expiry — turning the preview into a surface that shows destinations for time-expired links.

services/public_preview_service.py
```diff
@@ -0,0 +1,284 @@
...
+    PublicPreviewResponse,
+)
+from schemas.models.url import UrlStatus, UrlV2Doc
+from shared.datetime_utils import convert_to_gmt, parse_datetime
+from shared.validators import is_emoji_alias
+
+log = get_logger(__name__)
+
+
+def _split_destination(url: str) -> dict:
```
Zingzy: **[Nit] Move to `shared/url_utils.py` and harden.** This is duplicated verbatim from `routes/legacy/url_shortener.py:418` with a "keep in lockstep" comment — but it's pure URL-splitting with no I/O, and `shared/url_utils.py` already exists next door (`extract_hostname`, `extract_fqdn`, `normalise_fqdn`). Move it there and import from both callers — that deletes the lockstep comment and the drift risk in one move (this is the change that most directly answers the "lifted with a pointer comment" smell).

While moving it: wrap the `urlparse`. v2 destinations were validated at creation, but the v1/emoji `url` is a raw legacy value, and `urlparse` raises `ValueError` on some malformed hosts (e.g. an unclosed IPv6 bracket `http://[::1`). On a public endpoint that's an unhandled 500 rather than a graceful 200 — availability, not a leak. Catch and fall back to treating the whole value as the domain.

services/public_preview_service.py
```diff
@@ -0,0 +1,284 @@
...
+404) and derives the preview wire shape. The safety semantic the page is
+built on: the destination (and geo rules) ride the wire ONLY while the
+link is active and not password-protected — the preview never reveals
+more than the redirect would.
+
+Deliberately does NOT use ``UrlService.resolve`` — it raises on non-active
+statuses and its cache shape lacks ``created_at``. Raw docs are resolved
+here with the same dispatch heuristic (``UrlService._dispatch``), scoped
+to the system default domain. Custom-tenant aliases never resolve on this
+endpoint (known gap, out of scope).
```
Zingzy: **[Nit] Drop the scope-narration.** `(known gap, out of scope)` is PR-body/reviewer talk that rots after merge. State the constraint the next reader needs: e.g. `# custom-domain aliases resolve only via the redirect; this endpoint is system-domain-only`. (The rest of the comments in this file are careful and load-bearing — keep them.)

Zingzy (commented):
## Review: public link preview endpoint

**Mergeable, no blockers — and no leak.** I traced the destination/password-withholding invariant and ran an adversarial pass that tried specifically to refute it across every `(status × password × generation × geo)` combination. It holds: the preview emits a `destination`/`geo_destinations` only as a **strict subset** of when the redirect would serve a `Location:`, and no field carries a v1 plaintext password, a v2 hash, or owner `meta_tags`. What reads as "leaky/sloppy" at a glance — the re-implemented resolver, the lifted `_split_destination`, the raw-dict fetch — is deliberate and documented; the real findings are structural, not security. CI green 3.10–3.13; ran the suite in a worktree (22 passed).

**Done well:** the withholding gate as a single choke point on both paths; passwords boolean-coerced so neither generation's secret rides the wire (tested with `argon2`/`hunter2` non-leak assertions); the `aggregate() -> dict | None` contract correctly used (the legacy/emoji repos override the base list-return specifically for this — not the mock-hides-a-500 trap it first looks like); `>=` max-clicks matching the redirect exactly.

**Merge safety — note there's NO flag:** the router includes `public_preview` unconditionally, so `GET /api/v1/public/preview/{code}` is live for anyone the moment a release ships. "Ships dark" here is obscurity (frontend/Caddy don't link it yet), not a gate — so the withholding correctness is a release-time gate, not an enable-time one. It's correct now, so that's fine; just don't file remaining hardening under "we'll fix before enabling" — there's no enable step.

**Product:** custom-domain links can't be previewed (resolver is system-domain-only) while the redirect handles them — fine if `/{code}+` only runs on the system domain; worth confirming that's the frontend's assumption.

Inline: two should-fixes that attack the duplication (share the dispatch; move `_split_destination` to `shared/`) and a comment reword that prevents a future foot-gun.

PR #241, opened 2026-07-12
feat: public per-link stats endpoint

Zingzy:
## Summary

Adds `GET|POST /api/v1/public/stats/{short_code}` — the backend for the already-built public stats page (`/stats/{code}`, spoo-landing PR #4). The wire contract is frozen in the spoo-landing mock (`app/api/mock/[...path]/public.ts`, consumed via `lib/api/public-stats.ts`); this implements it against the real collections per `thoughts/public-stats-endpoint-trd.md`. The endpoint ships dark: nothing serves it until the frontend + Caddy wiring land.

- Resolves BOTH generations plus emoji aliases, scoped to the system default domain (custom-tenant aliases never resolve here). Emoji collapses to `generation: "v1"` on the wire.
- Privacy: a v2 link with `private_stats` answers **byte-identically** to a missing code (404, no oracle). Owner sessions (optional auth on the same request) bypass the privacy and password gates.
- Passwords only ever travel in the POST body — GET ignores `?password=` entirely. 401 `password_required` / `invalid_password` (argon2 for v2, plaintext compare for v1/emoji).
- Effective status is derived READ-ONLY (an ACTIVE link past `expire_after`/max-clicks reports `expired`; no writes from this endpoint), and `long_url` is withheld unless the link is effectively active (owners always get it).
- v2 analytics reuse `StatsService`'s $facet machinery with the $match scoped by `meta.url_id` (never `meta.short_code`, so a same-alias link on a custom domain can't bleed in). v1/emoji synthesize the same wire shape from the embedded counters: lifetime dimensions, windowed daily time series, `bots` dimension, countries converted to ISO alpha-2. v1/emoji docs are read RAW via the repos' `aggregate` helper (the typed model drops `creation-date`/`creation-time`) so `link.created_at` is populated for v1 — same pattern as the public preview endpoint, so `/stats/{code}` and `/{code}+` agree.

## Changes

- `routes/api_v1/public_stats.py` — GET/POST route pair sharing one impl; `dynamic_limit(API_AUTHED, API_ANON)` rate limiting; optional auth via `optional_scopes(STATS_SCOPES)`
- `services/public_stats_service.py` — resolution, privacy/password gates, derived status, v2 reuse + v1 synthesis
- `schemas/dto/requests/public_stats.py`, `schemas/dto/responses/public_stats.py` — `PublicStatsBody`, `PublicLinkFacts`, `PublicStatsResponse`
- `errors.py` — `PasswordRequiredError` / `InvalidPasswordError` (401s with stable codes)
- `routes/api_v1/__init__.py`, `dependencies/services.py`, `dependencies/wiring.py` — router include + service wiring
- `services/public_link_resolver.py` — shared resolver (resolution order, raw v1 reads, derived status, created_at) consumed by BOTH public endpoints (preview + stats) after #240 landed first; `services/stats_service.py` gained the public `compute()` seam + `normalize_timezone`/`to_user_tz`/`add_metadata` so this endpoint touches no privates
- `middleware/rate_limiter.py` — dedicated `PUBLIC_STATS_AUTHED`/`PUBLIC_STATS_ANON` budgets (aggregation cost + password-guess reasoning in the comment)

## Tests

- `tests/integration/api_v1/test_public_stats.py` — 22 tests covering the full contract matrix: byte-identical 404s, private-stats semantics (`True`/`None`/`False`), owner bypass (incl. `long_url` on non-active links), both password gates (incl. GET query-string password ignored and empty POST body), v1 wire synthesis (lifetime dims, windowed time series, ISO countries, `os` key, bots-not-city), v2 wire (city-not-bots, `meta.url_id` $match asserted on the captured pipeline), derived expiry for both generations, emoji resolution, lowercase status + system-domain `short_url`, resolution fallback, date validation, v1 `created_at` from raw creation-date/creation-time (populated, missing, and unparseable).
- Targeted: `uv run pytest tests/integration/api_v1/test_public_stats.py -p no:xdist -o addopts=""` — 22 passed.
- Full sweep: `uv run pytest` — 2147 passed.

## Summary by Sourcery

Add a public per-link statistics API endpoint for short codes, wired through a dedicated service and router, with optional owner authentication and strict privacy/password handling.

New Features:
- Introduce GET/POST /api/v1/public/stats/{short_code} endpoint to serve public stats for short links across V2, legacy, and emoji generations.
- Define request and response DTOs for the public stats endpoint, including link facts and stats payload compatible with the existing stats wire format.

Enhancements:
- Implement PublicStatsService to resolve short codes on the system default domain, enforce privacy and password rules, derive effective status, and either reuse V2 stats aggregation or synthesize V1 analytics.
- Wire PublicStatsService into the FastAPI app dependencies alongside StatsService, enabling DI-based usage from routes.
- Extend error handling with dedicated password_required and invalid_password authentication errors carrying stable error codes.

Tests:
- Add comprehensive integration tests for the public stats endpoint covering privacy semantics, owner bypass, password flows for V1/V2, emoji resolution, derived expiry, stats wiring, and date-window validation.

### DIFF
--- a/routes/api_v1/public_stats.py
+++ b/routes/api_v1/public_stats.py
@@ -25,8 +25,11 @@ from services.public_stats_service import PublicStatsService
 
 router = APIRouter(tags=["Public"])
 
+# Dedicated pair — NOT the generic API tier. Reasoning lives on
+# Limits.PUBLIC_STATS_*: anon hits can run 90-day $facet aggregations,
+# and this bucket doubles as the password-guess budget.
 _public_stats_limit, _public_stats_key = dynamic_limit(
-    Limits.API_AUTHED, Limits.API_ANON
+    Limits.PUBLIC_STATS_AUTHED, Limits.PUBLIC_STATS_ANON
 )
 
 _PATH = "/public/stats/{short_code}"
@@ -104,8 +107,8 @@ async def public_stats_get(
 
     **Rate Limits**:
 
-    - Authenticated: 60/min, 5,000/day
-    - Anonymous: 20/min, 1,000/day
+    - Authenticated: 60/min, 2,000/day
+    - Anonymous: 20/min, 500/day
     """
     return await _serve_public_stats(
         public_stats_service,
--- a/routes/api_v1/public_stats.py
+++ b/routes/api_v1/public_stats.py
@@ -1,0 +1,154 @@
+"""
+GET|POST /api/v1/public/stats/{short_code} — public per-link statistics.
+
+Serves the public stats page (/stats/{code}) for BOTH URL generations on
+the system default domain. No auth required; an authenticated owner
+session bypasses the privacy and password gates.
+
+GET never accepts a password — not as a query param, not as a header
+(an undeclared ``?password=`` is simply ignored). POST exists solely to
+carry ``{"password"}`` in a JSON body; it accepts the same query params.
+"""
+
+from __future__ import annotations
+
+from bson import ObjectId
+from fastapi import APIRouter, Depends, Query, Request
+
+from dependencies import STATS_SCOPES, CurrentUser, optional_scopes
+from dependencies.services import PublicStatsSvc
+from middleware.openapi import ERROR_RESPONSES, OPTIONAL_AUTH_SECURITY
+from middleware.rate_limiter import Limits, dynamic_limit, limiter
+from schemas.dto.requests.public_stats import PublicStatsBody
+from schemas.dto.responses.public_stats import PublicStatsResponse
+from services.public_stats_service import PublicStatsService
+
+router = APIRouter(tags=["Public"])
+
+_public_stats_limit, _public_stats_key = dynamic_limit(
+    Limits.API_AUTHED, Limits.API_ANON
+)
+
+_PATH = "/public/stats/{short_code}"
+
+_START_DATE_QUERY = Query(
+    default=None,
+    max_length=50,
+    description="Range start (ISO 8601). Defaults to 7 days before end_date.",
+    examples=["2025-01-01T00:00:00Z"],
+)
+_END_DATE_QUERY = Query(
+    default=None,
+    max_length=50,
+    description="Range end (ISO 8601). Defaults to now.",
+    examples=["2025-12-31T23:59:59Z"],
+)
+_TIMEZONE_QUERY = Query(
+    default="UTC",
+    max_length=50,
+    description="IANA timezone for time bucketing and formatting.",
+    examples=["UTC", "America/New_York"],
+)
+
+
+async def _serve_public_stats(
+    public_stats_service: PublicStatsService,
+    short_code: str,
+    *,
+    start_date: str | None,
+    end_date: str | None,
+    timezone: str,
+    password: str | None,
+    user: CurrentUser | None,
+) -> PublicStatsResponse:
+    user_id: ObjectId | None = user.user_id if user is not None else None
+    result = await public_stats_service.get_public_stats(
+        short_code,
+        start_date=start_date,
+        end_date=end_date,
+        tz_name=timezone,
+        password=password,
+        user_id=user_id,
+    )
+    return PublicStatsResponse.model_validate(result)
+
+
+@router.get(
+    _PATH,
+    responses=ERROR_RESPONSES,
+    openapi_extra=OPTIONAL_AUTH_SECURITY,
+    operation_id="getPublicStats",
+    summary="Public URL Statistics",
+)
+@limiter.limit(_public_stats_limit, key_func=_public_stats_key)
+async def public_stats_get(
+    request: Request,
+    short_code: str,
+    public_stats_service: PublicStatsSvc,
+    start_date: str | None = _START_DATE_QUERY,
+    end_date: str | None = _END_DATE_QUERY,
+    timezone: str = _TIMEZONE_QUERY,
+    user: CurrentUser | None = Depends(optional_scopes(STATS_SCOPES)),  # noqa: B008
+) -> PublicStatsResponse:
+    """Get public click statistics for a single short link.
+
+    Resolves both URL generations (plus emoji aliases) on the system
+    default domain and returns link facts plus the standard stats wire.
+
+    **Authentication**: Optional — an owner session additionally sees
+    private-stats links and skips the password gate.
+
+    **Privacy**: A link with private stats answers exactly like a missing
+    code (404, byte-identical). Password-protected links answer 401
+    `password_required`; send the password via POST — never in the URL.
+
+    **Rate Limits**:
+
+    - Authenticated: 60/min, 5,000/day
+    - Anonymous: 20/min, 1,000/day
+    """
+    return await _serve_public_stats(
+        public_stats_service,
+        short_code,
+        start_date=start_date,
+        end_date=end_date,
+        timezone=timezone,
+        password=None,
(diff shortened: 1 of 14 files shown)


services/public_stats_service.py
```diff
@@ -0,0 +1,555 @@
...
+        """Run the standard $facet aggregation scoped by ``meta.url_id``.
+
+        Matching on url_id — never meta.short_code — so a same-alias link
+        on a custom domain can never bleed into the public page.
+        """
+        click_query: dict[str, Any] = {
+            "meta.url_id": doc.id,
+            "clicked_at": {"$gte": start_date, "$lte": end_date},
+        }
+        summary, aggregation_results = await self._stats._execute_all_stats(
```
Zingzy: **[Should-fix] Reaching into four `StatsService` privates — cross-service coupling to internals.** This orchestrates `_execute_all_stats` → `_format_results` → `_add_metadata` (and the v1 path uses `_to_user_tz` + imports the private `_TIMEZONE_ALIASES`). `StatsService` already exposes a public `query()` (line 409), but this bypasses it to re-assemble the private pipeline — reasonably, since `query()` bakes in scope/owner/filter handling you don't want here. The cost: four leading-underscore methods are now a load-bearing contract with no test pinning them, so a refactor of `StatsService`'s internals breaks this public endpoint **silently**.

Add one public seam on `StatsService` — e.g. `compute(click_query, group_by, metrics, window, tz) -> dict` — that both `query()` and this service call. Turns the coupling into a supported interface. Top structural finding; blast radius is a silent break of a public endpoint on an unrelated stats refactor.

services/public_stats_service.py
```diff
@@ -0,0 +1,555 @@
...
+            end_date=window_end.isoformat(),
+            total_clicks=stats.get("summary", {}).get("total_clicks", 0),
+            duration_ms=duration_ms,
+        )
+
+        return {"generation": generation, "link": link, "stats": stats}
+
+    # ── Private: resolution ───────────────────────────────────────────────────
+
+    async def _resolve(
```
Zingzy: **[Should-fix] This whole resolution+derivation surface is a second copy of #240's `public_preview_service`.** `_resolve` (the dispatch heuristic), `_find_v1_raw`, the derived-expiry (`_effective_status`/`_expiry_reached` ≈ #240's `_v2_effective_status`), `_v1_created_at`, and `_as_utc` (== #240's `_as_aware_utc`) are all re-implemented here — the `_find_v1_raw` docstring literally says "same pattern as the public preview endpoint," which is the duplication admitting itself. Both PRs branch off the same `main`, so once both merge there are **two** full copies plus the redirect's canonical `_dispatch` plus the legacy preview. A drift in any one (a length threshold, a status-derivation tweak) silently desyncs the two public endpoints.

Extract a shared `PublicLinkResolver` (or a shared module of functions) that both public endpoints use for resolution + status-derivation + raw-fetch + created_at. Whichever of #240/#241 lands second should do the extraction. This is the single change that most shrinks the surface area accumulating across the public-endpoint PRs.

routes/api_v1/public_stats.py
```diff
@@ -0,0 +1,154 @@
...
+from dependencies.services import PublicStatsSvc
+from middleware.openapi import ERROR_RESPONSES, OPTIONAL_AUTH_SECURITY
+from middleware.rate_limiter import Limits, dynamic_limit, limiter
+from schemas.dto.requests.public_stats import PublicStatsBody
+from schemas.dto.responses.public_stats import PublicStatsResponse
+from services.public_stats_service import PublicStatsService
+
+router = APIRouter(tags=["Public"])
+
+_public_stats_limit, _public_stats_key = dynamic_limit(
```
Zingzy, line 31: **[Should-decide] Heavy public aggregation + password-guess surface on the generic anon tier.** This uses `dynamic_limit(API_AUTHED, API_ANON)` (20/min·1000/day anon), where #240's *cheaper* point-lookup preview got a dedicated `PUBLIC_PREVIEW` bucket with explicit "existence oracle" reasoning. This is the heavier of the two — each anon call can run a 90-day `$facet` against the hottest public link — and the same bucket doubles as the password-guess budget (~1000/day/IP; argon2 makes v2 guesses costly, but the v1 plaintext compare is cheap server-side).

Good: slowapi bills failed 401/404s (increments before the handler, no refund), so guessing is rate-limited, and the 90-day cap + degrade-to-empty bound the aggregation. **Decide:** give it a dedicated, tighter bucket sized for the aggregation cost and reason about it explicitly like the preview endpoint did, rather than inheriting the generic tier by default.

services/public_stats_service.py
```diff
@@ -0,0 +1,555 @@
...
+        generation = "v2" if is_v2 else "v1"
+
+        # Owner bypass rides the same request. The anonymous sentinel
+        # owner_id can never equal a real user id, so no special-casing.
+        is_owner = bool(is_v2 and user_id is not None and doc.owner_id == user_id)
+
+        # Private stats answer exactly like a missing code. Semantics:
+        # True = private (owned default), None = anonymous/unowned = public,
+        # False = explicitly public. v1/emoji have no flag — always public.
+        if is_v2 and doc.private_stats and not is_owner:
```
Zingzy, line 147: **[Nit / product-confirm] v1/emoji stats are always public.** This gate is `is_v2`-only, so legacy docs (no `private_stats` field) can't be private — any anonymous caller reads the full analytics (counts, countries, browsers, OS, referrers, bot hits) of every non-password legacy link. This mirrors the shipped legacy `/stats/{code}` page exactly (not a regression), but this endpoint turns it into a clean machine-readable JSON API, and a v1 link's owner has no way to make its stats private. Confirm that's intended; there's no test asserting the exposure is deliberate.

Zingzy (commented):
## Review: public per-link stats endpoint

**Mergeable, no blockers, no leak.** I traced the privacy/password/owner-bypass/cross-tenant surface and ran an adversarial pass that tried to refute each invariant — none broke. The security-critical decisions are all correct: private v2 → the *same* `NotFoundError` as a missing code (byte-identical, and a test asserts `missing.content == private.content`), with the privacy check ordered *before* the password check; owner bypass is `is_v2 and doc.owner_id == user_id` (no other authed user, no anonymous, no v1 can bypass); `meta.url_id` scoping (not short_code) blocks cross-tenant bleed, pinned by a test; v1 synthesis emits only counts, never raw IPs. CI green 3.10–3.13; ran the suite in a worktree (19 passed).

The findings are structural, not security — inline. Lower-priority (not inlined): the emoji-miss / 7-char-both-miss branches rely on `from_mongo(None)`→None (correct, no 500, but unguarded and untested, unlike #240's explicit null-check); the v1 plaintext password compare is non-constant-time (legacy reality; v2 uses argon2); malformed legacy embedded data can 500 via the generic handler (no stack leak).

**Merge safety:** no feature flag — the router includes `public_stats` unconditionally, so it's live for anyone the moment a release ships. "Ships dark" is obscurity, not a gate; the privacy/password correctness is the gate and it holds now, so it's safe to release — just don't file the rate-limit sizing under "tighten before enabling," there's no enable step.

**Done well:** byte-identical privacy 404 + privacy-before-password ordering; owner check fails closed for anonymous-owned and v1; `meta.url_id` scoping with a test pinning `short_code` out; password body-only with a test proving GET query passwords are ignored; count-only v1 synthesis; 90-day cap + degrade-to-empty aggregation. Disciplined work — the findings are about where code lives and what it couples to, not correctness.

PR #242, opened 2026-07-12
feat: report intake + contact API

Zingzy:
## Summary

Backend half of `thoughts/contact-report-pages.md`, contract per `thoughts/report-contact-intake-trd.md`. Bulk-first report intake: `POST /api/v1/reports` accepts 1..N bare codes or full short URLs (custom domains included), returns a per-item accepted/rejected breakdown, and stores **dedupe+velocity** records — one document per `(domain, code)`, re-reports `$inc` a counter and refresh `last_reported_at` instead of inserting rows. The report Discord webhook is **demoted to one summary embed per submission** (storage is the system of record; a failed send is logged, never surfaced). `POST /api/v1/contact` is the thin JSON twin of the Jinja form, reusing `ContactService.send_contact_message` verbatim.

The wire contract is **frozen** — the Next report page + contact wiring build against it in parallel. Ships dark: nothing links to these endpoints until the report page lands; the Jinja `/contact` and `/report` handlers are untouched (Phase 8 deletes them).

## Changes

- `routes/api_v1/reports.py` — both endpoints under the `Reports & Contact` tag. 503 `not_configured` when the corresponding webhook is unset (mirrors the Jinja 503s); missing captcha token → 400 before any service work, mirroring the forms.
- `services/report_intake_service.py` — normalization (`abc123` / `spoo.me/abc123?x=1` / `go.customer.com/deal` → `(domain, code)`, percent-decoded for emoji, host lowercased, code case preserved), within-batch dedupe (first wins), domain-scoped existence checks (system domain via the shared `PublicLinkResolver` across v1/v2/emoji; custom domains via `UrlRepository.find_by_alias` so the record keeps the tenant fqdn), storage, submission audit record, summary embed.
- `repositories/report_repository.py` — `reports` upsert (`$inc` count/reporters/source_counts, `$addToSet` reasons/vectors/reporter_ids, `$setOnInsert` status=open — the funnel owns transitions) + `report_submissions` insert.
- `repositories/indexes.py` — unique `(domain, code)`, secondaries on `last_reported_at` desc + `status`; `created_at` desc on submissions.
- Caps: anonymous ≤ 25 items/request (captcha-gated), authenticated ≤ 100, over-cap → 400 for the whole request. Rate limits per **submission**, not per item: `REPORTS_AUTHED` 30/min·500/day, `REPORTS_ANON` 5/min·40/day via the `dynamic_limit` pair. Contact rides the `Limits.CONTACT` values.
- **New `reports:create` API-key scope** (`REPORTS_SCOPES = {reports:create, admin:all}` via `optional_scopes`): a leaked read-only key must not be able to file reports attributed to its owner — reporter identity is a reputation handle. `ALLOWED_SCOPES` derives from the enum so key creation accepts it immediately; the dashboard scope picker needs the value added (frontend PR).
- `errors.NotConfiguredError` (503, `not_configured`), OpenAPI tag + 503 response set, wiring + dependency getters.

Contract notes (deviations from the TRD, flagged for the frontend PR):
- **DTO shape errors return 422** (invalid email, bad reason/vector enum, missing fields) with the same `{error, code: "validation_error", field?}` body — the codebase-wide convention every `/api/v1` endpoint follows (the TRD's own `EmailStr` prescription produces this). Semantic gates keep the TRD's 400s: empty items, over-cap, missing captcha token.
- Contact shares the `Limits.CONTACT` **values** with the Jinja form but not the literal slowapi counter — sharing one bucket would require touching the Jinja handler's decorator, which this PR must not do. Collapses to one budget at Phase 8.

## Tests

- `tests/integration/api_v1/test_reports.py` (21) — real service + real repositories over capturing fake collections, real resolver over dict-backed repo fakes. Pins the frozen wire bodies exactly, the `$inc`/`$addToSet` upsert shape at the Mongo boundary, per-item mixed-batch breakdown, both caps, captcha matrix (missing/fail/authed-skip), scope enforcement + `web|api` source attribution, emoji percent-decoding, domain-scoped custom-domain resolution, re-report increments, webhook-unset 503, webhook-failure resilience, and a 6th-request 429 smoke.
- `tests/integration/api_v1/test_contact.py` (9) — exact legacy embed pin, captcha semantics, 503, send-failure 500, validation.
- Full sweep: **2192 passed**. `ruff check` + `ruff format` clean.

## Summary by CodeRabbit

* **New Features**
  * Added API endpoints for submitting URL abuse reports in bulk and sending contact messages.
  * Reports provide accepted counts and item-level rejection details.
  * Added support for report reasons, contact details, optional CAPTCHA verification, and authenticated or anonymous submissions.
  * Added a `reports:create` API key permission.
* **Improvements**
  * Added rate limiting and clearer configuration errors for unavailable integrations.
  * Report submissions are recorded and notifications continue to work even if delivery fails.


services/report_intake_service.py
```diff
@@ -0,0 +1,383 @@
...
+    return (domain, code)
+
+
+@dataclass
+class RejectedItem:
+    """One rejected entry — ``input`` echoes the raw ``code_or_url``."""
+
+    index: int
+    input: str
+    code: Literal["invalid_input", "not_found", "duplicate_in_batch"]
```
Zingzy: **[Should-fix] Rejection-code `Literal` is a second source of truth.** This `Literal["invalid_input", "not_found", "duplicate_in_batch"]` is duplicated verbatim on the wire DTO (`schemas/dto/responses/reports.py:33`), and the route maps one to the other (`RejectedReportItem(code=r.code)`). Add a fourth rejection reason here and forget the DTO, and pydantic rejects the out-of-`Literal` value at **response serialization** — a 500 on an otherwise-valid submission, not a clean error.

Low likelihood (needs a future edit to trip), but the failure mode is a 500, so it's the one structural thing worth tightening. Extract the literal to one shared type alias both import (e.g. `RejectionCode` next to the enums). Honestly the strongest finding in an otherwise clean PR.

services/report_intake_service.py
```diff
@@ -0,0 +1,383 @@
...
+    async def _exists(self, domain: str | None, code: str) -> bool:
+        """Domain-scoped existence check.
+
+        System-domain codes resolve via the shared PublicLinkResolver —
+        the same generation dispatch (v1/v2/emoji) as the redirect, and
+        status-agnostic on purpose: expired/blocked links are still
+        reportable. Custom-domain codes are exact v2 lookups.
+        """
+        if domain is None:
+            return await self._resolver.resolve(code) is not None
```
Zingzy, line 308: **[Not a bug — flagging so the Sourcery comment isn't actioned.]** Sourcery flagged this as awaiting the boolean (`await (resolve(code) is not None)`). It isn't: `await` binds tighter than `is not`, so this parses `(await self._resolver.resolve(code)) is not None` — confirmed via AST, and the emoji/custom-domain tests that exercise `_exists` pass (awaiting a `bool` would raise at runtime). The code is correct as written; the suggested parenthesization is harmless but the premise is wrong.

services/report_intake_service.py
```diff
@@ -0,0 +1,383 @@
...
+
+        fields.append({"name": "IP Address", "value": f"```{ip}```"})
+
+        return {
+            "embeds": [
+                {
+                    "title": "New URL Report Submission",
+                    "color": 14177041,
+                    "fields": fields,
+                    "timestamp": datetime.now(timezone.utc).isoformat(),
```
Zingzy: **[Nit] Reuse the `now` from `submit` instead of a fresh `datetime.now()`.** `submit` already stamps `now = datetime.now(timezone.utc)` for the audit record; this recomputes it, so the audit `created_at` and the embed `timestamp` drift by microseconds and the embed is harder to pin deterministically. Thread `now` in as a param. (This is the one valid Sourcery comment.)

services/report_intake_service.py
```diff
@@ -0,0 +1,383 @@
...
+            },
+        ]
+
+        if accepted:
+            lines = []
+            for domain, code, item in accepted[:_SUMMARY_MAX_LISTED]:
+                display = f"{domain or self._system_default_domain}/{code}"
+                line = f"{display} — {item.reason.value}"
+                if len(line) > _SUMMARY_LINE_MAX:
+                    line = line[: _SUMMARY_LINE_MAX - 1] + "…"
```
Zingzy, line 356: **[Nit] The two summary-embed branches here are the only uncovered lines** (351 line-clip, 355 `… and N more`). Cosmetic, but a single test with 11 accepted items and one long code closes both and pins the overflow wording.

Zingzy (commented):
## Review: report intake + contact endpoints

**Mergeable, no blockers, no leak.** I traced the four things that could actually bite this surface — a destination/owner-info leak, cross-tenant bleed, the reputation-attack vector (false reports auto-actioning a link), and Discord embed injection — and ran the adversarial pass before forming the verdict. All hold. Unusually well-tested; the findings are one mild structural redundancy and a few nits, plus a bot false-positive to *not* act on (inline). Ran both new suites in a worktree (30 passed), 98% coverage on the new service; CI green 3.10–3.14.

**Architecture read (load-bearing calls, all correct):**
- **Reports are a funnel signal, not a trigger** — `$inc`/`$addToSet` velocity with `$setOnInsert status="open"`, and the resolution funnel owns every transition, never this layer. A false report can't auto-action a link; it adds a velocity tick a human triages, and `reporter_ids`/IP give triage the coordinated-flood signal. This is the most important product decision here and it's right (matches the harm-tiered, reputation-first safety posture).
- **Existence is status-agnostic on purpose** — expired/blocked links stay reportable.
- **Reuses the shared `PublicLinkResolver`** for system-domain existence (same generation dispatch the redirect serves) and `find_by_alias(code, domain)` for custom domains, keeping the tenant fqdn on the record. Resolution isn't re-implemented; the choke point is shared.
- **Framework-agnostic service**; `reporter_id`/`source` derived at the route boundary where the auth context lives — correct layering.
- **Model-less repos (`BaseRepository[None]`) + direct `update_one(upsert=True)`** are both the established house idiom (`ClickRepository`, `FeatureFlagRepository.upsert`), not a smell.
- **Wiring is complete** across every registration point (wire_services, services getter + alias, `__init__` import + `__all__`, router, indexes, all three smoke fakes, the index unit test, the `ApiKeyScope` enum, `REPORTS_SCOPES`, keys.py doc).

**Merge safety:** no feature flag — `/reports` (optional-auth, captcha-gated for anon) and `/contact` (public) go live the moment a release ships, same as #240/#241. Fine: the anon surface is bounded by captcha + the 25-item cap + 5/min·40/day, and correctness is the gate, which holds now. Inherited caveat (not this PR): the audit IP and anon rate-limit key both come from `get_client_ip`, which trusts client headers — trustworthy only insofar as the CF Tunnel keeps the origin unreachable directly.

**Edge note (not a finding):** a `PyMongoError` mid-batch stores the earlier items but not the audit record and 500s; a client retry re-`$inc`s the already-stored codes (velocity over-count). Rare + triage-only signal, so not worth engineering around — just know per-code count isn't retry-exact.

**Notably good:** the test design — real service + real repos over *capturing* fakes that pin the `$inc`/`$addToSet` shape at the Mongo boundary, a fake `reports` collection with **no insert method** so an insert-per-report regression fails loudly, frozen wire bodies, and the full captcha/scope/cap/emoji/custom-domain/re-report/resilience/429 matrix. The embed injection surface is identical to the shipped contact/report embeds (raw content in code blocks; Discord doesn't ping mentions inside embeds), so no new class of risk.

PR #246, opened 2026-07-13
fix: enforce time-based URL expiry at resolve + derive effective status

### LINKED ISSUE #185: [BUG] Links do not expire after the set expire_after time.
## Describe the bug
Links do not expire after the set `expire_after` time. The link continues to redirect normally even after the expiration time has passed. This was tested both via the website UI and the API.

## To Reproduce
1. Create a shortened URL with `expire_after` set to a time in the near future (e.g. 1-2 minutes ahead)
2. Wait until the expiration time passes
3. Visit the shortened link
4. Link still works and redirects normally — no expiration occurs

Via API:
```json
POST https://spoo.me/api/url
Content-Type: application/json
Authorization: Bearer {API_KEY}

{
  "long_url": "https://www.google.com",
  "expire_after": "2026-05-11T18:10:00Z"
}
```

## Expected behavior
After the `expire_after` datetime passes, visiting the shortened link should result in an expiration/error page, not a redirect to the original URL.

---

## Screenshots

Image

It's 20:18 in Poland so UTC+2, so 18:18 in UTC
---

## Desktop (please complete the following information):
 - OS: Windows
 - Browser: Brave (also tested via API / Postman)

---

## Additional context

Below is a timeline proving the link was accessed **after** the expiration time:

**API response after creating the link:**
```json
{
    "alias": "fDoMIzA",
    "short_url": "https://spoo.me/fDoMIzA",
    "long_url": "https://www.google.com",
    "owner_id": "6a020ec35746cac54b1da051",
    "created_at": 1778522988,
    "status": "ACTIVE",
    "private_stats": true
}
```

Note: the API response does not include `expire_after` in the returned object, which may indicate the field is being silently ignored.

**Stats after clicking the link:**
```json
{
    "summary": {
        "first_click": "2026-05-11T18:10:49.343000Z",
        "last_click": "2026-05-11T18:10:49.343000Z"
    },
    "short_code": "fDoMIzA"
}
```

The link was set to expire at `2026-05-11T18:10:00Z` and was successfully accessed at `2026-05-11T18:10:49Z` — **49 seconds after the expiration time** — and redirected normally.

---

## Test Environment:
N/A — tested against the hosted production instance at spoo.me

---

## Logs & Error Messages
No error messages were returned by the API. The link creation request returned HTTP 200 with `"status": "ACTIVE"`.

## Code Snippets
```json
{
  "long_url": "https://www.google.com",
  "expire_after": "2026-05-11T18:10:00Z"
}
```

Zingzy:
## The bug

\`expire_after\` is accepted, stored, validated, and documented — but was never enforced at redirect time. A URL whose expiry passed kept redirecting forever, and its stored \`status\` stayed \`ACTIVE\` everywhere (dashboard pill, list filters, update responses). The only thing that ever flipped status to \`EXPIRED\` was max-clicks.

This is a regression with a paper trail: expiry **was** enforced at redirect from Feb 2024 (\`51628a0\`) until the Dual-Cache rewrite (\`0c64b08\`, Jun 2025) dropped the check as collateral damage — the cache object stopped carrying the field, and the check never came back. The FastAPI migration faithfully ported the broken behavior, then the v2 API re-documented \`expire_after\` as a working feature.

Design doc: \`thoughts/time-based-expiry-trd.md\` (derive-on-read + opportunistic flip; no scheduler, no new index, no edge changes).

## What this does

**One derivation, four consumers.** \`UrlV2Doc.effective_status\` folds time lapse and max-click exhaustion into \`EXPIRED\` (only for stored-\`ACTIVE\` docs — \`INACTIVE\`/\`BLOCKED\` stay authoritative). The redirect, the DTOs, the list filter, and the public resolver all answer from this one predicate; a matrix test pins the model property and the Mongo clause against each other so they can't silently diverge.

- **Resolve enforcement** — \`resolve()\` now 410s past-expiry links on both the cache-hit and DB paths (\`expiration_time\` was already in every cache entry, so cache staleness is a non-issue). v2 additionally persists the flip via \`expire_if_time_reached\` — atomic conditional update mirroring \`expire_if_max_clicks\` (\`status: ACTIVE\` guard, \`modified_count\`, invalidate-only-when-flipped). v1/emoji have no status field and lazy-raise forever. Post-flip reads stay O(1): the link recaches as minimal \`EXPIRED\`.
- **Truthful reads** — \`UrlResponse\` / \`UpdateUrlResponse\` / \`UrlListItem\` serialize effective status; \`filter.status=EXPIRED\` now matches time-expired-but-stored-ACTIVE links (and \`ACTIVE\` excludes them) via a clause built from shared arms, composed with the search filter through \`$and\` so the two \`$or\`s can't clobber each other.
- **v1 tz-ambiguity alignment** — a naive \`expiration-time\` never enforces at the redirect, matching \`v1_is_expired\` (public pages) and \`convert_to_gmt\` (legacy stats), so all v1 surfaces agree on the ambiguous rows.
- **\`_auto_reactivate\` tightened** (adjacent bug) — reactivation now requires the improving field in the requested change; previously a max-clicks-expired link with a future \`expire_after\` was resurrected by any unrelated edit, including a no-op update.
- \`public_link_resolver.v2_effective_status\` shrinks to the lowercase wire adapter (casing pinned by test); management docstring updated to describe the reactivation path accurately.

## Deliberately out of scope

- **Edge**: expiring links stay barred from KV promotion (\`has_expiration\` skip). The add-expiry-to-already-promoted-link window is accepted as TTL-bounded (≤ ~6 min) — same invalidation-v1 staleness class as editing \`long_url\` on a promoted link. Follow-up PR (TRD §11) adds \`expires_at\` to the KV contract + a worker-side check with passthrough-on-expired, after this deploys.
- **Sweeper**: nothing needs to happen *at* expiry time until a \`link.expired\` webhook exists; derive-on-read makes enforcement exact without one.
- **v2 max-clicks at resolve**: cached v2 \`total_clicks\` is stale by design; the click path owns that flip. Unchanged.

## Verification

- 2245 tests pass; 46 new (enforcement on every path/ordering incl. BLOCKED-wins, flip idempotency, tz rules, DTO derivation, clause composition, predicate-pin matrix, resurrection regressions).
- Live smoke (uvicorn :8010, real Mongo/Redis, throwaway DB): stored-ACTIVE link with past expiry → 410 on first visit, Mongo status flipped to EXPIRED (\`url_expired reason=expiration_time_reached\`), second visit 410 from recache, future-expiry link 302s, public preview answers \`"status":"expired"\` with \`destination: null\`.

Post-release watch: \`url_resolve_expired_time\` volume in Axiom (steps up, then settles) and the 410 rate on the redirect route.

## Summary by CodeRabbit

* **Bug Fixes**
  * URL `status` is now consistently derived from time-based (`expire_after`) and click-based (`max_clicks`/`total_clicks`) expiry across responses, listings, redirects, and cached links.
  * Expired URLs are reliably prevented from resolving, including cache-hit enforcement with automatic DB status flipping and cache invalidation.
  * Listing filters for status and search now compose correctly without overriding each other.
  * Improved expiration handling for both timezone-aware and timezone-naive datetimes.

* **Documentation**
  * Updated guidance for `PATCH /api/v1/urls/{url_id}/status` reactivation rules for `EXPIRED` URLs.

### DIFF
--- a/infrastructure/cache/url_cache.py
+++ b/infrastructure/cache/url_cache.py
@@ -14,6 +14,7 @@ from pydantic import BaseModel, ConfigDict, Field
 
 from infrastructure.crypto import verify_password as verify_password_hash
 from infrastructure.logging import get_logger
+from shared.datetime_utils import to_unix_timestamp
 
 if TYPE_CHECKING:
     from schemas.models.url import UrlV2Doc
@@ -31,7 +32,8 @@ class UrlCacheData(BaseModel):
     long_url: str
     block_bots: bool
     password_hash: str | None
-    expiration_time: int | None  # Unix timestamp
+    # Unix timestamp; None = no expiry OR tz-ambiguous v1 value (never expires)
+    expiration_time: int | None
     max_clicks: int | None
     url_status: str  # ACTIVE, INACTIVE, BLOCKED, EXPIRED
     schema_version: str  # "v1" or "v2"
@@ -67,9 +69,9 @@ class UrlCacheData(BaseModel):
             long_url=doc.long_url,
             block_bots=bool(doc.block_bots),
             password_hash=doc.password,
-            expiration_time=(
-                int(doc.expire_after.timestamp()) if doc.expire_after else None
-            ),
+            # to_unix_timestamp treats naive as UTC — v2 datetimes come back
+            # from Mongo naive-UTC, so this stays host-TZ-independent.
+            expiration_time=to_unix_timestamp(doc.expire_after),
             max_clicks=doc.max_clicks,
             url_status=doc.status,
             schema_version="v2",
@@ -83,6 +85,13 @@ class UrlCacheData(BaseModel):
             meta_image_width=im.width if im else None,
             meta_image_height=im.height if im else None,
         )
+
+    def is_time_expired(self, now_ts: float) -> bool:
+        """Time-lapse check for the hot path. ``expiration_time`` is None
+        for links without expiry AND for v1 rows whose stored value is
+        tz-ambiguous (see ``_legacy_doc_to_cache``) — both never expire here.
+        """
+        return self.expiration_time is not None and self.expiration_time <= now_ts
 
     def verify_password(self, password: str | None) -> bool:
         """Check a password against this URL's stored hash.
--- a/repositories/url_repository.py
+++ b/repositories/url_repository.py
@@ -154,6 +154,24 @@ class UrlRepository(BaseRepository[UrlV2Doc]):
             {"$set": {"status": UrlStatus.EXPIRED}},
         )
 
+    async def expire_if_time_reached(self, url_id: ObjectId) -> bool:
+        """Conditionally expire the URL if its expire_after has passed.
+
+        Atomic conditional update mirroring ``expire_if_max_clicks`` —
+        matches only ACTIVE docs so BLOCKED/INACTIVE are never clobbered.
+        Returns True only if this call performed the flip (modified_count).
+        ``$lte`` on a date never matches null/missing (BSON type
+        bracketing), so no explicit null guard is needed.
+        """
+        return await self._update_modified(
+            {
+                "_id": url_id,
+                "status": UrlStatus.ACTIVE,
+                "expire_after": {"$lte": datetime.now(timezone.utc)},
+            },
+            {"$set": {"status": UrlStatus.EXPIRED}},
+        )
+
     async def find_by_owner(
         self,
         query: dict,
--- a/routes/api_v1/management.py
+++ b/routes/api_v1/management.py
@@ -153,7 +153,10 @@ async def update_url_status_v1(
 
     **Note**: `BLOCKED` is an admin-set status — blocked URLs cannot be modified
     or deleted by the owner. `EXPIRED` URLs (auto-set on max clicks or expiry
-    time) can be reactivated by setting status back to `ACTIVE`.
+    time) can be reactivated by setting status back to `ACTIVE` — but only
+    after the expiry condition is lifted (raise/clear `max_clicks`, extend/
+    clear `expire_after` via PATCH), otherwise the URL immediately reads
+    as expired again.
 
     **Use Cases**:
 
--- a/schemas/dto/responses/url.py
+++ b/schemas/dto/responses/url.py
@@ -74,7 +74,13 @@ class UrlResponse(ResponseBase):
         description="Creation time as Unix timestamp.",
         examples=[1704067200],
     )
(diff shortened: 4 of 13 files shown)


Zingzy (commented):
## Review: enforce time-based URL expiry at resolve + derive effective status

**Mergeable, no blockers.** Faithful implementation of `thoughts/time-based-expiry-trd.md` — I checked the code against the (validated) spec section-by-section and traced the one thing that could hide a real bug. `MERGEABLE`/`CLEAN`, full CI green (3.10–3.14, `test-api`, formatting, CodeQL, CodeRabbit), 0 behind main, no unresolved threads.

**Faithful to the design, verified in code (not trusted):**
- `UrlV2Doc.effective_status`, `expire_if_time_reached` (ACTIVE-guarded atomic flip), `_raise_if_time_expired` on both resolve paths, the DTO swaps, the `$and`-composed filter clause + `_derived_expiry_arms`, the v1 tz-ambiguity rule, and the `_auto_reactivate` tightening all match the TRD exactly.
- **The host-TZ point is genuinely fixed, not just claimed:** `to_unix_timestamp` normalizes naive→UTC before `.timestamp()`, so the cache `expiration_time` and the model property agree regardless of host TZ — and the *old* `from_v2_doc` (`int(naive.timestamp())`) was the latent host-TZ-dependent path this PR retires. The "four expressions agree" invariant is guarded by the `TestEffectiveStatusClause` predicate-pin matrix.
- Correctness spot-checks pass: `ObjectId(data.id)` only runs on the v2 branch (never a v1 short-code); `data.url_status == UrlStatus.ACTIVE` is the same str-enum equality the existing non-active checks rely on; concurrent expired hits all 410 (derive-on-read) with the flip idempotent; post-flip reads stay cache-hit O(1) (populate-before-raise).

**Merge safety:** no feature flag — correct and well-justified (bugfix restoring documented behavior; flagging would mean maintaining both code paths of a lie). Ships on release. **The one thing to keep on the radar:** this makes the *origin* enforce expiry while the *edge* still doesn't, so the TRD §11 follow-up (`expires_at` in the KV contract + worker passthrough-on-expired) is now load-bearing — land it promptly after this deploys, since until then an edge-promoted link that gains/reaches expiry keeps redirecting for up to the KV TTL (≤~6 min, same staleness class as editing `long_url` on a promoted link; consciously deferred and bounded, so not a blocker here). Post-release, watch `url_resolve_expired_time` volume + the 410 rate.

**Notably good:** the regression archaeology (pinned the 2024→2025 drop to `0c64b08`), one derivation with a matrix test forbidding drift between the model property and the Mongo clause, deriving-on-read + opportunistic-flip so there's no scheduler/index/edge change — and it quietly fixes the pre-existing host-TZ bug in the cache converter.

PR #247, opened 2026-07-13
fix(stats): null avg_redirection_time when no measurements exist

Zingzy:
## What changed

`avg_redirection_time` conflated "no measurement" with "instant redirect": a range with zero clicks returned `0`. The semantic rule is now: **null = no measurement in the range, a number = a real measured average.** `0` is never invented.

- `schemas/dto/responses/stats.py` - `StatsSummary.avg_redirection_time` is `float | None` (default `None`).
- `services/stats_service.py` - the empty-summary default is `None`, and `$avg`'s `None` (no clicks in range, or clicks that carried no `redirect_ms`) passes through instead of being coalesced to `0`. This covers both `GET /api/v1/stats` and the v2 branch of the public stats endpoint, which reuse the same `compute()` seam.
- `services/public_stats_service.py` - the v1/emoji synthesis maps a falsy stored `average_redirection_time` to null: the legacy schema stores `0.0` for "never measured", so it cannot be distinguished from a true zero average (which real redirects never produce).
- `services/export/formatters.py` - the CSV summary writes an empty cell instead of a fabricated `0`; JSON/XML/XLSX already serialize `None` honestly.
- Legacy Flask-era surfaces (`/stats/{code}` JSON + templates) intentionally keep their numeric `0`: that wire is shipped and old clients do not null-guard. Pinned by a new test.

## Blast radius

Additive-null on a wire field. The dashboard adapter already null-guards client-side (`lib/api/stats.ts` in the frontend repo), and the public stats page consumes the same summary shape via `lib/api/public-stats.ts`. Other server-side consumers found and handled: the export formatters (above) and the legacy routes (unchanged by design). No other reader of the field exists in the repo.

## Tests

139 passed across the affected suites (`uv run pytest -q`):

- unit: zero-clicks range yields null; clicks-without-measurement yields null; a real average still rounds (`120.456 -> 120.46`); DTO accepts and serializes null.
- integration (public stats, real StatsService machinery): v2 empty `$facet` window yields null; v1 docs with a missing field or the stored `0.0` default yield null; the existing `14.24` measured-value pin still passes.
- integration (legacy): missing doc field still surfaces as numeric `0` on the legacy JSON wire.
- export: CSV writes an empty cell for null.

Both lint gates pass: `ruff check` and `ruff format --check`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/schemas/dto/responses/stats.py
+++ b/schemas/dto/responses/stats.py
@@ -27,7 +27,9 @@ class StatsSummary(ResponseBase):
     unique_clicks: int
     first_click: datetime | None = None
     last_click: datetime | None = None
-    avg_redirection_time: float
+    # null = no redirect measurements in the range (never 0 — a zero would
+    # read as an instant redirect, not as absence of data)
+    avg_redirection_time: float | None = None
 
 
 class StatsTimeRange(ResponseBase):
--- a/services/export/formatters.py
+++ b/services/export/formatters.py
@@ -75,7 +75,8 @@ class CsvFormatter:
                 ("unique_clicks", summary.get("unique_clicks", 0)),
                 ("first_click", summary.get("first_click", "")),
                 ("last_click", summary.get("last_click", "")),
-                ("avg_redirection_time", summary.get("avg_redirection_time", 0)),
+                # null = no measurement — an empty cell, never a fabricated 0
+                ("avg_redirection_time", summary.get("avg_redirection_time", "")),
             ]
             _write_rows(zf, "summary.csv", summary_rows)
 
--- a/services/public_stats_service.py
+++ b/services/public_stats_service.py
@@ -364,7 +364,14 @@ class PublicStatsService:
             "unique_clicks": len(doc.ips or []),
             "first_click": None,  # not stored on v1 docs
             "last_click": self._stats.to_user_tz(last_click, tz_name),
-            "avg_redirection_time": round(doc.average_redirection_time or 0, 2),
+            # The legacy schema can't tell "never measured" from a true
+            # 0.0 average (it stores 0.0 for both), so falsy maps to null
+            # — same semantic as v2: null = no measurement, never 0.
+            "avg_redirection_time": (
+                round(doc.average_redirection_time, 2)
+                if doc.average_redirection_time
+                else None
+            ),
         }
 
         response: dict[str, Any] = {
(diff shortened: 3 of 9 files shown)


Zingzy (commented):
## Review: null avg_redirection_time when no measurements exist

**Mergeable, no blockers.** Textbook small semantic fix — `null` = no measurement, a number = a real average, `0` never invented — with the compat boundary drawn in exactly the right place. I traced every reader of the field and verified the one thing that could bite (a frontend consumer typed non-null). `MERGEABLE`/`CLEAN`, full CI green (3.10–3.14, `test-api`, formatting, CodeQL, CodeRabbit), 0 behind main.

**Verified, not trusted:**
- **Every reader handled.** Modern surfaces (`StatsService.compute` → `GET /api/v1/stats` + public v2, and the v1/emoji synthesis) now emit `null`; the CSV formatter writes an empty cell (the only export path that defaulted to `0` — JSON/XML/XLSX serialize `None` as-is); the legacy Flask routes/helpers (`routes/legacy/stats.py`, `legacy/helpers.py` `$ifNull … 0`) are **intentionally** left numeric-`0` and pinned by a test. No other server-side reader exists.
- **The cross-repo consumer is safe** — checked the frontend directly: `lib/api/stats.ts` already types `avg_redirection_time: number | null` and `components/stats-public/stats-view.tsx` null-guards at the render site (`!= null ? \`${…}ms\` : …`). The nullable wire lands cleanly on both the dashboard adapter and the public-stats page.
- **Semantics are correct at the source:** Mongo `$avg` yields `null` for an empty range or clicks without `redirect_ms`, passed through untouched; the v1 falsy→null map is sound because the modern avg is in ms (a real redirect is ≥1ms, so a true `0` can't occur) and the legacy schema stores `0.0` for "never measured" indistinguishably from a true zero.

**Compat boundary (the right call):** keeping the legacy `/stats/{code}` JSON at numeric `0` is correct, not a gap — that wire is shipped and old clients don't null-guard; only the modern surfaces (which the dashboard/public pages consume, and which already type nullable) move. Documented and test-pinned on both sides of the line.

**Merge safety:** additive-null on a wire field, no flag needed, ships on release; the sole consumer that mattered (the frontend) already handles it. Good to merge.

**Notably good:** the reader census in the PR body matches what's actually in the tree, the modern/legacy split is deliberate and pinned from both directions, and the tests cover the real distinctions (zero-clicks→null, clicks-without-measurement→null, real avg still rounds, DTO serializes null, legacy stays `0`, CSV empty cell).

PR #248, opened 2026-07-13
feat(api): add GET /api/v1/apps connected-apps list

Zingzy:
Adds the missing API for the dashboard Apps page: the user's own device-grant list as JSON. Today only the Jinja apps page renders grants; the Next frontend adapter (lib/api/apps.ts) swallows the 404 into an empty list, so the page always shows nothing connected.

## Wire

```
GET /api/v1/apps
Authorization: Bearer <jwt>   # or access_token cookie

200 {
  "items": [
    {
      "id": "665f1c2ab7e94d0c8a1f2b3c",
      "app": "spoo-cli",
      "app_name": "Spoo CLI",
      "icon": "spoo-cli.svg",
      "permissions": ["Access your spoo.me account", "View your analytics"],
      "granted_at": "2026-06-10T12:00:00+00:00",
      "last_used_at": "2026-07-12T09:30:41+00:00"
    }
  ]
}
```

Active grants only, newest first. `app` is the config/apps.yaml registry key (shares a namespace with the frontend catalogue slugs) and is the `app_id` handle for the existing revoke endpoint. The wire carries the registry's consent `permissions` strings, not scope slugs: grants are not scoped (device JWTs act as the full account), and a security surface must not understate access.

## Gating

JWT-only via `require_jwt`, mirroring /api/v1/keys: an API key is a delegated credential, and letting it enumerate the account's other delegated credentials plus their revoke handles is a confused deputy. No `apps:*` scope exists and no programmatic consumer needs one. Rate limit 60/min (`Limits.APP_GRANTS_READ`), same budget as the sibling dashboard reads.

## Notes

- No new service, index, or wiring: `AppGrantRepo` and `AppRegistryDep` already exist, `find_active_for_user` is reused as is, and the `(user_id, revoked_at)` index already covers the query.
- `POST /auth/device/revoke` is untouched; the frontend mock aligns to its shipped Form contract in a spoo-landing follow-up.

## Tests

`tests/integration/api_v1/test_apps.py` (6 tests): 401 anonymous, 403 API-key auth, empty list, full registry mapping with UTC-stamped timestamps, registry-gone fallback, newest-first ordering.

- `uv run pytest tests/integration/api_v1/test_apps.py -q`: 6 passed
- `uv run pytest -q`: 2205 passed
- `uv run ruff check .` and `uv run ruff format --check .`: clean

## Merge safety

New endpoint, purely additive. No existing route, schema, or wire changes; nothing dark to flip. The endpoint is inert until the frontend stops mocking.

## Summary by CodeRabbit

- **New Features**
  - Added `GET /api/v1/apps` to list the authenticated caller’s connected apps (JWT-only).
  - Returns app metadata (name, icon), permissions, and grant/last-used timestamps serialized in UTC.
  - Orders results from newest to oldest and falls back to basic app info when registry details are unavailable.
  - Added a dedicated rate limit for connected-app listing.
- **Tests**
  - Added integration coverage for auth requirements, empty results, mapping, fallback behavior, ordering, and UTC timestamp serialization.

### DIFF
--- a/schemas/dto/responses/app_grant.py
+++ b/schemas/dto/responses/app_grant.py
@@ -1,0 +1,89 @@
+"""
+Response DTOs for GET /api/v1/apps.
+
+AppGrantResponse      — one connected app (active device-auth grant)
+AppGrantsListResponse — GET /api/v1/apps (200)
+
+The wire serves the spoo-landing Apps page (lib/api/apps.ts). Grants are
+not scoped — a device grant acts as the full account — so the wire carries
+the registry's consent ``permissions`` strings, never scope slugs
+(thoughts/apps-grants-api-prd.md §2).
+"""
+
+from __future__ import annotations
+
+from datetime import datetime, timezone
+
+from pydantic import Field, field_serializer
+
+from schemas.dto.base import ResponseBase
+from schemas.models.app import AppEntry
+from schemas.models.app_grant import AppGrantDoc
+
+
+class AppGrantResponse(ResponseBase):
+    """A single connected app as returned by the list endpoint."""
+
+    id: str = Field(
+        description="Grant ID (row identity — revoke keys on `app`, not this)",
+        examples=["665f1c2ab7e94d0c8a1f2b3c"],
+    )
+    app: str = Field(
+        description=(
+            "App registry key (config/apps.yaml). Shares a namespace with the "
+            "frontend catalogue slugs and is the `app_id` handle for "
+            "POST /auth/device/revoke."
+        ),
+        examples=["spoo-cli"],
+    )
+    app_name: str = Field(
+        description="Display name from the registry (falls back to `app`)",
+        examples=["Spoo CLI"],
+    )
+    icon: str | None = Field(
+        default=None,
+        description="Registry icon filename, or null when the entry is gone",
+        examples=["spoo-cli.svg"],
+    )
+    permissions: list[str] = Field(
+        description="Consent-screen permission strings the user granted",
+        examples=[["Access your spoo.me account", "View your analytics"]],
+    )
+    granted_at: datetime = Field(description="When access was granted (ISO 8601 UTC)")
+    last_used_at: datetime | None = Field(
+        default=None,
+        description="Last token exchange/refresh by this app, null if never used",
+    )
+
+    @field_serializer("granted_at", "last_used_at")
+    def _ser_as_utc(self, dt: datetime | None) -> str | None:
+        # PyMongo returns naive datetimes; without explicit tzinfo the JSON
+        # form omits the offset and clients parse it as local time. Stamp
+        # UTC so the wire format is unambiguous (`...+00:00`).
+        if dt is None:
+            return None
+        if dt.tzinfo is None:
+            dt = dt.replace(tzinfo=timezone.utc)
+        return dt.isoformat()
+
+    @classmethod
+    def from_grant(cls, doc: AppGrantDoc, entry: AppEntry | None) -> AppGrantResponse:
+        # A grant can outlive its registry entry; it still holds live tokens,
+        # so it is listed with fallback labels rather than hidden.
+        return cls(
+            id=str(doc.id),
+            app=doc.app_id,
+            app_name=entry.name if entry else doc.app_id,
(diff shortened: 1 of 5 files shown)


schemas/dto/responses/app_grant.py
```diff
@@ -0,0 +1,89 @@
...
+    @classmethod
+    def from_grant(cls, doc: AppGrantDoc, entry: AppEntry | None) -> AppGrantResponse:
+        # A grant can outlive its registry entry; it still holds live tokens,
+        # so it is listed with fallback labels rather than hidden.
+        return cls(
+            id=str(doc.id),
+            app=doc.app_id,
+            app_name=entry.name if entry else doc.app_id,
+            icon=entry.icon if entry else None,
+            permissions=list(entry.permissions) if entry else [],
```
Zingzy: **[Nit / by-your-own-principle]** For a grant whose registry entry is gone, `permissions` falls back to `[]`. But the DTO docstring's rule is "never understate access" — and this grant is *still full-account* (an active device JWT), it just lost its catalogue entry. An empty permissions list reads as "harmless / no access" in the UI, which understates it exactly when the app is unknown/removed (arguably the case a user most wants to revoke). Consider a generic fallback string (e.g. `["Full access to your spoo.me account"]`) so the degraded case still errs toward over-stating access. Rare (app pulled from `apps.yaml`) and the app is still listed + revocable, so low priority — but it's the one spot the fallback contradicts the field's stated intent.

Zingzy (commented):
## Review: GET /api/v1/apps connected-apps list

**Mergeable, no blockers.** Clean, purely-additive read endpoint over existing infra, with the security-sensitive parts done right. I traced the gating and the query rather than trusting the description. Every real CI check is green (3.10–3.14, `test-api`, formatting, CodeQL); `UNSTABLE`/`MERGEABLE` is just CodeRabbit still pending. 0 behind main.

**Verified, not trusted:**
- **Gating is correct and correctly reasoned.** `JwtUser` = `Annotated[CurrentUser, Depends(require_jwt)]`, so an API key can't enumerate the account's other delegated credentials + their revoke handles — the confused-deputy argument holds and mirrors `/api/v1/keys`. Tests pin 401 anon / 403 API-key.
- **No IDOR.** `find_active_for_user` queries `{user_id, revoked_at: None}` with `user_id` taken from the JWT (unspoofable) and active-only, so no cross-user or revoked-grant leak. Reuses the existing `(user_id, revoked_at)` index.
- **Sort is safe.** `AppGrantDoc.granted_at` is a non-optional `datetime`, so `grants.sort(key=lambda g: g.granted_at)` can't hit `None`.
- **No secrets on the wire** — no tokens, no scope slugs; the exposed `id` is the user's own grant row-id (revoke keys on `app`), non-capability.

**Nits (neither blocks):**
1. **`/api/v1/apps` isn't added to `tests/smoke/test_routes_registered.py`.** That smoke test is the repo's "is the router wired?" guard, and #242 added its new routes (`/api/v1/reports`, `/api/v1/contact`) to both the path list and the method list. It's a subset check (`assert path in paths`), so CI stays green without it — but for consistency and to guard against the route silently dropping later, add `("/api/v1/apps", "GET")` there. `test_apps.py` covers behavior; this is the wiring-guard gap. (inline below)
2. Registry-gone fallback sets `permissions: []` — see inline; slightly understates access by the DTO's own principle.

**Merge safety:** additive new endpoint, JWT-only, no existing route/schema/wire touched, inert until the frontend stops mocking. No flag needed; ships on release.

**Notably good:** reuses the existing repo/dep/index (no new service or wiring), the confused-deputy gating with a written rationale, graceful registry-gone handling (a grant that outlived its registry entry still lists so it can be revoked), UTC-stamped timestamps for an unambiguous wire, consent-strings-not-scope-slugs (never understates), and 6 tests covering the full security matrix.

schemas/dto/responses/app_grant.py
Zingzy: Agreed, the fallback contradicted the field's own rule exactly where it matters most. Fixed in cea956c: registry-gone grants now carry ["Full access to your spoo.me account"], test updated to pin it. Wording matches the apps.yaml consent-string voice.

PR #249, opened 2026-07-13
feat: profile editing (user_name update, picture upload and unset)

Zingzy:
Two small account-profile features: post-registration display name editing and custom profile picture upload with unset.

### PATCH /auth/me

```
PATCH /auth/me
{"user_name": "Jane Doe"}     -> 200 {"user": {...UserProfileResponse, "user_name": "Jane Doe"}}
{"user_name": null}           -> 200, clears the name
{}                            -> 422 (field required; an accidental empty PATCH must not clear)
{"user_name": "x" * 256}      -> 422
```

Returns the same `MeResponse` shape as `GET /auth/me`, so the frontend can write it straight into the session cache. Normalization mirrors register exactly: strip, whitespace-only becomes null. Bounds mirror `RegisterRequest` (1..255) since these are the two write paths for the same field.

Gating: `JwtUser` (same `require_jwt` dependency keys.py uses), so an API key cannot rename the account. Rate limit 10/min (`PROFILE_UPDATE`).

### POST /dashboard/profile-pictures/upload

```
POST /dashboard/profile-pictures/upload
{"image": "data:image/png;base64,..."}   -> 200 {"message": "Profile picture updated successfully"}
bad type / oversized / bad base64        -> 400 {"error": ..., "code": "validation_error", "field": "image"}
R2 not configured (self-host)            -> 400 with a clear message
```

### DELETE /dashboard/profile-pictures

```
DELETE /dashboard/profile-pictures       -> 200 {"message": "Profile picture removed"}
```

Idempotent; sets `pfp` back to null so the initials avatar takes over. Both live next to the existing GET/POST `/dashboard/profile-pictures` pair with the same auth (`AuthUser`) and the same message response shape, so the Next settings page AvatarRow extends without new conventions.

## R2 reuse

The upload path reuses the meta-tags (PR #231) plumbing wholesale. The data-URI decode, base64 pre-decode gate, decoded size cap (`R2_UPLOAD_MAX_BYTES`, default 512KB), and magic-byte MIME match were extracted from `services/meta_tags/images.py` into `services/image_ingest.py` (pure move, no behavior change; the og ingest tests pass unchanged). Objects are stored content-addressed under `profile-pictures/{hmac-owner-prefix}/{sha256}.{ext}` mirroring the `og/` layout, via the same `R2StorageClient.put_object` (immutable cache headers, inline disposition, nosniff). Takedowns can prefix-sweep `profile-pictures/{prefix}/` the same way as `og/`; replaced objects are not deleted (same orphan-GC stance).

`pfp.source` gains an `upload` variant (`OAuthProvider | Literal["upload"]`) rather than widening the OAuth enum, which is also used to validate provider path params.

## Abuse posture note

Meta-tag images run two write-time checks: the ingest gates above, plus a blocklist regex scan over the user-supplied title/description/image URL text. Profile pictures get the identical ingest gates. The blocklist scan does not apply here because there is no user-supplied text or URL: the only stored value is the R2 URL we minted. No moderation step was invented beyond what the meta-tags path runs; if image-content scanning lands for og uploads later, both paths now share one ingest module to hang it on.

## Tests

- `tests/unit/services/test_profile_picture_service.py`: update_user_name set/clear/not-found; upload with real PNG bytes through the actual decode path (stores R2 URL with `source: "upload"`, key under `profile-pictures/`), rejects non-data-URI, MIME mismatch, oversized, unconfigured storage, unknown user; unset clears and is idempotent.
- `tests/integration/test_auth_routes.py`: PATCH /auth/me sets and strips, null clears, whitespace normalizes to null, 422 on missing/overlong, 401 unauthenticated, 403 via API key.
- `tests/integration/test_dashboard_routes.py`: upload 401/422/200/400(+field echo), delete 401/200.
- Full suite: 2224 passed. `ruff check` and `ruff format --check` clean.

## Merge safety

Additive only: no existing endpoint, response shape, or stored document changes. `ProfilePictureService` gains keyword-only constructor args with defaults (unconfigured R2 keeps the old constructor behavior; upload then 400s with a clear message, matching the meta-tags self-host degradation contract). The meta-tags refactor is a pure extraction with its existing tests green. Existing `pfp` documents deserialize unchanged; the `source` union only widens. Nothing here is gated by a flag because both surfaces are dark until the frontend wires them.

### DIFF
--- a/routes/dashboard_routes.py
+++ b/routes/dashboard_routes.py
@@ -23,8 +23,8 @@ from pydantic import BaseModel, Field
 from dependencies import (
     AppGrantRepo,
     AppRegistryDep,
-    AuthUser,
     CurrentUser,
+    JwtUser,
     OptionalUser,
     ProfilePictureSvc,
 )
@@ -265,7 +265,7 @@ class ProfilePictureMessageResponse(BaseModel):
 @limiter.limit(Limits.DASHBOARD_READ)
 async def get_profile_pictures(
     request: Request,
-    user: AuthUser,
+    user: JwtUser,
     svc: ProfilePictureSvc,
 ) -> AvailablePicturesResponse:
     pictures = await svc.get_available_pictures(user.user_id)
@@ -277,7 +277,7 @@ async def get_profile_pictures(
 async def set_profile_picture(
     request: Request,
     body: SetProfilePictureRequest,
-    user: AuthUser,
+    user: JwtUser,
     svc: ProfilePictureSvc,
 ) -> ProfilePictureMessageResponse:
     await svc.set_picture(user.user_id, body.picture_id)
@@ -289,7 +289,7 @@ async def set_profile_picture(
 async def upload_profile_picture(
     request: Request,
     body: UploadProfilePictureRequest,
-    user: AuthUser,
+    user: JwtUser,
     svc: ProfilePictureSvc,
 ) -> ProfilePictureMessageResponse:
     await svc.upload_picture(user.user_id, body.image)
@@ -300,7 +300,7 @@ async def upload_profile_picture(
 @limiter.limit(Limits.PROFILE_PICTURE_SET)
 async def unset_profile_picture(
     request: Request,
-    user: AuthUser,
+    user: JwtUser,
     svc: ProfilePictureSvc,
 ) -> ProfilePictureMessageResponse:
     await svc.unset_picture(user.user_id)
--- a/routes/dashboard_routes.py
+++ b/routes/dashboard_routes.py
@@ -10,6 +10,8 @@ GET  /dashboard/billing     → billing page
 GET  /dashboard/apps          → connected apps + ecosystem
 GET  /dashboard/profile-pictures  → available pictures (JSON)
 POST /dashboard/profile-pictures  → set profile picture (JSON)
+POST /dashboard/profile-pictures/upload  → upload a custom picture (JSON)
+DELETE /dashboard/profile-pictures       → unset picture (JSON)
 """
 
 from __future__ import annotations
@@ -243,6 +245,14 @@ class SetProfilePictureRequest(BaseModel):
     picture_id: str = Field(min_length=1, max_length=200)
 
 
+class UploadProfilePictureRequest(BaseModel):
+    # Size/type gates run in the service against the configured cap
+    # (R2_UPLOAD_MAX_BYTES) — same posture as meta-tag image uploads.
+    image: str = Field(
+        min_length=1, description="base64 data URI (image/png, image/jpeg, image/webp)"
+    )
+
+
 class AvailablePicturesResponse(BaseModel):
     pictures: list[AvailablePicture]
 
@@ -272,3 +282,26 @@ async def set_profile_picture(
 ) -> ProfilePictureMessageResponse:
     await svc.set_picture(user.user_id, body.picture_id)
     return ProfilePictureMessageResponse(message="Profile picture updated successfully")
+
+
+@router.post("/profile-pictures/upload")
+@limiter.limit(Limits.PROFILE_PICTURE_UPLOAD)
+async def upload_profile_picture(
+    request: Request,
+    body: UploadProfilePictureRequest,
+    user: AuthUser,
+    svc: ProfilePictureSvc,
+) -> ProfilePictureMessageResponse:
+    await svc.upload_picture(user.user_id, body.image)
+    return ProfilePictureMessageResponse(message="Profile picture updated successfully")
+
+
+@router.delete("/profile-pictures")
+@limiter.limit(Limits.PROFILE_PICTURE_SET)
+async def unset_profile_picture(
+    request: Request,
+    user: AuthUser,
+    svc: ProfilePictureSvc,
+) -> ProfilePictureMessageResponse:
+    await svc.unset_picture(user.user_id)
+    return ProfilePictureMessageResponse(message="Profile picture removed")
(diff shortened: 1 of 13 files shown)


routes/dashboard_routes.py
```diff
@@ -272,3 +282,26 @@ async def set_profile_picture(
...
     await svc.set_picture(user.user_id, body.picture_id)
     return ProfilePictureMessageResponse(message="Profile picture updated successfully")
+
+
+@router.post("/profile-pictures/upload")
+@limiter.limit(Limits.PROFILE_PICTURE_UPLOAD)
+async def upload_profile_picture(
+    request: Request,
+    body: UploadProfilePictureRequest,
+    user: AuthUser,
```
Zingzy: **[Nit / confirm intent]** Picture upload/set/delete gate on `AuthUser` (API keys allowed), but `PATCH /auth/me` (display name) gates on `JwtUser` (`routes/auth/routes.py:279`, API keys barred) — yet both are "profile editing." The split is defensible: pictures mirror the existing `/dashboard/profile-pictures` pair, and name-edit mirrors `/keys`' JWT-only stance. But the net effect is an API key can change/remove the account's avatar while it can't rename it, which is a slightly odd asymmetry for two cosmetic profile fields. No security hole (avatar is cosmetic), just worth a conscious "yes, intended" — and if it should match, note that tightening the picture endpoints to `JwtUser` would change the already-shipped set-picture behavior.

Zingzy (commented):
## Review: profile editing (name update, picture upload + unset)

**Mergeable, no blockers.** Thorough pass, with the R2-optional degradation traced end-to-end since that was the ask. Three features + a clean extraction of the shared image-ingest gates. `MERGEABLE`/`CLEAN`, full CI green (3.10–3.14, `test-api`, formatting, CodeQL, CodeRabbit). Note it's **10 behind main** (clean, no conflict) — optional update-branch.

**"Everything is optional / R2" — verified, it holds.** This was the thing to get right and it's correct:
- Wiring builds `r2_storage` only when R2 is fully configured (all five fields); partial/absent → `None` + a log, exactly like the meta-tags path.
- `upload_picture` guards `self._r2_storage is None or not self._r2_storage.is_configured` and raises a clean `ValidationError` → **400 "Image upload is not available on this deployment"**, never a 500. `R2StorageClient.is_configured` exists (`infrastructure/storage/r2.py:67`), so that guard can't `AttributeError`. This is byte-for-byte the shipped meta-tags self-host degradation contract.
- Name-edit, OAuth-picture-select, and delete touch no R2 and work without it. So a self-host without R2 loses *only* custom upload (with a clear message) — nothing else. Textbook optional.

**Other things I verified rather than trusted:**
- **The extraction is a genuine pure move.** `split_image_data_uri` / `decode_image_data_uri` / `owner_key_prefix` moved verbatim from `meta_tags/images.py` into `services/image_ingest.py` (same regex, gates, error messages, `field=`), and the surrounding flow (https passthrough, R2 guard) stays in `ingest_meta_image`. Both surfaces now share one gate module and the same HMAC key-prefix (`settings.secret_key`), so a future content-scanner hangs on one place. og tests green.
- **No response-shape divergence on PATCH /auth/me.** `fetch_user_profile` returns a `UserDoc`, and PATCH feeds `from_user` a `model_copy`'d `UserDoc` — same type, same shape — so "write straight into the session cache" is accurate, and the model_copy correctly avoids a re-read for a single-field update.
- **Upload security is solid and actually *safer* than meta-tags:** data-URI-only (no https passthrough), so there's **no SSRF surface** here; magic-byte MIME match is the gate (png/jpeg/webp only), size-capped pre- and post-decode, content-addressed key under an HMAC owner prefix (no ObjectId leak), and served from R2 with `nosniff` + inline disposition + immutable cache. No stored-XSS path.
- **`source` widened as a union, not the enum** — `OAuthProvider | Literal["upload"]` — deliberately leaving `OAuthProvider` (which validates provider path params) untouched, so `"upload"` can't leak into provider validation. Right call, and existing `pfp` docs deserialize unchanged.

**Nits (neither blocks):**
1. **Auth split** (inline): name-edit is `JwtUser` (API key barred), but picture upload/set/delete are `AuthUser` (API key allowed) — both are "profile editing." Defensible by precedent, but worth confirming it's intentional that an API key can change/remove the avatar but not the display name.
2. The three new routes (`PATCH /auth/me`, `POST /dashboard/profile-pictures/upload`, `DELETE /dashboard/profile-pictures`) aren't added to `tests/smoke/test_routes_registered.py` — same as #248, it's a subset check so CI stays green, but that's the repo's router-wiring guard; adding them (with methods) keeps it honest. Optional.

**Merge safety:** additive only — no existing endpoint/shape/document changes, constructor args are keyword-only-with-defaults (unconfigured R2 keeps old behavior), the meta-tags refactor is pure. Dark until the frontend wires them; no flag needed.

**Notably good:** the pure-move extraction sharing the security posture (not a drift-prone re-implementation), the optional-R2 handling matching the reference contract exactly, the data-URI-only upload closing the SSRF surface, the union-not-enum choice for `source`, and tests that push real PNG bytes through the actual decode path plus the unconfigured-R2 case.

routes/dashboard_routes.py
Zingzy: Aligned rather than asymmetric: all four profile-picture endpoints (the shipped GET/POST pair plus the new upload/delete) now gate on JwtUser, matching PATCH /auth/me. One rule for the whole account-profile surface: session required, API keys manage resources. Blast radius of tightening the shipped pair: the endpoints are dashboard-internal and cookie-called (Next settings page, Jinja dashboard); an API-key caller of a provider-picture picker is not a real client. Fixed in 7ca0fbb plus a format follow-up, new test pins 403 across all four.

PR #251, opened 2026-07-13
feat: emoji short links on the v2 API

Zingzy:
## What

Brings emoji short links (`spoo.me/🚀🔥`) into the v2 system as first-class citizens. Until now they only existed in the legacy `POST /emoji` flow — new emoji links get the full v2 feature set: ownership, link management, custom domains, expiry, click limits, passwords, v2 click pipeline, stats/exports, and Redis + edge caching.

No new endpoint: an emoji link is the same resource as any link, so `POST /api/v1/shorten` accepts emoji in `alias`, and a new `alias_type: "emoji"` field selects the auto-generation charset when `alias` is omitted (same one-endpoint-plus-field shape Dub and Bitly use). `check-alias`, `PATCH /urls/{id}`, list, stats, and exports all work unchanged.

## Why the old "any emoji" rule was wrong

v1 accepted anything the `emoji` library could parse. Research says most of that space is broken in practice:

| Category | Failure mode | Verdict |
|---|---|---|
| Single codepoint, emoji-presentation, ≤ E12.0 | none — stable everywhere | ✅ generate + accept |
| Single codepoint, E12.1–15.1 | tofu on Windows 10 (font frozen at E12.0, May 2019) | ✅ accept only |
| Skin-tone sequences (👍🏽) | degrade to 2 glyphs on old platforms, byte-stable | ✅ accept |
| VS16-dependent (❤️ ☺️) | keyboards/copy-paste drop or add the invisible `U+FE0F` → different bytes → 404 | ❌ |
| ZWJ sequences (👨‍👩‍👧) | split into constituent glyphs on unsupported platforms | ❌ |
| Keycaps (1️⃣) | raw digit + VS16 + U+20E3, byte-fragile | ❌ |
| Flags (🇺🇸) | render as bare letters in every Chromium browser on Windows (Segoe UI Emoji has no flag ligatures) | ❌ |
| Tag sequences (🏴󠁧󠁢󠁥󠁮󠁧󠁿) | invisible tag chars; arbitrary-data smuggling vector | ❌ |

Auditing the old hand-curated `data/emojis.json` (1590 entries used for generation) against Unicode data: **106 unqualified, 42 invalid RGI sequences, 172 ZWJ, 215 VS16-dependent, 12 keycaps, 107 too new for Windows 10**. Some of those generated aliases were unroutable the moment they were minted.

## How

- **`shared/emoji_policy.py`** — single source of truth. Accepts RGI fully-qualified graphemes that are single-codepoint or base+skin-tone, version-capped (`EMOJI_ACCEPT_MAX_VERSION=15.1`, generation capped at 12.0 = the Windows 10 glyph ceiling). Canonicalization = unquote → NFC → strip VS16, applied at create **and** resolve so `⭐️` and `⭐` are the same link.
- **Derived generation pool** — `data/emojis.json` is deleted; the pool (1058 emoji) derives from the pinned `emoji==2.15.0` package data at import. A unit test asserts every pool entry passes the acceptance policy, so a future package bump fails loudly instead of drifting. `scripts/audit_emoji_pool.py` prints the pool + version histogram for review.
- **Storage** — ordinary `urlsV2` rows under the canonical alias; the existing `(domain, alias)` unique index covers uniqueness. The legacy `emojis` collection becomes read-only-legacy; **no data migration**.
- **Resolution** — `resolution_order` now returns `(V2, EMOJI)` for emoji codes: `urlsV2` first (canonical form), then legacy `emojis` (raw bytes first — exact legacy semantics — then canonical). Shared by the redirect and the public preview/stats resolver so surfaces can't drift. Emoji detection is canonicalization-aware: redundant-VS16 forms like `⭐️` aren't even recognized by `emoji.emoji_list`, so raw-only detection would misroute exactly the variants canonicalization exists to rescue.
- **Cross-system uniqueness, both directions** — v2 creation collision-checks the legacy collection **VS16-insensitively** (a legacy `⭐️🎉` blocks a new canonical `⭐🎉`, so a v2 hit can never shadow a live legacy link), and legacy `POST /emoji` now checks `urlsV2` through the same `check_alias_available`.
- **DTO/service split** — mixed/garbage aliases fail as 422 at the DTO (like the old pattern violations); policy failures (flags, ZWJ, version caps) are service-side 400s because the caps are settings-configurable. `check-alias` gains an `emoji_policy` reason. `UpdateUrlRequest.alias` gets the same treatment (and gains validation the old DTO pattern was silently providing).
- **Legacy `POST /emoji`** — wire contract frozen (`emojies` param, `EmojiError` shapes, raw-emoji `short_url`), but new creations are held to the policy and its generator now draws from the safe pool. The permissive `validate_emoji_alias` is removed; `is_emoji_alias` stays deliberately lenient for resolution so every historical link keeps resolving.
- **Tenant middleware** — the custom-domain alias gate was an incomplete 5-block emoji allowlist (missed ⭐, dingbats, skin tones, and more); it's now a deliberately coarse "plausibly a short code" pattern with the policy staying at creation. Emoji aliases work on custom domains.
- **Edge cache** — no code changes needed (keys already store raw decoded characters); added a skin-tone contract fixture that both the pytest and vitest suites pick up. Hot v2 emoji links promote to the edge like any other link.
- Dashboard JS alias gates, k6 setup (v2 emoji creation via `alias_type`), and OpenAPI descriptions updated to match.

### Notable bug caught during verification

MongoDB `$regex` matches bytewise, so a bare `️?` in the VS16-insensitive collision pattern made only the selector's **last UTF-8 byte** optional — the check silently never matched selector-free ids. Python's per-codepoint `re` hid it in unit tests; a full-stack run against real Mongo caught it. Fixed by grouping (`(?:️)?`) and pinned with a repository test.

## Testing

- 2423 pytest (unit + integration + smoke) green; 70 edge-worker vitest contract tests green; ruff clean.
- Full docker-compose e2e: create with VS16-variant alias → canonical stored; percent-encoded and VS16-variant redirect paths both 302; `alias_type: "emoji"` generation; policy rejections (flags/ZWJ/keycaps → 400, mixed → 422); `check-alias` reason matrix; legacy `/emoji` rejects a flag and still serves its frozen JSON shape; cross-system collisions 409 in both directions; legacy VS16-stored ids still resolve by their exact raw path.

## Rollout

Ships ungated: the change strictly widens accepted input, resolution changes are backward-compatible (v2 emoji rows can't exist before creation lands), and there's no data migration or index work. Legacy behavior change is creation-only tightening — resolution of existing links is untouched.

## Summary by CodeRabbit

* **New Features**
  * Added support for emoji-only short aliases, including automatic emoji alias generation.
  * Emoji aliases are normalized consistently, including variation-selector handling and percent-encoded URLs.
  * Added configurable limits for accepted emoji versions, generated emoji versions, and alias length.

* **Bug Fixes**
  * Improved alias conflict detection across legacy and current URL systems.
  * Added clearer validation results and messages for unsupported emoji, invalid formats, and length limits.
  * Improved resolution of emoji aliases across URL formats.


services/url_service.py
```diff
@@ -432,6 +454,12 @@ async def resolve(
         """
         scope = domain or self._system_default_domain
         is_custom = scope != self._system_default_domain
+        # Emoji codes can arrive as byte-variant forms (stray VS16 from
+        # keyboards / copy-paste). Canonicalize once here so the cache key
+        # and every v2 lookup use one form; the raw form is kept for exact
+        # legacy ``emojis`` ``_id`` matches inside _dispatch.
+        raw_code = short_code
+        short_code = v2_lookup_code(short_code)
```
Zingzy, line 462: **Should-fix: click events carry the request-form code, and the variant-rescue paths silently lose them.**

`resolve()` can now succeed when the request form differs from the stored identity, but the redirect route still emits `ClickEvent(short_code=<raw request form>)`. Two concrete failures:

1. **Legacy emoji hit via the canonical fallback candidate** (`_dispatch` loop below): the legacy click handler does `emoji_repo.update(short_code, ...)` with the raw form, the `_id` filter matches nothing, and the click is silently dropped. `total-clicks` never increments, so analytics are lost and max-clicks enforcement never trips for variant-form traffic. Before this PR those requests 404'd, so the rescue path currently under-delivers its own contract.
2. **v2 emoji link with `max_clicks`**: the cache entry lives under the canonical key (this line), but the exhaustion invalidate in the v2 click handler uses `context.short_code`. If the exhausting click arrived via a VS16-variant path, the invalidation misses and the link keeps serving from cache for up to the 300s TTL.

One-line fix at the route: emit the event with `url_data.alias` instead of the path form. `UrlCacheData.alias` is the resolved stored identity in every generation (canonical alias for v2, matched candidate for emoji via `_emoji_doc_to_cache`, exact code for v1), so both failures close at one choke point.

tests/unit/repositories/test_emoji_url_repository.py
```diff
@@ -88,6 +88,33 @@ async def test_check_exists_false(self):
...
+
+        assert await repo.check_exists_vs16_insensitive("\u2b50\U0001f389") is True
+
+        query, projection = col.find_one.await_args.args
+        pattern = query["_id"]["$regex"]
+        # The emitted pattern must match the canonical form and every
+        # VS16-decorated variant of it, and nothing else.
+        import re
+
+        compiled = re.compile(pattern)
```
Zingzy: **Should-fix: this doesn't pin the bytewise bug it exists for.**

Python `re` on `str` matches per codepoint, so the pre-fix ungrouped pattern (`\uFE0F?` without the group) also passes every assertion here. I verified: both the buggy and the fixed pattern fullmatch the canonical form under str matching; only bytes matching separates them. A regression to the ungrouped form would sail through this test.

Adding a bytes-mode assertion makes it a real pin, since bytes `re` reproduces Mongo's bytewise behavior:

```python
assert re.compile(pattern.encode()).fullmatch("\u2b50\U0001f389".encode())
```

The buggy pattern fails exactly that. Worth also adjusting the `vs16_insensitive_pattern` docstring, which currently claims the test runs the pattern "through Mongo semantics".

static/js/alias-checker.js
```diff
@@ -31,21 +31,31 @@
         return out;
     }
 
+    // Coarse gates only — the server's check-alias endpoint is the source
+    // of truth (emoji policy, reserved words, collisions).
+    const ALNUM_ALIAS = /^[a-zA-Z0-9_-]+$/;
+    const EMOJI_ALIAS = /^[\p{Extended_Pictographic}\p{Emoji_Modifier}\u200D\uFE0F\u20E3]+$/u;
```
Zingzy: Nit: `\p{Extended_Pictographic}` excludes regional indicators and the keycap base characters, so 🇺🇸 and 1️⃣ fail this gate and get the generic charset message instead of deferring to the server's `emoji_policy` reason (which has the better copy for exactly these cases). Harmless since the server is authoritative, but adding `\p{Regional_Indicator}` and `0-9#*` to the class, or switching to "contains any pictographic char" instead of full-match, routes them to the right message. Same class is duplicated in `dashboard/links.js`.

scripts/audit_emoji_pool.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env -S uv run --script
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#     "emoji==2.15.0",
```
Zingzy, line 7: Nit: this header pins `emoji==2.15.0` while pyproject declares `emoji>=2.14.1` (uv.lock resolves 2.15.0 today). Two version declarations that can drift, and the whole point of the script is inspecting the exact dataset the app runs. A comment noting it must track uv.lock would do, or drop the PEP 723 header and run it via the project env.

config.py
```diff
@@ -474,6 +474,14 @@ def blocked_self_domains(self) -> tuple[str, ...]:
...
     blocked_url_regex_timeout: float = 0.2
     max_emoji_alias_length: int = 15
+    # Newest Unicode emoji version accepted in custom emoji aliases.
+    # 15.1 renders on iOS 17.4+ / Android 14+ / Windows 11 23H2+.
+    emoji_accept_max_version: float = 15.1
+    # Cap for auto-generated emoji aliases. 12.0 is the Windows 10 glyph
+    # ceiling (Segoe UI Emoji's last update, May 2019) — generated codes
+    # must render everywhere.
+    emoji_generate_max_version: float = 12.0
+    emoji_generated_alias_length: int = 3
```
Zingzy, line 484: Nit: validated `>= 1` below, but `generate_emoji_alias_v2` raises for lengths over 15, so `EMOJI_GENERATED_ALIAS_LENGTH=16` boots fine and then every `alias_type: "emoji"` request 500s at runtime. The `_emoji_generate_within_accept` model_validator next door is the natural home for the upper bound (fail at boot, not per-request).

Zingzy (commented):
Mergeable, nothing blocking. Two should-fixes worth doing before the release that ships this: the click pipeline loses clicks on exactly the variant-rescue paths this PR introduces, and the regression test for the Mongo bytewise regex bug doesn't actually pin it. Details inline; the rest below.

## Architecture read

The load-bearing decisions all hold up:

- **`shared/emoji_policy.py` placement is right.** Pure, settings-free, knobs as parameters with service-enforced caps. Exactly how the other `shared` validators work, and `resolution_order` stays the single contract binding the redirect and the public resolver, so the surfaces can't drift.
- **The derived pool beats the checked-in artifact.** The old `data/emojis.json` shipped VS16-carrying entries (`⭐️`, `⚡️` are literally in the deleted file), meaning v1 auto-generated byte-fragile ids for years. Deriving from pinned package data with a unit test asserting generator output passes the validator turns silent drift into a loud failure. That invariant test is the best line in the PR.
- **The VS16-insensitive collision check is load-bearing, not paranoia.** I initially suspected it could never fire (custom v1 input with redundant VS16 was rejected by the old validator, I verified `emoji_list` doesn't recognize `⭐️`), but the old *generation pool* minted redundant-VS16 ids, so legacy docs whose canonical form is a valid v2 alias really exist. Good that both creation directions check it.
- **DTO 422 / service 400 split** follows the settings-free-DTO rule cleanly, and `_validate_and_canonicalize_custom_alias` being the single enforcement point for create and update closes the edit-as-side-door hole (with the canonical no-op comparison tested).
- **Cache correctness under canonical keys**: since `_populate_cache` still excludes emoji-schema hits, the canonical cache key can only ever hold a v2 row, and creation-time collision checks guarantee no live legacy variant exists when one does. I chased a shadowing scenario here (two legacy docs differing only by VS16, cache collapsing them) and it dies precisely because legacy emoji hits are never cached. Worth keeping that exclusion in mind if emoji caching is ever revisited.

## One accepted asymmetry worth stating out loud

Canonicalization rescues variant-request to canonical-stored, but not the reverse: a legacy id stored *with* VS16 (e.g. an old auto-generated `⭐️🎉`) still 404s when visited in canonical form, while simultaneously blocking that canonical form from v2 registration. So check-alias says taken, visiting says 404. Identical behavior to main (it 404'd before too) and fixing it would cost a regex scan on the hot path, so leaving it is the right call. Just noting it's a deliberate trade, not an oversight.

## Merge safety

Merging ships nothing (release-based deploys), and this ships ungated by design: strictly-widening input acceptance, no migration, no index work, no edge worker change (fixture-only, both suites pick it up). The one legacy behavior change is creation-only tightening of `POST /emoji`, resolution of existing links is untouched, and the frozen wire shape is pinned by tests. Hot path cost of the pre-cache canonicalization is about 2µs per request for alphanumeric codes (measured), noise at 400k/day. I ran the touched suites (523 tests) green in a clean worktree; CI matrix is green.

## Product lens

One-endpoint-plus-field matches the Dub/Bitly shape and keeps emoji links first-class in ownership, stats, exports, and custom domains without a parallel resource. The research-backed narrowing is the right posture for a link product: every minted-but-unrenderable alias is a broken-link impression the domain can't afford. The version caps living in settings means a future widening is an env change, not a code change.

## Nits

Inline: the JS emoji gate misses regional indicators and keycaps (cosmetic, server stays authoritative), the audit script's PEP 723 header duplicates the emoji pin and can drift from uv.lock, and `emoji_generated_alias_length` has no upper bound so a config of 16+ boots fine and then fails every generation at runtime.

Zingzy (commented):
Re-checked 49f613f. All five findings verified fixed at the choke points, not just patched around:

- **Click identity**: the event now carries `url_data.alias`, which closes both failure modes at once (legacy `_id` updates via candidate resolution, v2 max-clicks cache invalidation under the canonical key). The new integration test pins the VS16-variant case. Bonus I hadn't asked for: the hotness consumer now aggregates variant traffic under one `hot:` key and edge promotion uses the canonical alias, which matches the `short_url` the API returns. Strictly better.
- **Bytewise regex pin**: both the repository test and the new `test_matches_under_bytewise_semantics` assert in bytes mode. I confirmed earlier that the pre-fix ungrouped pattern fails bytes-mode fullmatch, so a regression now breaks the test. Docstring claim corrected too.
- **Boot guard**: verified live, `EMOJI_GENERATED_ALIAS_LENGTH=16` raises `ValidationError` at settings construction. Tying it to `min(15, max_emoji_alias_length)` is more correct than the bare 15 I suggested.
- **JS emoji class**: regional indicators and keycap bases added, and the alnum-branch-first restructure correctly handles the new `0-9` overlap so pure-digit input keeps its length feedback instead of deferring to the server. Good catch on that interaction.
- **Audit script**: comment tying the pin to uv.lock, fine.

Ran the affected suites (264 tests) green in a clean worktree; CI matrix is green at the new head. Nothing new introduced by the delta. Good to merge.

services/url_service.py
Zingzy: Fixed in 49f613f — the event now carries `url_data.alias` (resolved identity) with a comment explaining why the path form is wrong, plus an integration test pinning variant path → canonical alias in the event.

tests/unit/repositories/test_emoji_url_repository.py
Zingzy: Fixed in 49f613f — both the repository test and the policy pattern test now assert in bytes mode (`re.compile(pattern.encode())`), which reproduces Mongo's bytewise matching and fails on the ungrouped pattern. Docstring overclaim corrected too.

static/js/alias-checker.js
Zingzy: Fixed in 49f613f — added `\p{Regional_Indicator}` and `0-9#*` to the class in both files, with the alnum branch reordered first so pure-numeric input still gets instant client-side length feedback instead of deferring to the server.

scripts/audit_emoji_pool.py
Zingzy: Fixed in 49f613f — added a must-track-uv.lock comment above the pin. Kept the PEP 723 header so the script stays runnable standalone per the scripts/ convention.

config.py
Zingzy: Fixed in 49f613f — `_emoji_generate_within_accept` now also rejects `emoji_generated_alias_length > min(15, max_emoji_alias_length)`, failing at boot instead of per-request.

tests/unit/repositories/test_emoji_url_repository.py
```diff
@@ -88,6 +88,33 @@ async def test_check_exists_false(self):
...
+        assert compiled.fullmatch("\u2b50\U0001f389")
+        assert compiled.fullmatch("\u2b50\ufe0f\U0001f389")
+        assert not compiled.fullmatch("\U0001f389\u2b50")
+        assert projection == {"_id": 1}
```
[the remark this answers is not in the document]
Zingzy: Addressed in 49f613f via a bytes-mode `re` assertion instead of a live Mongo fixture — bytes matching reproduces Mongo's bytewise PCRE behavior (it fails on the pre-fix ungrouped pattern, which str-mode matching lets through), and the repository unit suite intentionally runs without a real Mongo. The full-stack path was verified against a real MongoDB during development of the fix.

PR #252, opened 2026-07-13
fix(auth): serialize auth timestamps with explicit UTC offset

Zingzy:
## The bug

The app's PyMongo client is not `tz_aware`, so every datetime read back from Mongo is naive. The auth response DTOs serialized those values as-is, producing ISO strings with no UTC offset, for example `2026-07-13T09:00:00`. JavaScript's `Date` parser treats offsetless ISO strings as local time, so any dashboard viewer outside UTC saw these timestamps silently shifted by their own UTC offset, up to 14 hours.

On top of that, `POST /auth/onboarding/complete` was inconsistent with itself: the first completion returned a freshly stamped aware datetime (`...Z`), while idempotent repeat calls echoed the naive Mongo read-back with no offset. The same field flipped wire format between calls.

## The fix

The UTC-stamping serializer pattern already existed, duplicated verbatim, in the custom-domain and app-grant response DTOs. This PR promotes it to a single shared `UtcDatetime` annotated type in `schemas/dto/base.py` (a `PlainSerializer` that stamps naive values as UTC and emits `isoformat()`, so the wire form is `...+00:00`) and applies it to the affected auth fields:

- `UserProfileResponse.onboarded_at`, returned by `POST /auth/login`, `POST /auth/register`, `GET /auth/me`, `POST /auth/device/token`
- `AuthProviderInfo.linked_at` (inside the same profile payloads)
- `OAuthProviderDetail.linked_at`, returned by `GET /oauth/providers`
- `OnboardingCompleteResponse.onboarded_at`, returned by `POST /auth/onboarding/complete`

The onboarding endpoint now emits an identical format on first and repeat calls, since both paths flow through the same serializer. No route code changed.

The two DTOs that already had the hand-rolled serializer (`custom_domain.py`, `app_grant.py`) were refactored onto the shared type. That is a pure cleanup: the output is byte-identical (`+00:00` form), and their existing tests, including the integration tests that pin the exact wire strings, pass unchanged.

## Blast radius

These fields are dashboard-internal wire (plus the device-auth profile payload). The Next frontend already tolerates both aware and naive forms, so no compat window is needed. Shipped public formats (unix-epoch fields, legacy display strings, JWT claims) are untouched, and the Mongo client's `tz_aware` setting is unchanged.

## Tests

- New DTO unit tests pin that each affected field serializes with an explicit UTC offset for naive input, keeps the instant for aware input, and passes `null` through.
- New integration tests pin that `POST /auth/onboarding/complete` emits `+00:00` on the first call and that a repeat call echoing a naive stored value produces the exact same wire format.
- A regression test pins that the refactored custom-domain DTO still emits the exact `+00:00` strings it did before.
- Full affected suite (auth, oauth, onboarding, device auth, custom domains, apps DTO and route tests): 361 passed. `ruff check` and `ruff format --check` clean.

### DIFF
--- a/schemas/dto/base.py
+++ b/schemas/dto/base.py
@@ -1,11 +1,36 @@
 """
-Shared base classes for request and response DTOs.
+Shared base classes and field types for request and response DTOs.
 
 All DTOs inherit from these to get consistent Pydantic configuration
 without repeating ``model_config`` in every class.
+
+``UtcDatetime`` is the standard type for response timestamp fields:
+it serializes as ISO 8601 with an explicit UTC offset, stamping naive
+datetimes (PyMongo read-backs) as UTC on the way out.
 """
 
-from pydantic import BaseModel, ConfigDict
+from datetime import datetime, timezone
+from typing import Annotated
+
+from pydantic import BaseModel, ConfigDict, PlainSerializer
+
+
+def _stamp_utc(dt: datetime) -> str:
+    # PyMongo returns naive datetimes; without explicit tzinfo the JSON
+    # form omits the offset and clients parse it as local time. Stamp
+    # UTC so the wire format is unambiguous (`...+00:00`).
+    if dt.tzinfo is None:
+        dt = dt.replace(tzinfo=timezone.utc)
+    return dt.isoformat()
+
+
+UtcDatetime = Annotated[datetime, PlainSerializer(_stamp_utc, return_type=str)]
+"""Response datetime that always serializes with an explicit UTC offset.
+
+Use this for every new response timestamp field. Naive values (Mongo
+read-backs, the client is not ``tz_aware``) are stamped as UTC; aware
+values keep their instant. Wire form is ``2025-01-01T00:00:00+00:00``.
+"""
 
 
 class RequestBase(BaseModel):
--- a/schemas/dto/responses/app_grant.py
+++ b/schemas/dto/responses/app_grant.py
@@ -12,11 +12,9 @@ the registry's consent ``permissions`` strings, never scope slugs
 
 from __future__ import annotations
 
-from datetime import datetime, timezone
+from pydantic import Field
 
-from pydantic import Field, field_serializer
-
-from schemas.dto.base import ResponseBase
+from schemas.dto.base import ResponseBase, UtcDatetime
 from schemas.models.app import AppEntry
 from schemas.models.app_grant import AppGrantDoc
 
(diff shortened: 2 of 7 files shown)


Zingzy (commented):
## Review: serialize auth timestamps with explicit UTC offset

**Mergeable, no blockers.** Real bugfix (naive Mongo datetimes serialized offsetless → JS parses as local time → dashboard timestamps silently shifted) plus a clean DRY consolidation of a serializer that was already duplicated twice. Full matrix green, 0 behind main. I verified the two things that could bite.

**The shared `UtcDatetime` is correct — verified, not assumed:**
- **None handling.** `_stamp_utc` drops the explicit `None` guard the old hand-rolled serializers had — which is *correct here* because it's a type-level `PlainSerializer` on the `datetime` arm of `UtcDatetime | None`, so pydantic routes `None` through the union's NoneType arm and never calls `_stamp_utc(None)`. Pinned by `test_onboarded_at_none_stays_null` / `test_provider_linked_at_none_stays_null`. (The old `@field_serializer` sat on the whole `datetime | None` field, so it *did* receive `None` and needed the guard — the move to a type-level serializer is what makes dropping it safe.)
- **Byte-identical refactor.** For a datetime value the logic is identical to the deleted `_ser_as_utc` (stamp-if-naive → `isoformat()` → `+00:00`), and the regression test pins custom-domain's exact strings. The `custom_domain.py` / `app_grant.py` cleanup is a true no-op on the wire.
- Aware inputs keep their instant (documented) — and since every source here is either a naive Mongo read-back or a UTC-stamped value, the output is always `+00:00` in practice.

**Completeness:** all four affected auth/oauth ISO-datetime fields are converted (`onboarded_at` ×2, `linked_at` ×2) — grepped the response DTOs, no plain `: datetime` field left offsetless. Unix-epoch fields, JWT claims, and legacy display strings are correctly untouched.

**Nice side effect:** `POST /auth/onboarding/complete` now emits the same wire format on first and repeat calls (both flow through one serializer), fixing the self-inconsistency where the field flipped format between calls.

**Placement / merge safety:** `UtcDatetime` lives in `schemas/dto/base.py` next to the DTO bases — the right shared home, and the standard to reach for on future response timestamp fields. Blast radius is dashboard-internal + the device-auth profile payload; the frontend tolerates both forms, so no compat window. No route code changed. Good to merge — and it's the kind of consolidation that pays down every future timestamp field.

PR #253, opened 2026-07-13
feat(api): complete URL addressing with create-response id and single-resource GETs

Zingzy:
## Summary

The URL resource could be created and managed, but not addressed end to end. `POST /api/v1/shorten` returned `alias`, `short_url`, and friends, yet not the `id` that every management endpoint (`PATCH/DELETE /urls/{url_id}`) addresses by. A consumer had to call `GET /urls` and search for the alias just to manage a link it created one request earlier. There was also no way to read a single URL at all: the only read surface was the paginated list.

This PR closes both holes:

1. `id` on the create response
2. `GET /api/v1/urls/{url_id}`, fetch one owned URL by id
3. `GET /api/v1/urls/{domain}/{alias}`, fetch one owned URL by its natural key

## The create response now carries `id`

```
POST /api/v1/shorten
{"long_url": "https://example.com/launch"}

201
{
  "id": "665f1f77bcf86cd799439011",
  "alias": "mylink",
  "short_url": "https://spoo.me/mylink",
  "long_url": "https://example.com/launch",
  ...
}
```

Same serialization the list item already uses (ObjectId hex). Purely additive on a shipped wire: existing consumers see one new field, nothing moves or changes shape. The legacy `/emoji` endpoint has its own frozen response format and does not share this DTO, so it is untouched.

## Two addressing modes for the single-resource GET

Both return the exact same item shape one element of `GET /urls` has, byte for byte, so clients reuse their list-item parsing.

By id:

```
GET /api/v1/urls/665f1f77bcf86cd799439011

200
{
  "id": "665f1f77bcf86cd799439011",
  "alias": "mylink",
  "long_url": "https://example.com/launch",
  "status": "ACTIVE",
  "created_at": "2026-01-01T00:00:00Z",
  "password_set": false,
  "total_clicks": 42,
  "domain": "spoo.me",
  ...
}
```

By domain and alias, the name every short link is already known by:

```
GET /api/v1/urls/spoo.me/mylink          -> the default-domain link
GET /api/v1/urls/links.acme.com/promo    -> the same alias on a custom domain
GET /api/v1/urls/spoo.me/%F0%9F%90%8D%F0%9F%94%A5  -> emoji aliases arrive percent-encoded
```

The two routes cannot collide: one path segment is always the id form, two segments are always the composite key, so the router disambiguates purely on segment count. The existing `/urls` list, bulk delete, and `PATCH/DELETE /urls/{url_id}` routes are unaffected (different methods or different shapes).

Addressing a resource by its natural composite key next to its opaque id is the established shape for this kind of API: GitHub serves repos at `/repos/{owner}/{repo}`, Bitly addresses links by `{domain}/{hash}`. The domain segment is always explicit. It is normalized (lowercased, `:port` stripped), and the system hostname (including its `www.` form) resolves to the default-domain lookup, using the same source of truth the tenant middleware uses to answer "is this host the system domain". Any other value scopes the lookup to that custom domain.

## Ownership and 404 semantics

- Both GETs require the same scopes as the list endpoint: `urls:manage`, `urls:read`, or `admin:all`.
- Ownership is enforced inside the database query (`id`/`alias` AND `owner_id`), so a link owned by someone else answers exactly like a missing one: 404. These read endpoints never confirm what exists outside the caller's account, and unknown domains answer the same way.
- Reads are status-blind for the owner. Expired, disabled, and blocked links return normally with their `status` field, since the owner is reading inventory, not following a redirect.
- Malformed ids answer 400, reusing the management routes' id parsing.
- Rate limits match the list endpoint (60/min, 5,000/day authenticated).

## Merge safety

All changes are additive: one new response field, two new routes, two new repository finders, two new service methods. No existing route, DTO field, or query changes. Existing tests pass unmodified except for direct DTO constructions in unit tests, which now include the new field.

## Tests

- Create response carries `id` (anon and authed paths)
- By id: happy path serves the list-item shape, malformed 400, foreign 404, missing 404, auth and scope gates
- By domain and alias: happy on the system hostname (including case and port normalization), happy on a custom domain, emoji alias decoding, foreign 404, unknown-domain 404, disabled link returns with its status, auth and scope gates
- Service and repository units pin ownership-in-query (no unscoped read followed by a check) and the default-domain fallback

`pytest` on the affected files: 309 passed. `ruff check` and `ruff format --check` clean.

## Summary by CodeRabbit

- **New Features**
  - Added authenticated endpoints to retrieve shortened URLs by ID or domain and alias.
  - URL retrieval now returns consistent details, including status.
  - Shorten responses now include the URL’s ID for subsequent management actions.
  - Added support for normalized domains and percent-encoded aliases.

- **Bug Fixes**
  - Restricted URL lookups to the owning account, preventing disclosure of other users’ URLs.
  - Unknown or unauthorized URLs consistently return 404 responses.


routes/api_v1/urls.py
```diff
@@ -73,6 +88,130 @@ async def list_urls_v1(
...
+    **Errors**:
+
+    - `404` — no URL at that address in your account. Unknown domains and
+      links owned by someone else answer identically; this endpoint never
+      confirms what exists outside your account.
+
+    **Note**: only current-generation links resolve here — that is all the
+    managed collection holds.
+    """
+    short_code = unquote(alias)
```
Zingzy: **Should-fix (on rebase): canonicalize emoji aliases before the lookup.**

Since #251 merged, v2 emoji aliases are stored canonical (VS16-stripped) and every other surface canonicalizes before looking up: the redirect and public resolver via `v2_lookup_code`, `check_alias` via `canonicalize_emoji_alias`. This exact match is now the one read surface where a pasted VS16 variant (`⭐️🎉` for a stored `⭐🎉`) answers 404 while the redirect serves the link and check-alias calls it taken.

The seam already exists:

```python
from shared.alias_dispatch import v2_lookup_code

short_code = v2_lookup_code(unquote(alias))
```

No-op for alphanumeric codes (verified ~2µs), and it keeps this endpoint consistent with the rest of the v2 read story. Worth a test with a VS16-variant emoji alias alongside the existing percent-encoding one.

routes/api_v1/urls.py
```diff
@@ -1,28 +1,43 @@
...
     CustomDomainSvc,
+    Settings,
     UrlSvc,
     require_scopes,
 )
-from errors import ValidationError
+from errors import NotFoundError, ValidationError
 from middleware.openapi import AUTH_RESPONSES, ERROR_RESPONSES
 from middleware.rate_limiter import Limits, limiter
+from middleware.tenant import _normalise_host
```
Zingzy: **Should-fix: two private cross-module imports.**

`middleware.tenant._normalise_host` and `routes.api_v1.management._parse_url_id` are both privates reached into from another module, which is the usual sign the helper is sitting one level too low.

- `_normalise_host` is pure (no I/O), so it belongs in `shared/` — `shared/url_utils.py` next to `normalise_fqdn` is the natural home, since they are siblings (lenient host-header parsing vs strict fqdn validation). `middleware/tenant.py` then imports it from there.
- `_parse_url_id` now has callers in two route modules; promote it to a shared route helper instead of `urls.py` reaching into `management.py`.

Reusing them is right, the placement is the issue.

routes/api_v1/urls.py
```diff
@@ -1,28 +1,43 @@
...
+GET    /api/v1/urls/{domain}/{alias}   — fetch one owned URL by its natural key
+DELETE /api/v1/urls?domain=            — bulk-delete URLs scoped to one custom domain
 
 Requires authentication. Listing returns paginated list with camelCase keys.
+Single-resource GETs serve the exact same item shape the list serves per
+element, are status-blind for the owner, and answer 404 for foreign links
+(ownership is in the query — no existence oracle).
 Bulk delete is owner-scoped and refuses the system-default namespace.
+
+Route-ordering note: FastAPI matches routes in definition order. The two
```
Zingzy, line 13: Nit: this note is accurate for today but silent about the actual landmine. `GET /urls/{domain}/{alias}` occupies the entire two-segment GET namespace under `/urls`, and `management.router`/`stats.router` are included after `urls.router`, so any future `GET /urls/{url_id}/stats`-shaped route defined there gets silently shadowed into a 404 (the id parses as a domain, `stats` as an alias). Worth one sentence here stating the constraint: future sibling GETs under `/urls/x/y` must be defined in this module above this route, or take a different shape.

routes/api_v1/urls.py
```diff
@@ -73,6 +88,130 @@ async def list_urls_v1(
...
+    the system domain's names onto the canonical default-domain key,
+    mirroring the tenant resolver's system-default short-circuit (which
+    also accepts the ``www.`` alias). Anything else scopes the lookup to
+    that custom domain.
+    """
+    host = _normalise_host(raw)
+    if not host:
+        # Unparseable host — nothing can live there, same answer as unknown.
+        raise NotFoundError("URL not found")
+    if host in (system_default_domain, f"www.{system_default_domain}"):
```
Zingzy: Nit: this re-hardcodes the fold rule that `services/tenant_resolver/cached_mongo.py` owns (system domain + its `www.` alias short-circuit). Behaviorally identical today, but if the resolver's system-host set ever grows, this copy drifts silently and the API disagrees with the Host-header path about which hosts are the default domain. A tiny shared predicate (or a comment here pointing at the resolver as the source to keep in sync) would pin them together.

Zingzy (commented):
Mergeable, nothing blocking. Two should-fixes (one is merge-skew with #251 that lands the moment you rebase, one is placement) and a couple of nits, details inline. The security posture is the strongest part, so starting there.

## Security read

I went at the two new read endpoints adversarially and they hold:

- **Ownership lives in the query**, and the unit tests pin the negative (`find_by_id.assert_not_called()`), so the no-existence-oracle property is enforced by tests, not just intended. Foreign id, missing id, foreign alias, unknown domain: all indistinguishable 404s. Verified there is no code path that reads unscoped then checks.
- **Scope gating matches the sibling** (`URL_READ_SCOPES`, same as the list), rate limits and `responses=` match the list route exactly, and the 401/403 gates are tested for both endpoints.
- **No new indexes needed**: by-id rides `_id`, by-address rides the `(domain, alias)` unique index with the owner filter applied to at most one doc. Both O(1).
- **Route collisions**: verified against the full route table. The only existing two-segment path under `/urls` is `PATCH /urls/{url_id}/status`, method-filtered away. Today's claim in the module docstring is correct. Tomorrow's risk is inline.

## Should-fix on rebase: emoji aliases need canonicalization now

This branch forked before #251 merged, and #251 changed the ground under `GET /urls/{domain}/{alias}`: v2 emoji aliases now exist in `urlsV2`, stored canonical (VS16-stripped), and every other read surface canonicalizes lookups. This endpoint does an exact match, so it 404s a link the redirect happily serves. One line closes it, details inline at the `unquote` call.

## The id on the create response

Right call and overdue. Create-then-manage needed a list round-trip before, which is exactly the kind of thing API consumers write angry wrapper code around. Additive on a shipped wire, legacy `/emoji` untouched, serialization identical to the list item, and the test asserts round-trippability through `ObjectId()`. No notes.

## Merge safety

Additive only: no existing route, DTO field, or query changes. Ran the touched suites (290 tests) green in a clean worktree, CI matrix green. Merging ships nothing until the next release; nothing here needs a flag since the endpoints are strictly new surface behind existing auth.

## Product lens

The natural-key GET is the shape the frontend detail route needs, and the explicit-domain segment (rather than inferring from Host) keeps the API usable from scripts, the CLI, and the Raycast extension where there is no meaningful Host. Matching GitHub's composite-key convention means nobody has to learn anything. Status-blind owner reads are the right semantics for inventory and the docstrings state the reasoning where the next editor will find it.

tests/integration/api_v1/test_urls.py
```diff
@@ -58,3 +68,234 @@ def test_list_urls_api_key_missing_scope_returns_403(self):
...
+
+        application = _build_test_app(
+            {require_auth: lambda: user, get_url_service: lambda: mock_svc}
+        )
+        with TestClient(application, raise_server_exceptions=False) as client:
+            resp = client.get(f"/api/v1/urls/{ObjectId()}")
+
+        assert resp.status_code == 404
+
+    def test_get_url_missing_id_returns_404(self):
```
Zingzy: Nit: this is byte-identical to `test_get_url_foreign_id_returns_404` above it, both mock the same `NotFoundError` side effect. With a mocked service the foreign/missing distinction cannot be expressed at this layer anyway; the repo and service tests already pin it where it is real. Keep one.

routes/api_v1/urls.py
Zingzy: Fixed. The by-address lookup now canonicalizes with v2_lookup_code(unquote(alias)) before the exact match, so a pasted VS16 variant folds onto the canonical stored form (no-op for alphanumeric codes). Added a VS16-variant emoji test alongside the percent-encoding one.

routes/api_v1/urls.py
Zingzy: Done. Promoted _normalise_host to shared/url_utils.py as normalise_host (next to normalise_fqdn); tenant middleware and this endpoint both import it now. Moved parse_url_id into a new routes/api_v1/_helpers.py that both management.py and urls.py import, so no route module reaches into another for a private.

routes/api_v1/urls.py
Zingzy: Documented. Added a namespace note to the module docstring stating that any future sibling GET of the shape /urls/x/y must be declared in this module above this route or take a different shape, otherwise it is silently shadowed here.

routes/api_v1/urls.py
Zingzy: Extracted a shared predicate. is_system_default_host now lives in shared/url_utils.py and both this endpoint and the tenant resolver call it, so the system-domain plus www. alias rule has one source of truth.

tests/integration/api_v1/test_urls.py
Zingzy: Removed the duplicate. The two tests collapsed at the mocked-service layer, so I kept the foreign-id case (with its ownership-in-query docstring) and deleted the missing-id one. The real foreign vs missing distinction stays pinned in the repo and service unit tests.

Zingzy (commented):
Re-checked 8f4c1f5 (plus the main merge-in). All five findings verified fixed at the choke points:

- **Emoji canonicalization**: `v2_lookup_code(unquote(alias))` at the lookup, with a comment stating the constraint, and a new integration test driving a pasted VS16 variant and asserting the canonical form reaches the service. Exactly the seam intended.
- **Private imports**: `parse_url_id` promoted to `routes/api_v1/_helpers.py` with management.py migrated too, and `normalise_host` moved to `shared/url_utils.py` next to `normalise_fqdn` with the lenient-vs-strict sibling relationship documented. Both reach-arounds gone.
- **Fold-rule duplication**: resolved better than suggested — `is_system_default_host` is now a shared predicate and the tenant resolver itself was refactored onto it, so the rule literally cannot drift between the Host-header path and the API path.
- **Namespace note**: the two-segment GET shadowing constraint is now spelled out in the module docstring, including the concrete failure shape and where future sibling GETs must be declared.
- **Duplicate test**: dropped.

Ran the affected suites (279 tests) green in a clean worktree, CI matrix green at the new head. Nothing new introduced by the delta. Good to merge (deploy still after the frontend counterpart is ready, per the pairing).

PR #255, opened 2026-07-14
feat: bulk URL operations (delete, status, expiry)

Zingzy:
Backend counterpart to the dashboard's multi-select action bar. Today the frontend fans out one request per selected link, which trips per-item rate limits at trivially reachable selection sizes (select 70 links, hit delete, items 61-70 fail with 429) and gives no per-item failure report. This adds three bulk endpoints so one user intent is one request:

```
POST /api/v1/urls/bulk/delete   {ids: [...]}
POST /api/v1/urls/bulk/status   {ids: [...], status: "ACTIVE" | "INACTIVE"}
POST /api/v1/urls/bulk/expiry   {ids: [...], expire_after: datetime | null}
```

Move-domain is deliberately left for a follow-up PR: it is the one op with per-item interactions (alias collisions on the target namespace, first-wins ordering inside a batch) and deserves its own review.

## Contract

- Up to 100 ids per request, deduplicated (first occurrence wins). `urls:manage` or `admin:all` scope, no new scope.
- A processed batch always answers 200 with `{summary, results[]}`: exactly one row per unique id, in request order, each `ok: true` or `ok: false` with a machine `error_code` from the existing slug vocabulary (`not_found`, `forbidden`, `conflict`, plus `internal` and `not_attempted`). Per-item failures are answers, not errors.
- 4xx only for envelope rejection where zero items were attempted: malformed/empty/over-cap ids, past `expire_after`, missing scope, rate limit.
- Ownership is enforced inside the batch fetch query, never as a post-fetch compare. A side effect: someone else's id and a nonexistent id both report `not_found`, so bulk cannot be used as an existence oracle over foreign ObjectIds. This is a deliberate divergence from the single-item routes and is pinned by a test.
- Rate limits count per request, not per id (same posture as report intake): `60/min, 200/day` for status and expiry, `30/min, 100/day` for delete. Requests are what a management workflow spends (most real batches are small), so the request budget must never make bulk scarcer than looping the per-item routes; per-minute carries bursts like mass takedowns, the daily cap is a lid rather than a ration.
- Retrying an identical batch is safe by construction: every op sets an absolute value or deletes; already-deleted ids report `not_found`.

## Execution

Set-based rather than a per-item loop: one ownership-scoped fetch answers every verdict (missing, foreign, BLOCKED), then one `delete_many` or one `update_many` per slice mutates the actionable set. For a 100-link batch that is 2 Mongo round trips, 1-2 grouped Redis invalidations, and at most one detached Cloudflare KV call pair, instead of ~200 Mongo + 100 Redis + up to 100 CF calls.

Semantics mirror the single-item paths item for item, and a parity test suite drives the real single-item service and the bulk twin with identical mocks per scenario, asserting the same writes, cache keys, KV keys, and verdicts:

- status: same-status items are success no-ops with no write and no `updated_at` bump; EXPIRED set to ACTIVE reactivates.
- expiry: past values reject the envelope (single value for the whole batch); null clears; EXPIRED links whose expiry is extended or cleared reactivate via the same rules as `_auto_reactivate`.
- delete: BLOCKED refused per item, alias freed, historical clicks untouched (same as single-item).
- If the one set-write throws mid-batch, delete recovers by re-querying survivors so rows report the exact truth (gone means ok, survivor means `internal`); update slices report `internal` and still get their cache keys invalidated, since a spurious miss re-reads truth while a stale entry serves lies.

`BulkBatch.report()` raises if any requested id is missing a verdict, so a silently skipped item is structurally impossible.

## Side effects per successful item

- Redis `url_cache` invalidation, grouped into one variadic DEL per domain.
- Edge KV: one fire-and-forget bulk call per batch (never awaited in-request). Deactivate and delete purge `cache:{domain}:{alias}` for system-domain items, which drops og entries and any hot-promoted redirect entry in the same key. Purge-only, so plain links' promoted entries are never overwritten. This makes bulk deactivate/delete an actual takedown lever: the edge stops serving within seconds (the flush is detached) instead of waiting out the promotion TTL. Activate re-puts `og_only` entries for meta-tagged links, rendered through the same `build_og_entry` the write-through uses (extracted from `og_writethrough.sync` so the two paths cannot drift).
- `CloudflareKVClient` gains `bulk_put`/`bulk_delete` (10k-item chunks, same never-raises posture). wrangler dev's Explorer API only mirrors the single-key routes, so a 404 on the bulk path (which real CF never returns) degrades to per-key calls; the zero-cloud local loop exercises edge writes end to end.
- Logging: the same per-item `url_deleted`/`url_updated` events a loop would have emitted, plus one `urls_bulk_op` summary per batch with an error-code histogram (high-`not_found` batches are the enumeration signal in Axiom).

## Review follow-ups (pushed after review)

- CF bulk responses now check `result.unsuccessful_keys` (a 200 with partially-failed keys reports `False` and logs `cf_kv_bulk_partial_failure`), `expiration_ttl=0` is preserved on the wire, the by-ids delete failure gets its own log event key, and `error_code` is a typed `Literal` so the OpenAPI schema exposes the closed set.
- Rate limits raised to workflow scale (see above) after realizing the original numbers priced small-batch bulk gestures worse than the single-item loop they replace.
- The takedown purge is ported to the single-item paths: delete, deactivate, and alias/domain changes on plain links now drop the hot-promoted edge entry too, closing the path asymmetry and the alias-reclaim window where a freed alias kept serving the previous owner's destination until TTL. og-links stay event-managed by the write-through; the purge is plain-link-only so the mechanisms never double-fire. This makes the PR no longer purely additive: single-item mutations gain one detached best-effort KV delete for system-domain plain links. The delete parity test now pins both paths dropping the same key.

## Testing

- 2545 tests green (78 new): unit coverage for the service (mixed batches, no-op partitions, reactivation slices, write-failure recovery, report invariant), the parity suite, repository primitives with both-filters-defensive guards, KV bulk endpoints incl. chunking and the 404 fallback, DTO validation, route integration (auth, scope, envelope 422s, AppError passthrough), and smoke route/OpenAPI registration.
- Live full-loop smoke on a local stack (uvicorn + throwaway Mongo/Redis + `wrangler dev` with the Explorer API as the KV backend): drove every endpoint through mixed batches (owned, blocked, foreign-owned, nonexistent, duplicates), verified Redis keys invalidated, og and promoted KV entries purged on deactivate/delete and re-put on activate (confirmed through the worker: edge hit before, origin 410 after takedown; preview bot gets the rendered card after reactivation), EXPIRED links reactivating through both status and expiry, alias reclaim after delete, retry semantics, the 100-id boundary, scope enforcement for API keys, and the 429 at the sixth delete in a window.

## Summary by CodeRabbit

* **New Features**
  * Added bulk URL management API endpoints (delete, status change, expiry set/clear) supporting up to 100 URLs per request.
  * Bulk responses now include a per-URL results list plus an overall summary, returning 4xx only for request-level errors.
  * Enforced URL management scopes and dedicated per-endpoint rate limits; improved edge-cache consistency after bulk updates.
* **Tests**
  * Added integration and unit coverage for bulk request/response validation, authorization, per-item verdicts, and edge-cache/KV interaction behavior.


services/bulk_url_service.py
```diff
@@ -0,0 +1,469 @@
...
+                await self._kv.bulk_delete(purge_keys)
+            log.info(
+                "urls_bulk_edge_flush",
+                purged=len(purge_keys),
+                put=len(put_entries),
+            )
+        except Exception:
+            log.exception("urls_bulk_edge_flush_failed")
+
+    def _system_domain_keys(self, docs: list[UrlV2Doc]) -> list[str]:
```
Zingzy, line 214: **Should-decide: this purge makes bulk the better takedown lever than the single-item routes — port it back.**

Bulk deactivate/delete now drops hot-promoted redirect entries at the edge within seconds. The single-item `delete` and status paths still only touch KV for meta-tagged links, so the same action through `DELETE /urls/{url_id}` leaves a promoted entry serving until the promotion TTL expires. Two concrete consequences:

1. Same user intent, different takedown latency depending on whether they used the checkbox bar or the row action.
2. Single-item delete frees the alias while the edge may keep serving the old destination — if the alias is re-claimed, canonical-path requests can get the previous owner's redirect until TTL.

Both are pre-existing gaps, not regressions from this PR, and this PR builds everything needed to close them (`build_og_entry`, the shared KV client on the wiring, the purge-only posture). A small follow-up adding the same `cache_key` purge to `UrlService.delete` and the status handler would make the takedown story uniform. Worth a line in the follow-up list next to move-domain.

tests/unit/services/test_bulk_url_service.py
```diff
@@ -0,0 +1,718 @@
...
+        # Same doc removed, same Redis key cleared.
+        single_repo.delete.assert_awaited_once_with(_oid(1))
+        bulk_repo.delete_by_ids_and_owner.assert_awaited_once_with([_oid(1)], USER_OID)
+        single_cache.invalidate.assert_awaited_once_with("promo", SYSTEM_DEFAULT_DOMAIN)
+        bulk_cache.invalidate_many.assert_awaited_once_with(
+            ["promo"], SYSTEM_DEFAULT_DOMAIN
+        )
+        assert report.results[0].ok is True
+        # Plain link: single-item touches no KV (og.remove is og-only);
+        # bulk's purge is the DOCUMENTED superset (PRD §7.3 — takedown
```
Zingzy: Nit: `PRD §7.3` (and `PRD §8` at the foreign-id divergence test below) reference a planning doc that is not in the repo — the next reader cannot follow the pointer, and it dates the comment. The justification is already stated inline in both places ("takedown also drops a hot-promoted entry; delete-only, so plain-link untouchability holds"), so the citations can just go; the behavior contract lives in `bulk_url_service.py`'s docstrings, which is where the divergence is already documented.

services/bulk_url_service.py
```diff
@@ -0,0 +1,469 @@
...
+        for domain, aliases in by_domain.items():
+            await self._url_cache.invalidate_many(aliases, domain)
+
+    def _edge_flush(
+        self,
+        *,
+        purge_keys: list[str] | None = None,
+        put_entries: list[tuple[str, str]] | None = None,
+    ) -> None:
+        """At most one detached CF KV call-pair per batch.
```
Zingzy: Nit: "at most one detached CF KV call-pair per batch" is not quite true on the expiry failure path — a failed plain-slice `_apply_update` fires a purge flush and the surviving reactivate slice fires a put flush, so two tasks can detach for one batch. The property that actually matters (never awaited in-request, grouped rather than per-item) holds everywhere; suggest wording it as that instead of a count.

Zingzy (commented):
Mergeable, nothing blocking. This is the strongest-engineered PR of the recent batch: the parity suite driving the real single-item service against the bulk twin is exactly the anti-drift mechanism a set-based rewrite needs, and `BulkBatch.report()` raising on missing verdicts makes silent skips structurally impossible rather than review-enforced. One should-decide about the asymmetry this creates with the single-item path, two nits. What I verified before saying any of that:

## Verification

- **Parity claims checked against main line by line**, not taken from the PR body: `_auto_reactivate`'s four arms (bulk correctly reproduces the expire-extended/cleared arms, including the quirk where a max-clicks-expired link reactivates on an expiry change and re-expires on next hit — both paths share it, which is what parity means), `_handle_status`'s same-status no-op with no `updated_at` bump, `_handle_expire_after`'s truthiness clear-guard and past-value message, the exact `ForbiddenError` strings, and `og_writethrough.sync`'s put/delete conditions. All mirror.
- **The scariest candidate was a false alarm**: the service calls `url_cache.invalidate_many`, which is not in this diff and which every test mocks — a missing method would have sailed through the suite and blown up on the first real batch. It exists on main (added with the domain-scoped bulk delete), single-domain per call, and the service groups by domain before calling it. Correct.
- **Rate-limit arithmetic in the `Limits` comment checks out**: 100/day x 100 ids = 10k items/day = 5x the 2000/day per-item PATCH budget; 50/day x 100 = 5k = 5x the 1000/day per-item delete budget. The "same posture as report intake" phrasing refers to per-request counting, which is accurate.
- **No existence oracle**: ownership in the batch fetch, foreign and missing ids indistinguishable, pinned by a dedicated divergence test so it stays deliberate. The error-code histogram on `urls_bulk_op` giving Axiom an enumeration signal is good defensive product thinking.
- **Ran the touched suites (166 tests) green in a clean worktree**; CI matrix green.

## The one design conversation: takedown latency now differs by path

Bulk deactivate/delete purges `cache:{domain}:{alias}` for every system-domain item, so hot-promoted redirect entries stop serving within seconds. The single-item paths still only touch KV for meta-tagged links (`og_writethrough.remove`/`sync`), so deleting or deactivating one link through the existing routes leaves a promoted entry serving until the promotion TTL runs out. Inline with the details — it is a pre-existing gap this PR merely makes visible, and this PR builds everything needed to close it.

## Merge safety

Additive only: three new POST routes (no method or path overlap with anything registered, including the in-flight #253 GETs), one new service wired per the house three-file pattern, two new repo primitives with defensive both-filters guards, KV bulk methods that degrade to per-key on wrangler dev. Edge flush is best-effort and never awaited in-request, with the TTL as backstop. Merging ships nothing until the next release; the endpoints are new surface behind existing auth and scopes.

One wording note on the PR body rather than the code: "the edge stops serving within the request" overstates it slightly — the flush is deliberately detached, so it is within seconds, not within the request. The code comment has it right.

## Notably good

The delete write-failure recovery re-querying survivors so every row reports exact truth (gone means ok with side effects, survivor means internal) is the kind of failure-path honesty most bulk APIs skip. `build_og_entry` extracted from `sync` so the two rendering paths cannot drift. The wrangler-dev 404 fallback keeping the zero-cloud loop exercising real edge writes. And per-item `url_deleted`/`url_updated` events preserved so existing Axiom dashboards cannot tell bulk from a loop.

Zingzy (commented):
Re-checked through 1d4260f. All three findings resolved, and the delta earns a real review of its own since it adds behavior — that new behavior holds up too.

**The findings:**
- **Takedown parity (should-decide)**: implemented in-PR rather than deferred, and more completely than suggested — `UrlService` now purges the edge key on delete, deactivate, rename, AND move for plain links, with og-links staying event-managed by the write-through so the two mechanisms cannot double-fire. I traced the gating adversarially: the `is_og` split covers meta-tags-being-removed correctly, rename purges the pre-change key (right, since a link is never promoted under a key it has not served on), tenant links no-op, and `og_writethrough`/`edge_kv` are wired both-or-neither from the same `edge.enabled` block so the branch arms cannot disagree about whether the edge exists. The parity test now asserts both paths drop the identical key. Eight new unit tests cover the matrix including the no-KV no-op.
- **PRD citations**: gone, rationale kept inline pointing at the module docstring where the divergence is documented.
- **`_edge_flush` docstring**: reworded to the property that actually holds (detached and grouped, one flush per slice on partial failure).

**The delta beyond the findings, judged on its own:**
- **Rate-limit raise** (status/expiry 60/min 200/day, delete 30/min 100/day): I redid the math. Delete's item ceiling goes from 5k to 10k/day (10x the per-item route's budget), status/expiry to 20k/day. The reasoning comment is honest about the trade (requests are what a workflow spends; the cap is a lid, not a ration), everything stays ownership-scoped so the radius is self-harm plus the enumeration angle the histogram already watches. Deliberate product call, clearly documented, no objection — just flagging the delete ceiling doubled so it is a conscious fact.
- **`unsuccessful_keys` handling**: good catch that CF's bulk endpoints can 200 with per-key failures — the bool contract now tells the truth, with the non-JSON local-emulator case handled and both paths tested.
- **`expiration_ttl=0` truthiness fix**, the `BulkErrorCode` typed literal (the unused `conflict`/`validation_error` slots are a documented forward-compat seam for move-domain, fine), and the `repo_delete_by_ids_failed` log-event split are all improvements I had not asked for.

Ran the affected suites (268 tests) green in a clean worktree; CI green at the new head. Good to merge.

PR #256, opened 2026-07-14
feat(api): expose accepted emoji set at GET /api/v1/emoji-set

Zingzy:
## What

Adds a public, unauthenticated endpoint, `GET /api/v1/emoji-set`, that serves the emoji-alias acceptance policy as data. It is a static, build-time derived constant (like a config catalogue), so any client that needs the accepted set (an emoji picker, a dice suggester) has one source of truth instead of replicating the acceptance rules on the client. Each accepted emoji carries its name so a picker can search by name rather than making a user scroll a thousand-plus glyphs.

## Why

The rules for which emoji a custom alias may use (single-codepoint, fully-qualified, version-capped, no flags/keycaps/ZWJ families/text-style symbols) live in `shared/emoji_policy.py`. Any picker that wants to offer only acceptable emoji would otherwise have to reimplement that policy and keep it in sync as the pinned `emoji` package or the caps change. This endpoint hands the picker the exact set the server accepts, already named for search.

## Response shape

```json
{
  "accept_max_version": 15.1,
  "generate_max_version": 12.0,
  "max_graphemes": 15,
  "emoji": [
    { "c": "⭐", "n": "star", "gen": true },
    { "c": "🚀", "n": "rocket", "gen": true },
    { "c": "🎉", "n": "party popper", "gen": true, "k": ["tada"] },
    { "c": "🆎", "n": "ab button (blood type)", "gen": true, "k": ["ab", "ab button blood type"] }
  ]
}
```

- `emoji`: every single-codepoint emoji a user may choose, at the acceptance cap. This is the picker's list. Skin-tone variants are not enumerated: the base emoji suffices for a picker and skin tone is a client-side modifier appended to a base. Multi-codepoint accepted forms (base + skin tone) are therefore excluded by construction.
  - `c`: the raw canonical emoji character (no `U+FE0F` variation selector), matching how aliases are stored and echoed.
  - `n`: a human-readable, lowercased, space-separated name (the primary search key), taken from the same pinned `emoji` package the set is derived from (the `en` field, colons stripped and underscores spaced). No second dataset.
  - `gen`: whether the emoji is in the server's auto-generation pool (the lower generation cap). A client suggester can filter `gen=true` to match server auto-gen.
  - `k`: optional extra search aliases, present only when the source package lists any (e.g. `tada` for the party popper), sourced from its `alias` list and cleaned the same way. Omitted otherwise. Name search is the floor; these only widen it.
- Caps and `max_graphemes` come from settings, not hardcoded, so self-hosters that override them get a consistent catalogue.
- Categories/groups are not included: the pinned `emoji` package does not expose a group or subgroup field, and adding a dependency or a hand-maintained category map to synthesize one was out of scope. A picker falls back to search plus recents, which is fine.
- Ordering is the derivation's natural, deterministic order (the pinned package's data order).

Current build: `emoji` has 1170 entries, 1058 of them flagged `gen=true`.

## Caching

Low-volume, interaction-gated read, so the browser HTTP cache plus a client memo carry it, no edge rule or server-side cache layer needed. Response sets:

- `Cache-Control: public, max-age=86400, stale-while-revalidate=604800` (a day fresh, a week of serve-stale-while-revalidate).
- A strong, content-derived `ETag` (a hash of the caps plus every entry's char, name, gen flag, and aliases). This is deliberately not `immutable`: the set changes on deploy when the pinned Unicode version bumps, and `immutable` would strand clients on a stale list until the TTL. Instead a new set yields a new ETag, so conditional revalidation returns fresh data with no stale-forever tradeoff.
- Conditional requests: `If-None-Match` matching the current ETag returns `304` with no body.

## Single derivation

Refactored `shared/emoji_policy.py` so the accepted set and the existing generation pool share one derivation. Added `accepted_singletons(max_version)` (pure, cached), which materializes the single-codepoint arm of the acceptance predicate by running the same per-grapheme check the validator uses. `generation_pool` is now a filtered view of it (the accepted set at the lower generation cap, minus regional indicators). This guarantees the exposed set can never drift from the validator that accepts aliases, and keeps the existing pool invariants intact. Two small pure helpers, `emoji_display_name` and `emoji_keywords`, derive the name and alias strings from the same package, so names stay on the single source of truth.

## Tests

- Unit (`tests/unit/shared/test_emoji_policy.py`): `accepted_singletons` is non-empty and cached, single-codepoint only, agrees with `check_emoji_alias` for the whole set, contains known-safe emoji, excludes text-default and multi-codepoint and flag and ZWJ forms, the accept cap is a superset of the generate cap, the generation pool is a subset of the accepted set, and an empty set raises. Name and keyword derivation: names are non-empty, colon-free, and lowercased for the whole accepted set, a known emoji maps to its expected name, aliases are cleaned and never duplicate the name, and emoji with no aliases yield an empty tuple.
- Integration (`tests/integration/api_v1/test_emoji.py`): 200 unauthenticated; response shape is `{accept_max_version, generate_max_version, max_graphemes, emoji}` with `emoji` a non-empty list of `{c, n, gen}` objects (`k` optional); names are non-empty and colon-free; caps and `max_graphemes` match settings; known-safe emoji are present with their expected names (including `rocket` resolving by name); `gen=true` entries all pass the validator; the set excludes a flag, a ZWJ family, a keycap, and a text-default form; the derivation agrees with the validator across the whole set; `Cache-Control` and a stable `ETag` are present; and a conditional `If-None-Match` request returns `304` with no body.

All affected tests pass (90). `ruff check` and `ruff format --check` are clean.

## Merge safety

Additive: a new public read endpoint plus pure, cached derivation functions. No existing behavior changes; the generation pool refactor preserves its output and invariants. Nothing is wired into the create path.

## Summary by CodeRabbit

* **New Features**
  * Added an unauthenticated `GET /api/v1/emoji-set` endpoint that returns a structured emoji catalog with version limits and search-friendly metadata, plus per-entry info on whether an emoji is in the auto-generation pool.
* **Bug Fixes**
  * Improved emoji catalog ordering and grouping using a canonical emoji grouping dataset, ensuring the returned set strictly follows the configured acceptance rules.
  * Enhanced HTTP caching with stable `ETag` values and correct `304` handling for strong, weak, and wildcard conditional requests.
* **Tests**
  * Expanded integration and unit tests to validate response schema, grouping/order invariants, pool boundaries, and conditional caching behavior.


routes/api_v1/emoji.py
```diff
@@ -0,0 +1,130 @@
...
+    **Caching**: Fresh for a day, then served stale for a week while
+    revalidating. A content-derived ``ETag`` is returned; send it back as
+    ``If-None-Match`` to get a ``304`` when the set is unchanged.
+
+    **Rate Limits**: 60/min, 2,000/day.
+    """
+    payload = _build_set(settings)
+    etag = _etag(payload)
+
+    if_none_match = request.headers.get("if-none-match", "")
```
Zingzy, line 153: **Should-fix: use weak comparison for `If-None-Match`, or the 304 path may never fire in prod.**

RFC 9110 §8.8.3.2 requires `If-None-Match` to be compared ignoring the `W/` prefix, and that matters concretely here: intermediaries that compress responses are allowed to weaken a strong ETag in transit, and prod serves through Cloudflare, which compresses JSON. The browser then caches `W/"abc"` and echoes it back, this exact-string comparison misses, and every revalidation ships the full 56KB 200 — silently defeating the 304 machinery this PR built. No wrong answers, just the feature not doing its job behind the real proxy chain.

One line closes it:

```python
tags = {tag.strip().removeprefix("W/") for tag in if_none_match.split(",")}
if etag in tags or "*" in tags:
```

(the `*` arm is the other RFC case — pedantic, but it is the same line.) Worth a test with a `W/`-prefixed `If-None-Match` asserting 304, since that is the form prod clients will actually send.

routes/api_v1/emoji.py
```diff
@@ -0,0 +1,130 @@
...
+
+router = APIRouter(tags=["URL Shortening"])
+
+# A day fresh, a week of serve-stale-while-revalidating in the background.
+# Not `immutable`: the set changes on deploy, and the ETag below makes that
+# revalidation return fresh data rather than stranding clients until TTL.
+_CACHE_CONTROL = "public, max-age=86400, stale-while-revalidate=604800"
+
+
+def _build_set(settings) -> EmojiSetResponse:
```
Zingzy: **Should-fix: memoize the built payload and ETag — this became material with the enrichment.**

Measured at this head: ~2.6ms to build (1170 `EmojiEntry` constructions plus per-entry keyword derivation) and ~0.6ms to hash, on every request — including conditional ones, so a 304 costs exactly as much CPU as a 200. The whole response is a constant per `(accept_cap, generate_cap, max_graphemes)`, and this is a public unauthenticated endpoint on a single shared box.

A small cached helper keyed on the three caps returning `(payload, etag)` makes every request after the first effectively free:

```python
@lru_cache(maxsize=4)
def _catalogue(accept_cap: float, generate_cap: float, max_graphemes: int):
    payload = ...  # current body of _build_set
    return payload, _etag(payload)
```

At the c0bad05 revision this was a take-it-or-leave-it nit (~190µs); the enrichment multiplied the cost by ~15x, which is what moves it to should-fix.

schemas/dto/responses/emoji.py
```diff
@@ -0,0 +1,70 @@
...
+"""
+
+from __future__ import annotations
+
+from pydantic import Field
+
+from schemas.dto.base import ResponseBase
+
+
+class EmojiEntry(ResponseBase):
```
Zingzy, line 19: Should-decide before merge, since this ships a public wire contract: `c`/`n`/`gen`/`k` vs spelled-out `char`/`name`/`generated`/`keywords`. The compactness argument is weaker than it looks — Cloudflare compresses JSON, and 1170 repetitions of the same field name cost almost nothing after brotli (the raw-bytes savings is real, ~15-20KB, but the on-wire savings post-compression is small). The cost side is permanent: every client reads `e.c` and `e.n` forever, and it diverges from the descriptive snake_case every other response DTO in the API uses. Compact keys are a defensible convention for emoji datasets (emoji-mart does the same), so this is a legitimate style call rather than a defect — but it is the kind of decision to make deliberately on a public contract, not inherit from a draft.

Zingzy (commented):
Mergeable, nothing blocking. Reviewed at cd884ff (the names/keywords enrichment included). Two should-fixes, both small, plus one wire-contract question worth deciding before merge since this ships a public shape. What I verified:

- **The `generation_pool` refactor is provably output-identical**: the old inline filter and the new `accepted_singletons`-then-filter path apply the same predicate arms, including the `.get("E", inf)` default. Single derivation shared with the validator is the right shape, and the subset invariant holds under self-host cap overrides too, since the boot guard pins `generate <= accept`. (Side note: I re-verified that guard fires on current main via env override — `AppSettings` only reads env, so an earlier kwargs-based check of mine was testing nothing.)
- **Rate limiting present and honest**: `API_CHECK_ANON` (60/min, 2000/day) matches the docstring; the flat anon tier is right for a response that is identical for everyone.
- **Public is correct** — the policy is already public in the repo, and the payload exposes nothing exploitable.
- **Measured rather than guessed**: at this head, building the payload costs ~2.6ms plus ~0.6ms to hash it, and the serialized body is ~56KB. That measurement is what sets the tier of the caching finding inline.
- 90 affected tests green in a clean worktree; CI green. The integration suite driving real settings with no mocks is right for a pure-derivation endpoint, and the 304 test implicitly pinning ETag stability across app instances is a nice property to have locked.

The ETag-over-immutable reasoning in the module docstring (a deploy that bumps the Unicode pin revalidates to fresh data instead of stranding clients) is exactly the decision documentation that makes review easy. The names/keywords enrichment reusing the pinned package's `en`/`alias` fields instead of introducing a second dataset is the same discipline as the derived pool. Good PR.

schemas/dto/responses/emoji.py
```diff
@@ -0,0 +1,77 @@
...
+
+    c: str = Field(
+        description="Raw canonical emoji character (no U+FE0F variation "
+        "selector), matching how aliases are stored and echoed."
+    )
+    n: str = Field(
+        description="Human-readable name, lowercased with spaces (e.g. "
+        '"rocket"). The primary search key.'
+    )
+    g: str = Field(
```
Zingzy, line 34: Fold this into the open wire-shape decision: `g` repeats the full display name per entry (~1170 repetitions of ten strings, ~20KB of the payload's raw growth). The artifact itself already uses the smaller shape — a `groups` array plus an index per entry — so if the compact-key direction wins the naming question, mirroring that here is the consistent move. If the spelled-out direction wins, per-entry names are fine (brotli flattens the repetition). Either way, decide both together before this ships as a public contract.

Zingzy (commented):
Reviewed the grouping delta (aadc919) on its own — it holds up. The three findings from the previous review are unchanged and still open (weak `If-None-Match` comparison, payload/ETag memoization, the compact-key naming decision); notes on how this delta affects two of them below.

## The committed artifact is the right exception to the derive-don't-commit rule

After #251 deleted `data/emojis.json` precisely because checked-in emoji artifacts drift, reintroducing one deserved scrutiny, and it survives it:

- The pinned `emoji` package genuinely has no group data, so the alternatives were a runtime dependency or nothing. Distilling `emoji-test.txt` once into a committed artifact with no runtime network is the sober choice.
- The artifact is presentation-only and the code says so at every layer — it never decides acceptance, and membership-preservation is pinned at both unit and integration level.
- **The drift story is the strong part**: `test_every_accepted_char_has_a_canonical_group` asserts full coverage of the accepted set against the artifact, so a future `emoji` pin bump that outpaces `emoji_groups.json` fails CI loudly and forces a regeneration — the same loud-failure invariant #251 established for the pool. I verified coverage empirically at this head: zero accepted chars missing. The runtime fallback (annotate as Symbols, sort last) then only exists for the window where someone overrides the test, which is the right belt-and-suspenders layering.
- The generator script is clean: version-pinned default matching the package pin, VS16-stripping that mirrors the policy's canonicalization (so text-default entries join correctly), pure `distill()` that is unit-testable, and a generated file formatted to stay reviewable.

The lazy, cached file read in `shared/emoji_policy.py` is a mild bend of the shared-is-pure rule, but it is the same shape the old `generators.py` artifact load had and the package's own data is file-backed anyway — not worth relocating.

## Effect on the open findings

- **Memoization should-fix — now stronger.** Measured at this head: ~3.7ms per request to build (sort + group join on top of the enrichment) and the payload is now ~79KB. That is per request including 304s, on a public unauthenticated endpoint. The `lru_cache`d `(payload, etag)` helper from the previous review remains the fix and now saves ~20x more than when first suggested.
- **Wire-shape should-decide — one more input.** `g` repeats full display names ("Smileys & Emotion") per entry, ~1170 times. If the compact-key direction wins, the consistent shape is `groups: [...]` once at the top level with `g` as an index (exactly how the artifact itself is stored); if the spelled-out direction wins, per-entry names are fine since brotli flattens the repetition either way. Worth settling both in the same pass.

Ran the affected suites (100 tests) green in a clean worktree; CI green at the new head.

schemas/dto/responses/emoji.py
Zingzy: Keeping the compact c/n/gen/k + per-entry g shape, as a deliberate call on this contract. It is a defensible convention for emoji datasets (emoji-mart uses the same shape), the client already consumes it, and brotli at the Cloudflare edge flattens the repeated-key and repeated-group-name cost, so the post-compression wire savings of the alternatives are small. Recorded as decided, not inherited.

schemas/dto/responses/emoji.py
Zingzy: Keeping the compact c/n/gen/k + per-entry g shape, as a deliberate call on this contract. It is a defensible convention for emoji datasets (emoji-mart uses the same shape), the client already consumes it, and brotli at the Cloudflare edge flattens the repeated-key and repeated-group-name cost, so the post-compression wire savings of the alternatives are small. Recorded as decided, not inherited.

routes/api_v1/emoji.py
Zingzy: Fixed: If-None-Match now uses RFC 9110 weak comparison, stripping the W/ prefix (the form Cloudflare emits on compressed responses) before matching and honoring the * wildcard. Added a test asserting a W/-prefixed conditional request returns 304.

routes/api_v1/emoji.py
Zingzy: Fixed: added an lru_cache(maxsize=4) helper keyed on the three policy caps (accept version, generate version, grapheme limit) that returns the (payload, ETag) pair, so every request after the first for a given triple skips the ~3ms build and hash, 304s included. Body and ETag value are unchanged.

Zingzy (commented):
Re-checked dbcbc81. Both should-fixes are resolved and verified:

- **Weak If-None-Match**: `removeprefix("W/")` plus the `*` wildcard, with the RFC citation and the Cloudflare rationale stated where the next reader needs them. Both new tests pin the forms real clients actually send — the `W/`-prefixed one is the case that would have silently disabled every 304 in prod.
- **Memoization**: `_build_set` is now cached per policy triple and returns `(payload, etag)` together, so the 304 path shares the memo. Measured at this head: the request-time cost dropped from ~3.7ms to ~92ns per call after the first — the endpoint is now effectively free regardless of catalogue size. 102 affected tests green in a clean worktree, CI green.

One tiny wart in the fix commit: `_build_set` annotates `accept_cap` and `generate_cap` as `int`, but the version caps are floats (15.1 flows through at runtime). Nothing enforces annotations here so behavior is fine, but the signature misleads the next reader — `float` is the honest type.

The third item — the `c`/`n`/`gen`/`k` wire naming and the per-entry group-name encoding — is unchanged in this delta. If keeping the compact shape is the decision, that is a perfectly defensible call (emoji-dataset convention, documented field descriptions); just say so and it is closed. It only stays flagged because it ships a permanent public contract and deserves one deliberate sentence rather than a default.

Zingzy (commented):
Re-checked 6a4b600 — the cap params now annotate as `float` (with `max_graphemes` correctly staying `int`); annotation-only change, nothing else moved.

Recording the wire-shape decision per discussion: **the compact `c`/`n`/`g`/`gen`/`k` shape with per-entry group names is kept deliberately.** That closes the last open item on this PR. All findings across the review are now resolved: weak If-None-Match comparison (verified with the `W/` form tests), payload+ETag memoization (measured ~3.7ms to ~92ns), the coverage-pinned grouping artifact, and this annotation. Good to merge.

PR #257, opened 2026-07-14
feat: bulk move domain endpoint

Zingzy:
Completes the bulk operation set from #255 with the fourth action-bar operation: moving links between domains.

```
POST /api/v1/urls/bulk/domain   {ids: [...], domain: fqdn | null}
```

One target for the whole batch (matching the dashboard's single-picker dialog); `null` moves back to the system default. Same report grammar, scope, cap, and envelope rules as the other three; rate limit `60/min, 200/day`.

## Why this one is per-item where the others are set-based

Move is the only bulk op whose items interact: two links with the same alias moving to one target cannot both win, and the `(domain, alias)` unique index means one `update_many` can fail mid-flight with no way to attribute which item collided. So the writes go per item, sequentially, in request order, which buys two things: intra-batch collisions are deterministic (first in request order wins, the rest report `conflict`), and `DuplicateKeyError` maps cleanly to a per-item `conflict` verdict when a third party claims the alias between the availability check and the write. The batch fetch, verdict partition, report, cache grouping, and edge flush are all the shared machinery from #255.

## Parity with the single-item domain handler

- Target ownership is the route-edge precondition (`assert_owned_and_active`, same as PATCH): an unusable target rejects the whole request since it could never be right for any item. Explicit system-default and `null` skip the check.
- Already on target is a success no-op with no write and no `updated_at` bump.
- A reserved alias cannot move onto the system default (`validation_error`, same message).
- Availability is the SAME check the single-item handler calls (`check_alias_available`), which matters: on the system default it also consults the legacy v1 and emoji namespaces, which the v2 unique index alone would never catch. The service borrows the method rather than re-deriving the rule.
- Per moved item: both cache keys invalidated (pre-move and post-move, grouped per domain), old system-domain edge keys purged (og entries re-home, plain links' promoted entries stop serving an alias that left), and og-links landing on the system default begin KV management with a freshly rendered entry.
- Same per-item events (`url_updated`, `url_domain_moved`, and the reserved/conflict info events) plus the `urls_bulk_op` summary.

Also fixed while in there: `_og_put_entries` judged edge-put eligibility on the pre-write doc, which was correct for status/expiry (domain never changes) but wrong for moves; it now evaluates the merged doc and requires ACTIVE, so moving an INACTIVE og-link onto the system default does not create a KV entry.

## Testing

- 2601 tests green (18 new): the mixed-batch matrix (ok, no-op, reserved, conflict, intra-batch first-wins, DuplicateKeyError race, write-failure with defensive dual-key invalidation, blocked, foreign, missing), edge effects both directions (off-moves purge, onto-moves re-put from merged state), a parity suite driving the real single-item handler against bulk-move-of-one (same `$set` fields, same messages, same dual-key invalidation, same no-op), DTO normalisation, and route integration (ownership check fired with the normalised fqdn, null skips it, unowned target rejects the envelope with zero rows).
- Live smoke on a local stack with a real custom domain (mock DCV): moved links to the tenant and back, verified Mongo state and that redirects follow the move on the right Host (tenant host 302s, old host 404s), reserved and conflict verdicts with exact single-item messages, idempotent retry, and the envelope rejections (unowned target, malformed fqdn, missing field).

## Summary by CodeRabbit

* **New Features**
  * Added a bulk endpoint to move up to 100 URL aliases between verified custom domains or the system default.
  * Added validation and normalization for target domains.
  * Added rate limiting for bulk domain moves.
  * Added per-item handling for conflicts, unavailable aliases, reserved aliases, and no-op moves.

* **Bug Fixes**
  * Improved duplicate-key conflict handling without unnecessary error logging.

* **Tests**
  * Added coverage for endpoint validation, permissions, domain moves, conflicts, caching, and edge cases.


services/bulk_url_service.py
```diff
@@ -419,6 +436,138 @@ async def bulk_set_expiry(
...
+                    "conflict",
+                    f"Alias '{doc.alias}' is already in use on {target}",
+                    alias=doc.alias,
+                )
+                continue
+            try:
+                await self._url_repo.update_by_ids_and_owner(
+                    [doc.id], owner_id, set_ops
+                )
+            except DuplicateKeyError:
```
Zingzy, line 507: Nit, observability: this expected race still logs at ERROR before landing here — `update_by_ids_and_owner` catches every `PyMongoError` (which `DuplicateKeyError` is) and emits `repo_update_many_failed` at error level before re-raising. So each benign third-party alias race produces a repo-layer error event in Axiom that this handler then correctly treats as an ordinary `conflict`. Anyone alerting on repo error events gets paged for behavior that is working as designed. Cheapest fix is in the repo: let `DuplicateKeyError` pass through without the error log (or downgrade it to info there), keeping ERROR for genuine write failures.

Zingzy (commented):
Mergeable, nothing blocking. One observability nit inline. This completes the #255 set cleanly, and the two design decisions that needed defending both hold up under adversarial reading.

## The per-item write decision is correct, and I checked the mechanism it leans on

Move is genuinely the one op where set-based breaks: items interact through the `(domain, alias)` unique index, and an `update_many` cannot attribute a mid-flight collision to an item. The sequential design gets determinism (the `claimed` set makes intra-batch first-wins a code-level guarantee rather than a DB-ordering accident) and clean attribution. The load-bearing assumption is that a unique-index violation on an update surfaces as `DuplicateKeyError` — it does: PyMongo maps write error code 11000 to `DuplicateKeyError` for updates as well as inserts, and the service's except-ordering (subclass before `PyMongoError`) is right. Worst case cost is ~400 sequential Mongo calls for a 100-item move onto the system default; inherent to the correctness need, bounded by the cap and the 60/min limit, fine on a write endpoint.

## Borrowing `check_alias_available` is the important call

This is the exact reason move deserved its own PR: on the system default the availability rule spans the legacy v1 and emoji namespaces, which the v2 unique index would never enforce. Re-deriving that in the bulk service would have forked the rule; injecting `UrlService` and calling the same method means a v2 link cannot move onto a code that a nine-year-old legacy link still serves. The constructor comment says why the dependency exists, which is what keeps it from looking like a layering accident.

## Parity, verified not trusted

I read main's `_handle_domain` side by side: no-op on same target with no write and no `updated_at` bump, the reserved-alias guard with the identical message and same `url_domain_move_alias_reserved` event, the same conflict message, dual-key cache invalidation matching `update()`'s old+new clearing, and `url_domain_moved` (which does exist on the single path — checked) emitted with the same fields. The parity suite drives the real single-item handler for the message assertions, so drift breaks tests rather than clients.

## The `_og_put_entries` fix is a real catch

Eligibility was judged on the pre-write doc, which was accidentally correct for status/expiry (domain never changes there) and wrong for moves — and it was missing the ACTIVE gate that `sync` enforces. Judging the merged doc with the ACTIVE requirement fixes the move case, closes the moving-an-INACTIVE-og-link-creates-a-KV-entry hole, and changes nothing for the #255 paths (their put slices are ACTIVE-by-construction; existing tests still pass, I ran them). Good instinct to fix it here where the bug becomes reachable rather than shipping it latent.

## Merge safety

Additive: one route, one DTO, one service method, one limit constant. Smoke registration updated. Envelope precondition (`assert_owned_and_active`) mirrors the single-item route split, explicit system-default and null skip it correctly, and the DTO requiring an explicit `null` (no default) means nobody moves links to the default domain by omission. 123 tests green in a clean worktree; CI green.

Zingzy (commented):
Re-checked 8a94398. The nit is fixed at the choke point: `DuplicateKeyError` now passes through `update_by_ids_and_owner` without the repo-layer error event (subclass arm correctly ordered before the `PyMongoError` arm), the comment states why, and the new test pins both halves — the exception still propagates AND `log.error` is not called, so a regression re-adding the noise fails CI. Affected suites (82 tests) green in a clean worktree, CI green. Good to merge.

PR #259, opened 2026-07-18
feat(analytics): add device and UTM click dimensions

Zingzy:
Adds four new dimensions to v2 click analytics: **device**, **utm_source**, **utm_medium**, and **utm_campaign**.

## What changed

**Capture (hot path):** the redirect route now reads `utm_source`, `utm_medium` and `utm_campaign` from the short link's query string and stamps them on the `ClickEvent`. That is the only hot-path addition: three query param reads. Sanitization (control char strip, 100 char cap, empty to null) is enforced structurally by a `ClickEvent` field validator, following the same pattern as the password hash strip, so every producer inherits the bound.

**Derivation (click handler / worker):** device classification runs next to the existing UA parse in `V2ClickHandler`, so both the inline sink and the stream consumer get it with no extra parsing cost. `classify_device()` buckets into `mobile` / `tablet` / `desktop` / `unknown` using the OS family plus the Android `Mobile` token. This is the same fallback chain Matomo's DeviceDetector uses when its model regexes don't decide, and it is the approach used across the industry (ua-parser-js `device.type`, PostHog's hand rolled classifier). The signals are the ones Chrome's UA reduction deliberately preserved, so the classification is stable going forward.

**Persistence:** four new nullable fields on `ClickDoc`, same pattern as `meta.domain`: old time-series buckets keep their shape, missing values read as unknown/untagged. Legacy v1/emoji analytics are untouched.

**Stats API:** all four are now valid `group_by` dimensions and filters (both the `filters` JSON and dedicated query params). Untagged clicks aggregate under `(none)` for the UTM dimensions, mirroring referrer's `Direct`. Filtering by `(none)` matches null/missing documents; since more than one null-sentinel filter can now appear in a single query, the previous bare `$or` handling was generalized to nest multiple groups under `$and` (a second bare `$or` key would have overwritten the first). Exports (CSV/XLSX/JSON/XML) pick the new dimensions up automatically since they iterate the metrics dict.

**Public stats:** `device` joins the fixed v2 dimension set on the public per-link payload (additive key, existing consumers are unaffected). The UTM dimensions are deliberately excluded from the public payload: campaign tagging is the link owner's marketing data.

## Known limits (accepted, industry-wide)

- iPad Safari has sent a byte-identical Mac UA since iPadOS 13, so those visits land in `desktop`. This is not distinguishable server side; GA4, Adobe and Matomo count it the same way.
- Windows tablets report a stock desktop UA and land in `desktop`.

## Tests

- `classify_device` unit tests over representative UAs, including post-reduction Android (`Android 10; K`), Android tablet (no `Mobile` token), and iPad desktop mode
- UTM sanitization tests on `ClickEvent` (control chars, truncation, whitespace, pre-existing stream payload compat)
- Handler tests asserting the new `ClickDoc` fields
- Query builder tests for the `(none)` sentinel, mixed values, and the multi-group `$and` nesting
- Route-level tests asserting UTM params ride the emitted event
- Full suite green: 2667 passed

## Summary by CodeRabbit

* **New Features**
  * Added device classification for click analytics and public statistics.
  * Added UTM source/medium/campaign capture through redirects into click tracking.
  * Stats can now group and filter by device and UTM parameters.
* **Documentation**
  * Updated stats API descriptions and public stats dimension rules to reflect device and UTM support.
* **Bug Fixes**
  * Improved handling of “none”/missing-value filters so queries combine correctly across multiple dimensions.
  * Ensured UTM values are sanitized and stored consistently (including safe defaults when absent).

### DIFF
--- a/services/stats_service.py
+++ b/services/stats_service.py
@@ -62,6 +62,16 @@ _TIMEZONE_ALIASES: dict[str, str] = {
 # this set instead.
 _KNOWN_TIMEZONES: frozenset[str] = frozenset(available_timezones())
 
+# Dimensions whose aggregation output maps null/missing to a sentinel value
+# ("Direct" referrers, "(none)" for untagged utm clicks). Filtering by the
+# sentinel must match those null/missing documents too.
+_NULL_SENTINEL_FILTERS: dict[StatsDimension, str] = {
+    StatsDimension.REFERRER: "Direct",
+    StatsDimension.UTM_SOURCE: "(none)",
+    StatsDimension.UTM_MEDIUM: "(none)",
+    StatsDimension.UTM_CAMPAIGN: "(none)",
+}
+
 
 class StatsService:
     """Analytics query service.
@@ -153,7 +163,10 @@ class StatsService:
         # Time range
         query["clicked_at"] = {"$gte": start_date, "$lte": end_date}
 
-        # Dimension filters
+        # Dimension filters. Null-sentinel dimensions build $or groups;
+        # multiple groups must nest under $and (a second bare "$or" key
+        # would overwrite the first).
+        or_groups: list[list[dict[str, Any]]] = []
         for dimension, values in filters.items():
             if not values:
                 continue
@@ -169,25 +182,25 @@ class StatsService:
                     )
                     continue
                 query["meta.short_code"] = {"$in": values}
-            elif dimension == StatsDimension.REFERRER:
-                # "Direct" means null/missing referrer
-                if "Direct" in values:
-                    non_direct = [v for v in values if v != "Direct"]
-                    if non_direct:
-                        query["$or"] = [
-                            {"referrer": {"$in": non_direct}},
-                            {"referrer": {"$in": [None, ""]}},
-                            {"referrer": {"$exists": False}},
-                        ]
-                    else:
-                        query["$or"] = [
-                            {"referrer": {"$in": [None, ""]}},
-                            {"referrer": {"$exists": False}},
-                        ]
+            elif dimension in _NULL_SENTINEL_FILTERS:
+                sentinel = _NULL_SENTINEL_FILTERS[dimension]
+                if sentinel in values:
+                    non_null = [v for v in values if v != sentinel]
+                    clauses: list[dict[str, Any]] = []
+                    if non_null:
+                        clauses.append({dimension: {"$in": non_null}})
+                    clauses.append({dimension: {"$in": [None, ""]}})
+                    clauses.append({dimension: {"$exists": False}})
+                    or_groups.append(clauses)
                 else:
-                    query["referrer"] = {"$in": values}
+                    query[dimension] = {"$in": values}
             else:
                 query[dimension] = {"$in": values}
+
+        if len(or_groups) == 1:
+            query["$or"] = or_groups[0]
+        elif or_groups:
+            query["$and"] = [{"$or": group} for group in or_groups]
 
         return query
 
--- a/services/stats_service.py
+++ b/services/stats_service.py
@@ -63,10 +63,14 @@ _TIMEZONE_ALIASES: dict[str, str] = {
 _KNOWN_TIMEZONES: frozenset[str] = frozenset(available_timezones())
 
 # Dimensions whose aggregation output maps null/missing to a sentinel value
-# ("Direct" referrers, "(none)" for untagged utm clicks). Filtering by the
-# sentinel must match those null/missing documents too.
+# ("Direct" referrers, "(none)" for untagged utm clicks, "unknown" for
+# clicks that predate device tracking). Filtering by the sentinel must
+# match those null/missing documents too. Note "unknown" is ALSO a real
+# stored value (the classifier's fallback) — the filter branch keeps the
+# sentinel in its stored-value $in for exactly that reason.
 _NULL_SENTINEL_FILTERS: dict[StatsDimension, str] = {
     StatsDimension.REFERRER: "Direct",
+    StatsDimension.DEVICE: "unknown",
     StatsDimension.UTM_SOURCE: "(none)",
     StatsDimension.UTM_MEDIUM: "(none)",
     StatsDimension.UTM_CAMPAIGN: "(none)",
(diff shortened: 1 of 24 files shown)


services/stats_service.py
```diff
@@ -62,6 +62,16 @@
 # this set instead.
 _KNOWN_TIMEZONES: frozenset[str] = frozenset(available_timezones())
 
+# Dimensions whose aggregation output maps null/missing to a sentinel value
+# ("Direct" referrers, "(none)" for untagged utm clicks). Filtering by the
+# sentinel must match those null/missing documents too.
+_NULL_SENTINEL_FILTERS: dict[StatsDimension, str] = {
```
Zingzy, line 71: **Should-fix: the device unknown bucket splits in two and neither half is filterable.**

The device strategy in the aggregation factory predates this PR and has no `default=`, so it falls back to `$ifNull: [$device, "Unknown"]` — capital U. The classifier emits lowercase `"unknown"`, and the API docs added here promise `mobile/tablet/desktop/unknown`. Three consequences on any link with pre-deploy clicks (all of them, on day one):

1. Group-by device returns TWO buckets outside a coherent contract: `"Unknown"` (all historical clicks, from the aggregation fallback — a value the docs never mention) and `"unknown"` (new unclassifiable clicks).
2. `filter device=unknown` matches only the stored lowercase value, so it excludes the entire historical bucket.
3. Filtering by `"Unknown"` matches nothing at all — it is an aggregation-time synthetic, never a stored value.

The ClickDoc comment says missing device "reads as unknown", which is currently only true modulo capitalization and only for grouping. Two one-liners make the whole story coherent and match what the utm dimensions already do:

- give the device strategy `default="unknown"` in the factory (its neighbors gained `default="(none)"` in this PR; device missed it because its entry pre-existed), and
- add `StatsDimension.DEVICE: "unknown"` to this dict so filtering by `unknown` also matches null/missing docs, exactly like `(none)` and `Direct`.

Worth a test pinning that group-by and filter agree for a missing-device doc — that is the invariant that was easy to miss here.

Zingzy (commented):
Mergeable with one should-fix: the device dimension's unknown bucket is incoherent for every link with pre-deploy history, which is every link on day one (inline). Everything else held up under checking:

## Verified

- **The hot-path claim is true**: three `query_params.get` calls, nothing else — no new parsing, no allocation-heavy work on the 400k/day path. Device classification rides the UA parse the handler already does, so both sinks get it free.
- **Sanitization is structural and tested**: the `ClickEvent` validator bounds visitor-controlled input (control chars, 100-char cap, empty-to-null) so every producer inherits it — the same shape as the password-hash strip, which is the right precedent to copy. Tests cover each arm.
- **Stream compat both directions**: old payloads decode (defaults, with a test that strips the fields from a serialized payload), and new payloads reaching a not-yet-redeployed consumer are safe since the event model tolerates unknown fields. No deploy-order constraint.
- **The `$or` overwrite fix is a real catch**: the old referrer-only code was correct by luck (only one sentinel dimension existed, so two bare `$or` keys could never collide); with three more sentinel dimensions the collision became reachable, and the `$and`-of-`$or`-groups generalization is the right shape, with tests for the multi-group case.
- **`classify_device` is the right engineering call**: the OS-family + Android `Mobile`-token chain is what the industry converged on, the chosen signals are exactly what Chrome's UA reduction preserves, the iPad-desktop limit is documented with its industry context instead of hidden, and the None-safety on `ua.os`/`ua.device` is handled.
- **Product boundaries are right**: utm dimensions excluded from the public payload with the rationale stated (owner's marketing data), device added to the public fixed set as an additive key, aggregation cardinality bounded (20/50 caps), legacy v1/emoji untouched.
- Ran the touched suites (228 tests) green in a clean worktree; CI green.

## One product note, not a finding

The redirect drops the visitor's query string (`RedirectResponse(destination)` serves the stored URL verbatim), so UTM tags on a short link land ONLY in spoo's analytics — they never reach the destination, meaning GA/Plausible on the target site will not see them. That is a coherent story (tag the short link, read the campaign in spoo) and arguably a feature, but it is worth saying explicitly in the docs/dashboard copy when the FE widgets ship, because marketers will otherwise expect pass-through.

services/stats_service.py
Zingzy: Fixed in e712ece. Both suggested changes, plus one wrinkle the sentinel map surfaced: unlike Direct and (none), unknown is also a real stored value (the classifier fallback), so the filter branch now keeps the full value list in its stored-value $in instead of stripping the sentinel. Filtering device=unknown matches stored docs and pre-tracking docs in one query.

Added the invariant test you suggested: it pins the aggregation $ifNull default, the classifier fallback, and the sentinel map entry to the same "unknown" string, so any of the three drifting fails the build. Also verified against the local docker stack with a hand-inserted legacy-shaped doc (no device field): group-by shows a single lowercase unknown bucket and filter device=unknown returns it.

Zingzy (commented):
Re-checked e712ece. The device-unknown finding is resolved at the choke point, and the filter refactor is actually a small correctness improvement beyond the fix.

- **Group-by**: `device` strategy now carries `default="unknown"`, so a null/missing device folds into the same lowercase `unknown` bucket the classifier emits — no more synthetic capital-`Unknown` split. The `test_pipeline_uses_unknown_as_null_fallback` parametrization gained the `device` case.
- **Filter**: `DEVICE: "unknown"` is in `_NULL_SENTINEL_FILTERS`, so filtering by `unknown` now matches null/missing docs too. The branch was unified to keep the sentinel in the stored-value `$in` — which is exactly right: for device, `unknown` is a real stored value (the classifier fallback) so it must stay in the `$in`; for referrer/utm the extra literal matches nothing, and if a visitor ever sent the literal, group-by already merges it with null, so the filter now agrees. That is more consistent than the pre-fix code, which stripped the sentinel from the stored `$in`.
- **The invariant is pinned**: `test_device_groupby_and_filter_agree_on_missing_docs` asserts the aggregation `$ifNull` default and the filter sentinel are both `unknown`, and `test_device_unknown_matches_stored_and_missing` covers the stored+missing match. That is the exact group-by/filter agreement the finding was about.

Ran the stats + aggregation suites (99 tests) green in a clean worktree; CI green. No new surface. Good to merge — the UTM-pass-through note from the first review still worth a line of dashboard copy when the FE widgets land, but that is product framing, not a code gate.

PR #260, opened 2026-07-18
feat(auth): PKCE and grant-scoped tokens for connected apps

Zingzy:
Hardens the connected-apps (device auth) flow with mandatory PKCE and moves authorization from "a grant is the full account" to grant-scoped tokens.

## What changed

**PKCE, S256 only**
- `GET /auth/device/login` now requires `code_challenge` + `code_challenge_method=S256`. Invalid or missing PKCE renders the device error page with a clear "update the app" message. Extra OAuth params some clients send (`client_id`, `response_type`, `scope`) are ignored as before.
- The challenge rides through both mint paths (auto-approve and the consent form, where it is re-validated server side) and is stored on the one-time code.
- `POST /auth/device/token` requires `code_verifier` (43-128 chars, RFC 7636 charset). The S256 hash is compared with `secrets.compare_digest`, and every PKCE failure returns the same generic invalid-code error as a bad code. Codes live 5 minutes, so there is no migration window: codes without a stored challenge are rejected outright.
- New `pkce_s256_challenge` helper in `infrastructure/crypto.py`, tested against the RFC 7636 Appendix B vector.

**Grant-scoped tokens**
- Registry entries (`config/apps.yaml`) declare `scopes`. The loader skips a live device app that declares none, so a misconfigured entry can never mint an unrestricted grant. `spoo-cli` declares `shorten:create, urls:read, urls:manage, stats:read, domains:read, domains:manage, keys:manage`; `spoo-raycast` stays `coming_soon` (the extension is not published yet) with `shorten:create, urls:read, urls:manage, stats:read` and its hosted redirect URIs staged, so going live later is a one-line status flip.
- Consent snapshots the registry scopes onto the grant. Access tokens carry `scp` + `app_id`; refresh tokens carry `app_id` only, so scope edits propagate at the next refresh. Exchange and refresh both re-check the grant and re-read its scopes inside `DeviceAuthService`.
- The credential scope check now covers app tokens next to API keys (same OR-semantics intersection, same 403). Sessions stay unrestricted. Interactive-only surfaces (`/api/v1/me/*`, `GET /api/v1/apps`, `PATCH /auth/me`, device revoke) reject scp-bearing tokens.
- The session refresh flow (`/auth/refresh`) rejects app-bound refresh tokens; without that, an app could trade its scoped refresh token for an unrestricted session token.

**keys:manage**
- New `keys:manage` scope, grantable only via the registry. API keys can never be created with it (422 at creation, enforced in the request validator), and `admin:all` does not imply it, so the "a key cannot mint keys" guard holds.
- `/api/v1/keys` routes switch from JWT-only to a new guard: interactive session OR a credential holding `keys:manage`. This keeps `spoo keys` working from the CLI after it re-authenticates. Key creation still requires a verified email.

**API surface**
- `GET /api/v1/apps` items gain `scopes` (effective slugs, empty list means a legacy unrestricted grant) and `permissions` becomes derived consent copy from a shared scope-description module that mirrors the frontend's wording. The consent screen renders from the same source.
- `POST /auth/device/revoke` additionally accepts a JSON body with `app_id` and/or `grant_id` (resolved to the app for the calling user). The form shape and the `X-Requested-With: fetch` requirement are unchanged.

## Compatibility

- Current CLI and Raycast builds fail at NEW logins only (missing PKCE, clear error page). Raycast already sends the challenge but not yet the verifier at exchange.
- Existing refresh tokens keep working; they rotate through the grant check exactly as before and pick up scopes as soon as the grant has them.
- Legacy grants (no scope snapshot) inherit the registry scopes of their app at mint time; if the app has left the registry they stay unrestricted until re-consent.
- Sessions and API keys are untouched.

## Wire contract for client updates

- Login: `GET /auth/device/login?app_id=...&state=...&redirect_uri=...&code_challenge=<43 chars base64url>&code_challenge_method=S256`
- Exchange: `POST /auth/device/token` with `{"code": "...", "code_verifier": "..."}`
- Revoke: `POST /auth/device/revoke` with JSON `{"app_id": "..."}` or `{"grant_id": "..."}` plus `X-Requested-With: fetch`

Follow-up PRs: spoo-cli (send PKCE pair, drop `/auth/me` for `whoami` in favor of the exchange response), spoo-raycast (send `code_verifier`), spoo-landing (render `scopes`, JSON revoke).

## Testing

- Full suite: 2711 passed.
- New coverage: RFC 7636 vectors, PKCE happy path and downgrade/tamper rejections, scope minting on exchange and refresh, registry loader fail-safe, scp guard behavior, keys:manage enforcement with real JWTs on live routes.
- Manual e2e against the docker compose stack: register, consent with challenge, exchange with verifier (wrong verifier 401), scoped calls (in-scope 201, out-of-scope 403, interactive-only 403), refresh re-mints scp, JSON grant_id revoke, refresh after revoke 401.

## Summary by CodeRabbit

* **New Features**
  * Added PKCE (S256) end-to-end to device authentication, including PKCE-bound code exchange and consent.
  * Introduced scoped app grants: device tokens carry effective scopes, and connected-app responses now show scope-derived consent details (with legacy grants handled for backward compatibility).
  * Expanded access control so app/device tokens can manage API keys when granted `keys:manage`, with clear restrictions.
* **Bug Fixes**
  * Tightened authorization around scoped vs interactive sessions and improved enforcement for invalid/revoked grant and credential-scope scenarios.
  * Strengthened validation and authorization rules for live app scope declarations and API-key scope creation rules.


Zingzy (commented):
## Review: mergeable, nothing blocking

CI is green across all five Python versions plus test-api and formatting. I read the full diff, traced the security paths, ran the new test files in isolation (89 passed), and ran an adversarial pass on the PKCE flow specifically. No blockers. The architecture is sound and the change lands where the existing seams already are.

### Architecture read

The load-bearing decisions are the right ones:

- **PKCE state lives on the one-time code, not a session.** The challenge rides through both mint paths and is stored on the `verification-tokens` doc next to the code, so the atomic `consume_by_hash` covers challenge and code together. `code_challenge` is a keyword-only param with no default on `create_device_auth_code`, so omitting it is a `TypeError`, not a silent unrestricted mint. Both mint callers gate on `is_valid_pkce_challenge` first, and exchange re-checks independently (`not token_doc.code_challenge or not compare_digest(...)`). Defense in depth, no downgrade path.
- **Grant is the source of truth for scopes, re-read at every mint.** Access tokens carry `scp` + `app_id`, refresh tokens carry `app_id` only, so a scope edit on the grant propagates at the next refresh instead of freezing into the 30-day credential. `effective_scopes_for` (snapshot -> registry -> None) is a pure module function shared by the service and `/api/v1/apps`, so the wire and the enforcement can't drift.
- **The scope check generalized without a fork.** `check_credential_scopes` covers API keys and `scp` tokens with the same OR-semantics intersection; sessions without `scp` stay unrestricted exactly as before. `scp` is never minted as `[]` (empty registry scopes become `None` at consent), so an empty scope list is only reachable via a tampered token, which fails closed. That's the correct default-deny posture.
- **`keys:manage` self-propagation guard holds.** Excluded from `ALLOWED_SCOPES` and deliberately not implied by `admin:all`, so an API key can never mint keys. `require_session_or_scopes` lets a session or a `keys:manage` app token through, mirroring the existing `require_scopes` factory pattern rather than special-casing.
- **Closed a pre-existing hole:** `/auth/refresh` now rejects app-bound refresh tokens. Without it, an app could trade its scoped refresh token for an unrestricted session token through the cookie flow. Good catch beyond the stated scope.

IDOR check: `find_by_id_for_user` scopes the query to `user_id`, and revoke targets `grant_repo.revoke(user.user_id, app_id)`, so a `grant_id` from another account can't drive a cross-user revoke. The consent template renders the challenge as hidden inputs, so the non-auto-approve path re-validates correctly (the e2e confirmed this end to end).

### Should-decide (conscious calls, not defects)

- **Legacy grants silently narrow from full-account to registry scopes at the next mint.** For an existing spoo-cli grant, the token goes from unrestricted to the seven declared scopes within one access-token lifetime after deploy, with no re-consent. Safe here because the declared scopes cover every endpoint the CLI calls. The standing rule this sets: **a client's registry scopes must stay a superset of what it actually calls**, or existing installs start returning 403 within ~15 minutes of a deploy rather than at re-consent time. Worth stating in the apps.yaml region as a maintenance invariant.
- **Consume-before-verify on the code.** A wrong verifier still burns the one-time code (the atomic consume happens before the user lookup path completes). This forces a legit app to restart the flow if someone submits a stolen code with a bad verifier. Standard for single-use auth codes and it requires code interception (which PKCE isn't the control for), so acceptable, flagging it as a known property.

### Nits

- `_extract_revoke_target`: an over-length `app_id` raises `ValidationError("app_id is required")`, which is misleading (it *was* supplied). The wording is carried over from the old inline check, but now that there's a dedicated extractor it's a one-word fix to "app_id too long".
- `_extract_revoke_target(request, user_id, grant_repo)` leaves `grant_repo` unannotated while the rest of the module type-annotates params. Trivial.
- `check_api_key_scope = check_credential_scopes` back-compat alias: this is one repo, so the call sites could just be renamed, but it's documented as a deliberate keep, no objection.

### Merge safety

Merging ships nothing (deploys fire on release publish), and this is the device-auth surface only, sessions and API keys are untouched. The one deploy-time behavior to expect: already-issued device access tokens minted before the deploy run unrestricted until they expire (<=15 min), then refresh re-mints them with `scp`. That's the accepted stateless-token propagation window, not a regression. The hardening tail before this matters in prod is just the three client PRs landing so live spoo-cli/raycast installs re-authenticate cleanly.

### Notably good

The test coverage is real, not decorative: RFC 7636 Appendix B vectors on the S256 helper, PKCE downgrade/tamper rejection, scope minting on both exchange and refresh, the registry loader fail-safe (a live app with no scopes is skipped, not defaulted), and live-route enforcement with real JWTs. The service-layer split (grant re-check and scope resolution moved out of the routes into `DeviceAuthService`) matches the layering doctrine and made the routes noticeably simpler.

routes/api_v1/keys.py
```diff
@@ -46,7 +48,7 @@
 async def create_api_key(
     request: Request,
     body: CreateApiKeyRequest,
-    user: JwtVerifiedUser,
+    user: KeysAccessVerifiedUser,
```
Zingzy: **Blocker: this now admits scoped app tokens, but key creation never intersects the requested scopes against the caller's own.** Below, `body.scopes` is passed verbatim to `ApiKeyService.create` (`keys.py:92`), which stores it unchanged. `require_keys_access` upstream only checks the caller HOLDS `keys:manage`; it does not bound what the caller may GRANT.

Concrete exploit with the only live app, `spoo-cli` (holds `keys:manage`, not `admin:all`, not `reports:create`):

`POST /api/v1/keys {"name":"x","scopes":["admin:all"]}` → a `spoo_…` key with `admin:all`.

Verified in the repo `.venv`: `admin:all ∈ ALLOWED_SCOPES` and satisfies `REPORTS_SCOPES`, so the key passes the reports route the app token itself cannot — privilege widening. Worse, `revoke_device_tokens` deletes only `TOKEN_TYPE_DEVICE_AUTH` rows, never the `api-keys` collection, so the minted key survives grant revocation: a revocable scoped credential laundered into a permanent unrestricted one. This defeats the property the PR exists to create.

Fix: when `_granted_scopes(user) is not None` (scoped credential — app token or key), require `set(body.scopes) <= granted` and reject `admin:all` unless the caller holds it; interactive sessions (`granted is None`) still mint anything. Pin it with a test: a `keys:manage` app token minting a key with a scope it lacks must 403.

dependencies/auth.py
```diff
@@ -184,14 +203,16 @@ async def require_verified_email(
...
-    Use on endpoints where API key auth must be explicitly prohibited —
-    e.g. key management routes (an API key must not be able to create,
-    list, or delete other API keys).
+    Rejects API keys AND scoped device-app tokens (``scp`` claim). Use on
+    account-security surfaces — profile, app management, device revoke —
+    where a delegated credential must never act.
     """
     if user.api_key_doc is not None:
         raise ForbiddenError("API keys cannot be used to manage API keys")
+    if user.scopes is not None:
```
Zingzy: **Should-fix: this rejects on `scopes is not None`, but the invariant you want is `app_id is None`.** An app token with `app_id` set and no `scp` claim (a legacy/scopeless grant, where `effective_scopes_for` returns None) has `scopes=None`, so it passes `require_jwt` and reaches every interactive-only surface — profile edit, device revoke, `/me/*` — as a full session. Not reachable via spoo-cli today (it's scoped, so its tokens always carry `scp`), which keeps this should-fix rather than blocker, but the guard trusts the wrong claim and the docstring's promise ("a delegated credential must never act") isn't actually enforced for the scopeless case. Reject when `user.app_id is not None` as well (same at line 232). Worth a test with an `app_id`-set, scp-absent token asserting 403 on `PATCH /auth/me`.

Zingzy (commented):
One blocker: a `keys:manage` app token can mint an API key broader than itself that survives grant revocation — which defeats the exact property this PR exists to establish. The PKCE half is excellent and I could not break it; the scope half is right everywhere except this one laundering path. Details below, blocker first.

I reviewed this adversarially: three parallel deep-dives (baseline map, PKCE-refute, scope-refute) plus my own trace, and I verified every load-bearing claim against the code and the repo `.venv` rather than trusting the description. 226 auth tests green in a clean worktree.

## Blocker — `keys:manage` launders a scoped, revocable token into a broader, permanent API key

Key creation never intersects the requested key scopes against the *caller's own* scopes. `routes/api_v1/keys.py:92` hands `body.scopes` straight to `ApiKeyService.create`, which writes them verbatim. The new `require_keys_access` only checks the caller *holds* `keys:manage` — not what it may *grant*. So the one live app, `spoo-cli` (grant: `shorten:create, urls:read, urls:manage, stats:read, domains:read, domains:manage, keys:manage`, and deliberately NOT `admin:all` or `reports:create`), can:

`POST /api/v1/keys {"name":"x","scopes":["admin:all"]}` → a real `spoo_…` key with `admin:all`.

I verified both axes in the `.venv`, not just the shape:
- **Privilege widening**: the token lacks `reports:create`, but `admin:all` is in `ALLOWED_SCOPES` and satisfies `REPORTS_SCOPES`, so the minted key passes the reports route the token itself cannot. Strictly more authority than the credential that created it.
- **Revocation escape (the worse axis)**: `revoke_device_tokens` deletes only `TOKEN_TYPE_DEVICE_AUTH` rows — it never touches the `api-keys` collection. Revoking the spoo-cli grant leaves the minted `admin:all` key fully live. A short-lived, scoped, revocable delegated credential has been converted into a permanent, unrestricted account credential. That is precisely the failure the PR's own `api_key.py` comment names ("a key that can mint keys defeats revocation") — the self-propagation axis (key→key) is closed; the delegation axis (app-token→key) is wide open.

**Why blocker, with honest timing**: not exploitable the instant it merges — no spoo-cli grants exist yet (it was `coming_soon`), and the current CLI can't complete a login without PKCE (the follow-up PR adds that). But this PR is what flips spoo-cli to `live` AND grants it `keys:manage`, so the hole goes live the moment the CLI follow-up ships and one user logs in — and that reviewer will be reading CLI code, not this key-creation path. It defeats the PR's central security property and the fix lives in the surface this PR newly opened, so it belongs here, not in a follow-up.

**Fix**: when the creator is a scoped credential (`_granted_scopes(user) is not None`), require `body.scopes` to be a subset of the caller's granted scopes (and reject `admin:all` unless the caller holds it). Interactive sessions (`granted is None`) keep minting anything. A test pinning "a keys:manage app token cannot mint a key with a scope it lacks" locks it.

## Should-fix — interactive-only guards key off `scp`, not `app_id` (fail-open by construction)

`require_jwt` (`dependencies/auth.py:214`) rejects a delegated credential only when `user.scopes is not None`. An app token carrying `app_id` but no `scp` claim has `scopes=None`, so it passes `require_jwt` and reaches every interactive-only surface (`/api/v1/me/*`, `PATCH /auth/me`, device revoke, `GET /apps`) as if it were a full session. Such a token is minted whenever `effective_scopes_for` returns None — a legacy grant (`scopes=None`) whose app has no registry scopes. Not reachable via spoo-cli today (it's scoped, so its tokens always carry `scp`), which is why this is should-fix not blocker — but the guard is trusting the wrong claim. The account-security surfaces should reject any `app_id`-bearing token regardless of `scp` presence; that also makes the intent in the new `require_jwt` docstring literally true. One-line-per-guard fix, worth a test with an `app_id`-set/scp-absent token.

## What I tried to break and could not

- **PKCE (all five classes)**: RFC 7636 Appendix B vector matches exactly; S256-only enforced at both mint paths with server-side re-validation on the attacker-writable consent form field; `secrets.compare_digest`; urlsafe-unpadded encoding with the DTO charset/length bounds exact (43–128, no off-by-one); challenge bound to the specific code; missing-challenge codes rejected outright; PKCE failure returns the same generic error as a bad code (no oracle); and the code is atomically consumed *before* the PKCE check, so a wrong verifier burns the code — one guess per code, no replay, no brute-force. Genuinely solid.
- **scp→session upgrade**: `/auth/refresh` (session, cookie) now rejects any `app_id`-bearing refresh token (`credentials.py`), so a scoped refresh token can't be traded for an unrestricted session. Confirmed the rejection fires.
- **Fail-closed on empty/malformed scp**: `scp: []` and non-list `scp` both resolve to "grants nothing" (403), never unrestricted. Only *absent* scp reads as session — which is the Finding 2 surface.
- **Scope propagation**: consent snapshots registry scopes; refresh re-reads (snapshot wins, else registry, else None), bounded by the registry entry — no widening beyond consent for a scoped app.

## Notably good

The registry loader skipping (not defaulting) a live app that declares no scopes is the right fail-safe — a misconfigured entry can never mint an unrestricted grant. `admin:all` deliberately not implying `keys:manage`, and `keys:manage` excluded from `ALLOWED_SCOPES`, correctly close the key→key axis (it's only the app-token→key axis that leaks). Access token carries `scp`, refresh carries only `app_id` so edits propagate at refresh rather than freezing into the long-lived credential — the right split. And the single shared scope-description module so consent copy can't drift from the frontend.

## Merge safety

Against main this is a real tightening — today every app token is an unrestricted JWT that bypasses all scope checks, so scoping them is strictly better. The blocker is not a regression from main (main's spoo-cli was dark); it's a hole in the *new* model that becomes live when spoo-cli + the CLI PKCE follow-up ship together. Fix the intersection before that pairing lands.

Zingzy:
Correcting my earlier review: I cleared this as "nothing blocking," and that was wrong. An app token holding `keys:manage` could `POST /api/v1/keys` with any scopes (including `admin:all`) and mint an API key that survives grant revocation, since revoke only deletes device tokens and never touches the api-keys collection. That turns a revocable delegated credential into a permanent one. My review checked the key-to-key self-propagation guard but missed the app-token-to-key path. Fixed in 0514b5d.

Decision: minting a credential is a first-party action, so key creation now requires an interactive session (`JwtVerifiedUser`). `keys:manage` covers list and delete only. This matches how other platforms treat credential creation and needs no scope-subsetting or cascade-revoke machinery, because no app can create a key at all. The tradeoff is that `spoo keys create` over device auth stops working; list and delete keep working, and creation moves to the dashboard. A spoo-cli follow-up will surface a clear message on that path.

Also fixed a related guard bug: the interactive-only surfaces keyed off the `scp` claim, so a legacy grant token (which carries `app_id` but no `scp`) slipped into `/me`, device revoke, and key management as if it were a session. Delegation is now detected by `app_id`/`api_key_doc`, not `scp`.

CodeRabbit's two items are in the same commit: `create_or_reactivate` takes `scopes` as a required keyword argument so reactivation can never silently erase a snapshot, and the apps response distinguishes an explicitly empty scope list from a legacy (None) grant instead of labelling both as full access.

Tests: full suite 2715 passing, including new coverage for an app token being denied key creation, a legacy scp-less token being denied interactive-only surfaces and key management, and the empty-vs-None response shape.

Zingzy (commented):
Re-checked through cbb0428. **Blocker resolved, and the fix is stronger than what I proposed.** I verified the closure end-to-end rather than trusting the diff.

Instead of scope-subsetting the key-creation path (my suggestion), you made credential minting first-party only: `create_api_key` now takes `JwtVerifiedUser`, so no API key and no app token — regardless of scope — can mint a key. That is the better call: it removes the laundering surface entirely rather than bounding it, so there is no cascade-revoke problem to reason about and no future scope that could re-open it. `keys:manage` correctly degrades to list+delete, and the SCOPE_DESCRIPTIONS copy follows ("List and revoke your API keys"). The `test_app_token_cannot_create_key` test flipped from asserting 201 to 403 — the exact exploit I found is now pinned closed. I re-ran the create-path guard against a keys:manage app token in the .venv: 403 "This operation requires an interactive session."

**Should-fix also resolved, and more completely than flagged.** Both `require_jwt` and `require_session_or_scopes` now discriminate delegation by `app_id`/`api_key_doc` rather than the `scp` claim, so a legacy scp-less app token (app_id set, no scp) is treated as delegated everywhere — barred from `/me`, profile edit, device revoke, AND key management, instead of passing as a session. New tests pin exactly that shape (`_make_legacy_app_token`, `test_legacy_app_token_403_*` at both the route and unit layer). The guard now enforces the invariant its docstring promises.

Also good: `create_or_reactivate(scopes=...)` is now keyword-only with no default, so a caller can't silently mint a scopeless (unrestricted) grant by omission.

Ran the auth suites (118 tests across the app-token-scope, keys, device-auth, and auth-deps files) green in a clean worktree; CI green. No new surface introduced by the delta. The accepted product tradeoff — `spoo keys create` over device auth now 403s, dropped from the CLI in the follow-up — is the right line: minting a credential is a human, first-party act. Good to merge (still backend-first ahead of the CLI/Raycast/frontend PRs).

PR #266, opened 2026-07-22
feat: api key last_used_at and failed auth logging

Zingzy:
Stacked on #265.

## What

- `last_used_at` on API key documents, stamped on successful auth and returned by `GET /api/v1/keys` as a Unix timestamp (null when never used).
- Failed API key auth is now logged as `api_key_auth_failed` with a reason (`unknown`, `revoked`, `expired`).

## Performance

The stamp is debounced: it only writes when the stored value is missing or older than one hour, so a key making 100k requests a day costs about 24 single-document writes. The write is best-effort inside a try/except; a failed or slow stamp can never fail or block authentication. The hot path stays read-only otherwise.

## Security

Failure logs carry the display prefix only (the same 8 characters the dashboard shows), never hashes or raw material. Volume is bounded upstream by the rate limiter. Revoked and expired entries include `key_id` and `user_id` so a revoked key still being retried by a forgotten script is visible and attributable.

## Notes

- `last_used_at` gives the dashboard a "never used / last used N days ago" signal, which is the piece that makes revoking old keys feel safe.
- Existing key documents simply read as null until first use after deploy.
- Unit tests cover the debounce window, naive datetime handling from Mongo, stamp failure tolerance, and all three failure reasons.

## Summary by CodeRabbit

* **New Features**
  * API key listings now show when each key was last used, or indicate that it has never been used.
  * API key usage timestamps are recorded automatically after successful authentication.

* **Bug Fixes**
  * Authentication remains successful even if updating usage metadata fails.
  * Improved handling of UTC timestamps and clearer diagnostics for invalid, revoked, or expired API keys.

### DIFF
--- a/dependencies/auth.py
+++ b/dependencies/auth.py
@@ -24,6 +24,7 @@ from repositories.api_key_repository import ApiKeyRepository
 from repositories.user_repository import UserRepository
 from schemas.dto.requests.api_key import ApiKeyScope
 from schemas.models.api_key import ApiKeyDoc
+from shared.datetime_utils import as_aware_utc
 
 log = get_logger(__name__)
 
@@ -96,9 +97,8 @@ async def get_current_user(
             except Exception:
                 return None
 
-            # Failed attempts log the display prefix only (same 8 chars the
-            # dashboard shows) — enough to identify the credential, useless
-            # for recovering it. Volume is bounded by the rate limiter.
+            # Failure logs carry the display prefix only — never hashes or
+            # raw key material.
             if key is None:
                 log.warning("api_key_auth_failed", reason="unknown", key_prefix=raw[:8])
                 return None
@@ -113,21 +113,16 @@ async def get_current_user(
                 return None
 
             now = datetime.now(timezone.utc)
-            if key.expires_at:
-                exp = (
-                    key.expires_at.replace(tzinfo=timezone.utc)
-                    if key.expires_at.tzinfo is None
-                    else key.expires_at
+            exp = as_aware_utc(key.expires_at)
+            if exp is not None and exp <= now:
+                log.warning(
+                    "api_key_auth_failed",
+                    reason="expired",
+                    key_prefix=key.token_prefix,
+                    key_id=str(key.id),
+                    user_id=str(key.user_id),
                 )
-                if exp <= now:
-                    log.warning(
-                        "api_key_auth_failed",
-                        reason="expired",
-                        key_prefix=key.token_prefix,
-                        key_id=str(key.id),
-                        user_id=str(key.user_id),
-                    )
-                    return None
+                return None
 
             try:
                 user = await UserRepository(db["users"]).find_by_id(key.user_id)
@@ -136,9 +131,7 @@ async def get_current_user(
 
             # Best-effort last-used stamp — debounced so a busy key costs at
             # most one extra write per hour, and never fails the request.
-            last_used = key.last_used_at
-            if last_used is not None and last_used.tzinfo is None:
-                last_used = last_used.replace(tzinfo=timezone.utc)
+            last_used = as_aware_utc(key.last_used_at)
             if last_used is None or now - last_used > _LAST_USED_DEBOUNCE:
                 try:
                     await key_repo.touch_last_used(key.id)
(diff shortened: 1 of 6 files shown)


dependencies/auth.py
```diff
@@ -86,12 +90,26 @@ async def get_current_user(
...
             try:
-                key = await ApiKeyRepository(db["api-keys"]).find_by_hash(token_hash)
+                key = await key_repo.find_by_hash(token_hash)
             except Exception:
                 return None
 
-            if key is None or key.revoked:
+            # Failed attempts log the display prefix only (same 8 chars the
+            # dashboard shows) — enough to identify the credential, useless
+            # for recovering it. Volume is bounded by the rate limiter.
```
Zingzy: Nit: "Volume is bounded by the rate limiter" is justification for the reviewer, not a constraint the next editor needs. The part worth keeping is the invariant: log the display prefix only, never raw material. One line covers it.

dependencies/auth.py
```diff
@@ -102,13 +120,31 @@ async def get_current_user(
...
 
             try:
                 user = await UserRepository(db["users"]).find_by_id(key.user_id)
             except Exception:
                 return None
 
+            # Best-effort last-used stamp — debounced so a busy key costs at
+            # most one extra write per hour, and never fails the request.
+            last_used = key.last_used_at
+            if last_used is not None and last_used.tzinfo is None:
```
Zingzy: Nit: `shared.datetime_utils.as_aware_utc` exists for exactly this naive-UTC normalization. The expiry check above hand-rolls the same thing, so this matches the file as it stands; a follow-up could switch both to the helper.

tests/unit/test_auth_deps.py
```diff
@@ -563,3 +566,144 @@ async def test_legacy_app_token_403(self):
...
             await require_keys_access(user)
+
+
+# ── TestApiKeyHygiene ────────────────────────────────────────────────────────
+
+
+class TestApiKeyHygiene:
+    """last_used_at stamping (debounced, best-effort) and failed-auth logging."""
+
+    def _mocks(self, key_doc):
```
Zingzy: Nit: `_mocks` takes `key_doc` but never uses it.

Zingzy (commented):
Mergeable, nothing blocking. Two claims verified, small nits inline.

Verified:
- Prefix parity: key creation stores `token_prefix = raw[:8]` (services/api_key_service.py:73), so the unknown-key log line exposes exactly what the dashboard already shows for real keys, nothing more.
- Merge-order coupling: `key_prefix` and `key_id` in `api_key_auth_failed` only survive the log redactor because of the safe list added in #265 (both names contain "key" and would otherwise land in Axiom as `***REDACTED***`). The stack ordering enforces this; do not cherry-pick this without #265.

Notes:
- Concurrent stale auths can each fire `touch_last_used`; it is a single-document `$set` writing the same value, so harmless.
- DB errors during lookup still return None silently (pre-existing, unchanged here). `reason="unknown"` only fires when the lookup succeeded and found nothing, which is the right split: a Mongo outage should not masquerade as credential misuse.
- The debounce write is awaited inline, so roughly one API-key request per key per hour pays a single-document write of latency. Fine at current volume; if it ever shows up in duration percentiles, fire-and-forget is the escape hatch.

Affected test files pass at this head (341). Full matrix will run once this retargets to main.

dependencies/auth.py
Zingzy: Trimmed in af9aa97. The comment now states only the invariant: failure logs carry the display prefix, never hashes or raw key material.

dependencies/auth.py
Zingzy: Done in af9aa97. Both the expiry check and the last_used debounce now go through as_aware_utc, which also deleted the hand-rolled tzinfo branch above.

tests/unit/test_auth_deps.py
Zingzy: Dropped in af9aa97, all five call sites updated.

Zingzy:
All three resolved at af9aa97, verified in the diff rather than the descriptions: the comment now states just the invariant, both datetime normalizations go through `as_aware_utc` (behavior-equivalent: a None expires_at still skips the check, and the hand-rolled tzinfo branch is gone), and the unused test param is dropped from all five call sites. Nothing further.

PR #267, opened 2026-07-22
feat: carry resolved identity into request_completed logs

Zingzy:
Stacked on #266 (which stacks on #265).

## What

The auth dependency stashes the resolved identity (`user_id`, `auth_method`, and for API keys `key_id`/`key_prefix`, for app tokens `app_id`) on `request.state`; the request logging middleware merges it into the `request_completed` event.

## Why

`request_completed` currently carries no identity at all. The auth dependency binds structlog contextvars, but `BaseHTTPMiddleware` runs the downstream handler in a child task, so those bindings never propagate back to the middleware's dispatch context where `request_completed` is emitted. Service events get identity, the request log does not. `request.state` rides the shared ASGI scope, so it crosses that boundary.

## Performance and security

No extra queries or allocations beyond a small dict on authenticated requests. Fields match exactly what service-layer logs already carry; nothing new is exposed. Anonymous requests and requests rejected before auth runs (rate-limited 429s) are unchanged.

Unit tests cover both the authenticated and anonymous shapes of the event.

### DIFF
--- a/dependencies/auth.py
+++ b/dependencies/auth.py
@@ -152,6 +152,16 @@ async def get_current_user(
                 key_id=str(key.id),
                 key_prefix=key.token_prefix,
             )
+            # Contextvars bound here don't reach the middleware's own
+            # request_completed log call (BaseHTTPMiddleware runs the
+            # handler in a child task), so the identity is also stashed
+            # on request.state, which is shared via the ASGI scope.
+            request.state.auth_ctx = {
+                "user_id": str(key.user_id),
+                "auth_method": "api_key",
+                "key_id": str(key.id),
+                "key_prefix": key.token_prefix,
+            }
             return CurrentUser(
                 user_id=key.user_id,
                 email_verified=email_verified,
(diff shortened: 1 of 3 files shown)


Zingzy (commented):
Mergeable, nothing blocking, and the mechanism is right.

- The task-boundary claim checks out: BaseHTTPMiddleware runs downstream in a child task, contextvars copied at spawn never propagate back, and `request.state` rides the shared ASGI scope. `request.state.tenant` from the tenant middleware is the existing precedent, and the `getattr(..., None)` read matches how the redirect routes consume that one.
- The `**auth_ctx` merge cannot collide with the explicit kwargs of `request_completed`, and the keys come from a fixed internal set, never from anything user-controlled.
- `key_prefix`/`key_id` in `request_completed` survive redaction via the #265 safe list, same coupling as #266.
- The comments here are the kind worth having: they state a constraint the code cannot show (why `request.state` instead of contextvars). Keep them.

Middleware and auth test files pass at this head (118).

PR #268, opened 2026-07-22
feat: persist created_via on shortened urls

Zingzy:
Stacked on #267 (chain: #265 -> #266 -> #267 -> this).

## What

- `UrlV2Doc` gains a `created_via` field storing the validated `X-Spoo-Client` slug (`dashboard`, `landing`, `snap`, `raycast`, `cli`, `bot`) at creation time. Null for untagged creations.
- The header parser moves from the logging middleware to `shared/client_tag.py` so the middleware and the shorten route share one implementation.
- `POST /api/v1/shorten` parses the header and passes the slug into `UrlService.create` as an explicit parameter.

## Why

Log-based attribution expires with log retention. Persisting provenance on the document makes "where did this link come from" queryable forever, supports per-client counts straight from Mongo, and leaves room for a created-via hint in the dashboard later.

## Security and performance

The stored value is never the raw header: it must match `^[a-z0-9_-]{1,32}$` or it is dropped to null, so arbitrary header content cannot reach the database. No extra queries; one nullable field on insert. The field is not exposed on any public response DTO.

## Notes

- Legacy routes are untouched; legacy clients never send the header.
- Existing documents read as null, no migration needed.

### DIFF
--- a/routes/api_v1/shorten.py
+++ b/routes/api_v1/shorten.py
@@ -30,6 +30,7 @@ from middleware.rate_limiter import Limits, dynamic_limit, limiter
 from schemas.dto.requests.url import AliasCheckQuery, CreateUrlRequest
 from schemas.dto.responses.url import AliasCheckResponse, UrlResponse
 from services.feature_flag_service import GEO_TARGETING_FLAG, META_TAGS_FLAG
+from shared.client_tag import CLIENT_TAG_HEADER, parse_client_tag
 from shared.ip_utils import get_client_ip
 
 router = APIRouter(tags=["URL Shortening"])
@@ -116,7 +117,10 @@ async def shorten_v1(
         base_url = settings.app_url
         scoped_domain = None
 
-    doc = await url_service.create(body, owner_id, client_ip, domain=scoped_domain)
+    created_via, _ = parse_client_tag(request.headers.get(CLIENT_TAG_HEADER))
+    doc = await url_service.create(
+        body, owner_id, client_ip, domain=scoped_domain, created_via=created_via
+    )
     return UrlResponse.from_doc(doc, base_url)
 
 
--- a/routes/api_v1/shorten.py
+++ b/routes/api_v1/shorten.py
@@ -30,7 +30,7 @@ from middleware.rate_limiter import Limits, dynamic_limit, limiter
 from schemas.dto.requests.url import AliasCheckQuery, CreateUrlRequest
 from schemas.dto.responses.url import AliasCheckResponse, UrlResponse
 from services.feature_flag_service import GEO_TARGETING_FLAG, META_TAGS_FLAG
-from shared.client_tag import CLIENT_TAG_HEADER, parse_client_tag
+from shared.client_tag import CLIENT_TAG_HEADER, first_party_client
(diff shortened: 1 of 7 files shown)


routes/api_v1/shorten.py
```diff
@@ -116,7 +117,10 @@ async def shorten_v1(
         base_url = settings.app_url
         scoped_domain = None
 
-    doc = await url_service.create(body, owner_id, client_ip, domain=scoped_domain)
+    created_via, _ = parse_client_tag(request.headers.get(CLIENT_TAG_HEADER))
```
Zingzy: Should-decide: persistence accepts any shape-valid slug, not just the six first-party ones. `curl -H 'X-Spoo-Client: whatever'` stamps `created_via="whatever"` on the document forever. For logs, advisory-only is clearly right (a new client shows up without a deploy). Durable data is a different calculus: junk slugs sit in Mongo indefinitely and per-client group-bys need a known-slug filter to stay clean.

Options: (a) keep as-is and filter to known slugs at query time; (b) export a first-party slug set from `shared/client_tag.py` and persist only members, with logging still keeping the raw slug. Cost of (b) is that a new first-party client needs a backend deploy before its provenance persists. Blast radius of (a) is small (32 lowercase chars, the writer's own document, never exposed), so either is defensible. Just make the call deliberately, since mis-tagged history cannot be re-tagged.

Zingzy (commented):
Mergeable. One decision to make consciously before this lands, inline below. Everything else verified clean.

Verified:
- `shorten_v1` is the only v2 creation path (the emoji router only serves the emoji set, bulk has no create), so the single plumbing point covers all tagged creations; legacy stays untouched as stated.
- `created_via` appears in no response DTO and no route serialization.
- Old documents and anything cached before deploy decode fine: nullable field, None default, nothing reads it yet.
- `shared/client_tag.py` fits the shared/ convention exactly (pure function, no I/O, no framework import), and moving the parser out of the middleware instead of importing middleware code from a route was the right direction.

Chain note: merge bottom-up and let CI re-run after each retarget; the stacked PRs only ran check_formatting because the test workflows filter on the base branch. Full unit suite passes at this tip locally (2001).

routes/api_v1/shorten.py
Zingzy: Went with (b) in dde86cf. shared/client_tag.py now exports FIRST_PARTY_CLIENTS and a first_party_client() helper; the route persists only members while logging keeps the raw slug. The deploy cost of (b) is a non-issue in practice since a new first-party client already ships through this repo's release train, and untagged history that cannot be re-tagged felt like the worse permanent outcome.

Zingzy: Verified at dde86cf: `first_party_client` persists members only, `parse_client_tag` keeps raw slugs visible for logging, and the test table covers the unknown-slug and `curl/8.0` cases. The comment on `FIRST_PARTY_CLIENTS` makes the log-vs-persist split self-documenting. Agreed on the reasoning: the release train makes the deploy cost of (b) negligible, and untagged history was the only irreversible outcome on the table. Unit suite passes at the tip (2010).

PR #269, opened 2026-07-22
refactor: move operator Discord pings out of infrastructure/webhook

Zingzy:
## What

`infrastructure/webhook/` was never a webhooks feature. It is the internal transport that pings the maintainer's Discord when someone submits the contact form or reports a URL. It only ever used Discord webhook URLs as a delivery mechanism, so the name oversold it and squatted on vocabulary we want to keep clean.

This PR rebuilds it as `infrastructure/ops_notify.py`, a single flat module:

- `OpsNotifier` protocol with semantic methods: `contact_message()`, `url_report()`, `report_summary()`
- `DiscordOpsNotifier` implementation that owns channel routing (contact vs reports) and all embed formatting

## Why the semantic protocol

The old `WebhookProvider.send(payload)` claimed to be a swappable transport, but callers built Discord-specific embed dicts themselves, so swapping Discord for anything else meant editing every calling service anyway. Now services state facts and the notifier owns presentation. `ContactService` and `ReportIntakeService` keep only their failure policies (contact treats a failed send as a user-visible error, report intake logs and moves on since storage is the system of record).

## Behavior

No functional change:

- Embed structures are preserved byte for byte; the integration tests pin the exact Discord bodies by running the real notifier over a capturing HTTP fake
- `CONTACT_WEBHOOK` / `URL_REPORT_WEBHOOK` env vars keep their names (they are Discord webhook URLs, and self-hosting docs reference them)
- The 503 not-configured gates in the routes are untouched
- Transport log events renamed to `ops_notify_*`

## Tests

Full suite passes (2751). `test_webhook.py` renamed to `test_ops_notify.py` and extended with channel-routing and embed-content coverage that moved down from the service tests.

## Summary by CodeRabbit

* **New Features**
  * Added centralized operator notifications for contact messages, URL reports, and report submission summaries.
  * Notifications now use dedicated contact and report channels with consistent formatting.
  * Report summaries include submission details, accepted and rejected results, reporter information, and source IP.

* **Bug Fixes**
  * Improved notification failure handling so delivery errors are reported without disrupting report submissions.


infrastructure/ops_notify.py
```diff
@@ -0,0 +1,210 @@
...
+        ip: str,
+        now: datetime,
+    ) -> bool: ...
+
+
+class DiscordOpsNotifier:
+    """Discord implementation — routes each notification to its channel
+    (contact vs reports) and builds the embeds.
+
+    Embed structures are preserved exactly from the previous per-service
```
Zingzy: Nit: "preserved exactly from the previous per-service builders (originally `utils/contact_utils.py`)" is history that rots the moment this merges — the next reader has no "previous builders" to compare against, and the integration tests are what actually pin the shapes now. One line like "Embed shapes are pinned by the integration tests (test_contact / test_reports)" tells the next editor the real constraint: change a field and a test breaks.

Zingzy (commented):
Mergeable, nothing blocking. This is the right refactor, and the interesting part is the protocol inversion, not the rename.

Architecture:
- `WebhookProvider.send(payload)` was a fake seam: it claimed transport-swappability while every caller hand-built Discord embed dicts, so no transport was ever actually swappable. `OpsNotifier` with semantic methods puts presentation in the implementation and facts in the callers, which is what the protocol should have been from the start. The PR body's own diagnosis is correct.
- The display-target computation (`domain or system_default_domain`) staying in `ReportIntakeService` while formatting moves out is the right split: which target a report resolves to is a domain fact, and the notifier has no business knowing the system default domain.
- The two failure policies are now stated where they live: contact raises because the notification IS the deliverable; intake logs and moves on because storage is the system of record. Both preserved exactly.
- Keeping `CONTACT_WEBHOOK`/`URL_REPORT_WEBHOOK` env names honors the shipped self-host contract, and the wiring comment pins that constraint for the next editor.
- Flat module over a `protocol.py` + impl package (the captcha layout): with one implementation and ~200 lines, flat is the right call; the package split isn't a rule worth paying a directory for here.

Verified rather than trusted:
- No dangling references to `infrastructure.webhook` or `WebhookProvider` anywhere on the branch (the one hit is the new module's own docstring).
- The log event renames break nothing in-repo: a repo-wide grep including `axiom/dashboards/*.json` finds nothing charting the old `discord_webhook_*` / `*_webhook_send_failed` names. If any Axiom monitor outside the repo watches those, update it at deploy time; otherwise the rename is free.
- Embed bodies compared field by field against the deleted builders: titles, colors, field order, code-block wrapping, footer, and timestamp semantics (including `now` passed through for the summary so the embed and audit record agree) are identical.
- The integration-test upgrade is a real strengthening, not a rename: running the actual `DiscordOpsNotifier` over a capturing HTTP fake pins the exact wire bodies Discord receives, where the old fake stopped at the protocol boundary.
- 57 affected tests pass at this head locally; full matrix is green.

One docstring nit inline.

infrastructure/ops_notify.py
Zingzy: Fixed in 4e52e12: the docstring now states the constraint (embed shapes are pinned by test_contact / test_reports over a capturing HTTP fake) instead of pointing at builders that no longer exist.

Zingzy: Verified at 4e52e12: comment-only delta, the docstring now names the actual enforcement (the integration tests over the capturing fake) instead of history. Nothing further, still mergeable.

PR #270, opened 2026-07-22
feat: webhooks

### LINKED ISSUE #271: [FEAT] Webhooks: real-time event deliveries for links
## Explanation

Add a webhooks system to spoo.me: users register HTTPS endpoints and receive signed, real-time event deliveries when things happen to their links (created, updated, deleted, clicked, expired). This turns spoo.me from a tool you visit into infrastructure other tools can build on: custom integrations, Discord/Slack notifications, analytics pipelines, and an eventual Zapier app all become possible from one primitive.

## Is your feature request related to a problem? Please describe.

Everything your links do is currently trapped inside the dashboard. You can look at stats whenever you want, but spoo.me never speaks first. There is no way to react to a click in your own system, to know a link expired without checking, or to pipe link activity into the tools a team already lives in.

## Describe the solution you'd like

- An endpoint management API and dashboard section: register up to 5 HTTPS endpoints, choose event types (with `link.*` and `*` wildcards), optionally scope to specific links.
- Five event types: `link.created`, `link.updated` (with a field-level old/new changes map), `link.deleted`, `link.clicked` (full analytics dimensions), `link.expired`.
- [Standard Webhooks](https://www.standardwebhooks.com/) compliance: HMAC-SHA256 signatures, `webhook-id` / `webhook-timestamp` / `webhook-signature` headers, `whsec_` secrets shown once at creation.
- At-least-once delivery with automatic retries over ~10 hours, automatic disabling of dead endpoints, and a per-endpoint delivery log showing every attempt.
- A public event catalog endpoint with exact sample payloads, and test sends that deliver any event type's sample through the real pipeline.
- Payload presentation flavors so a Discord or Slack webhook URL can be used directly, without writing any code (follow-up).

## Describe alternatives you've considered

- Polling the stats API: works for dashboards, not for reactions; forces integrators to build diffing and scheduling for something the server already knows the moment it happens.
- Building a Zapier app directly without webhooks underneath: webhooks are the primitive Zapier triggers need anyway, and they serve every other integration path at the same time.
- Client-side value filtering only (no link scoping): a busy account subscribing to clicks would drown; per-link scoping at the source is the one filter with a volume argument.

----

## Code Example

```bash
# Register an endpoint (secret is returned once, store it)
curl -X POST https://spoo.me/api/v1/webhooks \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -d '{"url": "https://api.yourapp.com/hooks/spoo", "events": ["link.clicked", "link.expired"]}'
```

```python
# Verify a delivery (Standard Webhooks)
from standardwebhooks import Webhook

wh = Webhook("whsec_...")
payload = wh.verify(request.body, request.headers)  # raises on bad signature
if payload["type"] == "link.clicked":
    print(payload["data"]["alias"], payload["data"]["country"])
```

## Additional context

Backend implementation is in progress in #270 (merges dark behind a feature flag). Event payloads never contain password material or raw IP addresses; endpoint URLs are SSRF-checked at registration and again before every delivery. A full architecture writeup follows in the comments.

Zingzy:
## What

Real-time event deliveries to subscriber URLs, following the [Standard Webhooks](https://www.standardwebhooks.com/) specification. When something happens to your links, spoo.me tells your other tools: your own server, a Zapier hook, anything that accepts an HTTPS POST.

Fully opt-in and dark by default: `WEBHOOKS_ENABLED=false` wires a null sink and mounts nothing, and access is additionally gated per user by the `webhooks` feature flag (default deny, 403 when not granted).

## Events

Five event types, one rule: events split by who caused them. Actor edits ride `link.updated` and its `changes` map (status included); system-discovered facts get named events.

| Event | Fires when |
|---|---|
| `link.created` | a link is created |
| `link.updated` | a link is edited; `changes` maps each field to old/new |
| `link.deleted` | a link is deleted |
| `link.clicked` | every tracked click, bots included (`is_bot` in payload) |
| `link.expired` | expiry is discovered (max clicks or time), once per link |

Payloads carry the full public link snapshot plus event context. `link.clicked` is flat and dimensional: country/city, browser/os/device, bounded raw user agent, referrer, UTM, `long_url`, running click total. Bulk operations emit per item, indistinguishable from a loop. Nothing password-shaped, IP-shaped, or internal ever rides the wire: password fields become presence booleans, meta tags reduce to their four public fields, and payload validation is pinned to the registry models by test.

`GET /api/v1/webhooks/event-types` is a public catalog with exact sample payloads; the same fixtures power test sends, so docs and behavior cannot drift.

## Delivery

- Standard Webhooks signing: `webhook-id` / `webhook-timestamp` / `webhook-signature` headers, HMAC-SHA256, `whsec_` secrets shown once and stored AES-GCM encrypted (the server signs at delivery time, so hash-only storage is not an option). Verified against the published reference test vector.
- At-least-once with retries: immediate, 5s, 5m, 30m, 2h, 5h, 10h. `webhook-id` is stable across retries and bodies are frozen after first render, so consumer dedup works on either.
- 410 Gone disables an endpoint immediately; ten consecutive exhausted deliveries disable it too.
- Per-endpoint pending cap (default 1000): beyond it deliveries are dropped and counted, and the next successful delivery carries `dropped_since_last` as a reconcile signal.
- Delivery log per endpoint: every attempt with status, latency, error, response snippet, and the exact body sent. 30-day TTL, coupled to the event store so a delivery can never outlive its event.

## Architecture

Two Redis Streams feed a dispatcher: the existing click stream gains a `webhooks` consumer group (reader plus claimer pair, DLQ guard, same machinery as the stats and hotness groups), and a new low-volume `events:domain` stream carries link lifecycle facts from `UrlService` and the bulk operations. The dispatcher matches subscriptions (owner-level Redis cache in front, so accounts without webhooks cost nothing on the click path) and records delivery intent in Mongo. A claim-loop executor owns delivery: atomic `findOneAndUpdate` claims with leases make N concurrent executors safe, and the loop needs only Mongo, so it runs in the click worker in production or embedded in the app process for self-hosts without a queue Redis. Deployments without any Redis still work: dispatch happens inline at emit time.

Events are stored once per fact; delivery rows are thin references. Rendering happens at first attempt, keeping dispatch insert-only on the consumer ack path.

## Security

- Endpoint URLs are SSRF-checked at registration and re-resolved before every delivery (public addresses only, connection pinned to the resolved IP, redirects refused), reusing the existing `safe_fetch` hardening.
- Registration requires a verified email plus the feature flag; `webhooks:manage` and `webhooks:read` API key scopes; five endpoints per user (config), 20KB payload cap, tight rate limits on create and test sends.

## Testing

Full suite green (2815). New coverage: signing vectors, registry/wildcard expansion, matcher scoping, dispatcher fan-out and cap, executor retry/disable paths, route integration with frozen wire shapes, bulk emission parity. Also verified end to end against live Mongo and Redis with a real signed HTTPS delivery to a public echo endpoint.

## Rollout

Merges dark. Enabling requires `WEBHOOKS_ENABLED=true` plus seeding the `webhooks` feature flag. Env reference is in `.env.example`; an Axiom dashboard for dispatch/delivery health ships in `axiom/dashboards/`. Follow-ups tracked separately: endpoint auto-disable email notification, Discord/Slack payload flavors, dashboard UI.

## Summary by CodeRabbit

* **New Features**
  * Added webhook endpoint management API (create/list/get/update/delete), event catalog, synchronous test sends, delivery log listing, and manual redelivery retries with scoped event subscriptions.
  * Added signed webhook deliveries with endpoint health tracking, retry/backoff handling, and auto-disable on repeated failures.
  * Added link lifecycle/click webhook payloads with wildcard subscriptions and new Discord/Slack rendering flavors.
  * Added dispatch/delivery execution for worker and embedded modes.
* **Documentation**
  * Updated the example environment file with optional Webhooks configuration (off by default), including timeouts, caps, and secret requirements.
* **Monitoring / Tests**
  * Added a webhooks performance dashboard and expanded webhook test coverage.

Closes #271.

### DIFF
--- a/infrastructure/safe_fetch.py
+++ b/infrastructure/safe_fetch.py
@@ -248,6 +248,20 @@ class PostResult:
     status_code: int | None
     error: str | None
     body_snippet: str | None  # first 256 bytes, for the delivery log
+    # Parsed integer Retry-After, when the receiver sent one (429/503 flow
+    # control). HTTP-date form is not parsed — callers fall back to their
+    # own delay.
+    retry_after_seconds: float | None = None
+
+
+def _parse_retry_after(value: str | None) -> float | None:
+    if value is None:
+        return None
+    try:
+        seconds = float(value.strip())
+    except ValueError:
+        return None
+    return seconds if seconds >= 0 else None
 
 
 async def post_public(
@@ -295,7 +309,12 @@ async def post_public(
                 except (asyncio.TimeoutError, TimeoutError):
                     buf = bytearray()
                 snippet = bytes(buf).decode("utf-8", errors="replace") if buf else None
-                return PostResult(resp.status_code, None, snippet)
+                return PostResult(
+                    resp.status_code,
+                    None,
+                    snippet,
+                    _parse_retry_after(resp.headers.get("retry-after")),
+                )
             finally:
                 await resp.aclose()
     except (httpx.TimeoutException, httpx.TransportError) as exc:
--- a/infrastructure/safe_fetch.py
+++ b/infrastructure/safe_fetch.py
@@ -59,10 +59,19 @@ class FetchedBody:
     final_url: str
 
 
+_NAT64_PREFIX = ipaddress.ip_network("64:ff9b::/96")
+
+
 def _is_public(ip: str) -> bool:
     addr = ipaddress.ip_address(ip)
-    if isinstance(addr, ipaddress.IPv6Address) and addr.ipv4_mapped:
-        addr = addr.ipv4_mapped
+    if isinstance(addr, ipaddress.IPv6Address):
+        if addr.ipv4_mapped:
+            addr = addr.ipv4_mapped
+        # NAT64 (RFC 6052) reports is_global=True, but on a NAT64/DNS64
+        # host 64:ff9b::7f00:1 reaches 127.0.0.1 — reject the whole
+        # prefix; a translation address is never a legitimate target.
+        elif addr in _NAT64_PREFIX:
+            return False
     # is_global (not a flag union) catches CGNAT 100.64.0.0/10 and
     # transitional ranges the union missed; exclude multicast, which is
     # is_global=True but must never be a fetch target.
@@ -260,7 +269,9 @@ async def post_public(
         if parsed.scheme != "https":
             return PostResult(None, "non-https URL", None)
         ip = await _resolve_public_ip(parsed.host)
-    except FetchHardError as exc:
+    except (FetchHardError, FetchTransientError, httpx.InvalidURL) as exc:
+        # Transient DNS failures included: delivery outcomes are DATA for
+        # the retry ladder, never exceptions that skip attempt recording.
         return PostResult(None, str(exc), None)
 
     pinned = parsed.copy_with(host=_bracket(ip))
--- a/infrastructure/safe_fetch.py
+++ b/infrastructure/safe_fetch.py
@@ -228,3 +228,74 @@ async def fetch_public_image(
         max_redirects=max_redirects,
         user_agent=user_agent,
     )
+
+
+@dataclass(frozen=True)
+class PostResult:
+    """Outcome of a webhook-style POST. Status outcomes are DATA here
+    (the delivery executor owns retry policy) — only SSRF violations and
+    transport errors surface as fields, never exceptions."""
+
+    status_code: int | None
+    error: str | None
+    body_snippet: str | None  # first 256 bytes, for the delivery log
+
+
+async def post_public(
+    url: str,
+    body: str,
+    *,
+    headers: dict[str, str],
+    timeout: float = 15.0,
+    snippet_bytes: int = 256,
+) -> PostResult:
+    """POST *body* to *url* with the same SSRF guards as fetch_public.
+
+    No redirects are followed — a redirect defeats IP pinning, so it is
+    reported as an error outcome. Never raises: webhook delivery failure
+    is a recorded fact, not an exception path.
+    """
+    try:
+        parsed = httpx.URL(url)
+        if parsed.scheme != "https":
+            return PostResult(None, "non-https URL", None)
+        ip = await _resolve_public_ip(parsed.host)
+    except FetchHardError as exc:
+        return PostResult(None, str(exc), None)
+
+    pinned = parsed.copy_with(host=_bracket(ip))
+    try:
+        async with httpx.AsyncClient(follow_redirects=False, timeout=timeout) as client:
+            request = client.build_request(
+                "POST",
+                pinned,
+                content=body.encode(),
+                headers={"Host": parsed.host, **headers},
+                extensions={"sni_hostname": parsed.host},
+            )
+            resp = await client.send(request, stream=True)
+            try:
+                if resp.status_code in _REDIRECT_STATUSES:
+                    return PostResult(resp.status_code, "redirect not followed", None)
+                try:
+                    buf = await asyncio.wait_for(
+                        _read_body(resp, snippet_bytes, True), timeout=timeout
+                    )
+                except (asyncio.TimeoutError, TimeoutError):
+                    buf = bytearray()
+                snippet = bytes(buf).decode("utf-8", errors="replace") if buf else None
+                return PostResult(resp.status_code, None, snippet)
+            finally:
+                await resp.aclose()
+    except (httpx.TimeoutException, httpx.TransportError) as exc:
+        return PostResult(None, f"{type(exc).__name__}: {exc}", None)
+
+
+async def validate_public_https_url(url: str) -> None:
+    """Creation-time gate for user-registered outbound URLs (webhook
+    endpoints): https + all resolved addresses public. Raises
+    FetchHardError with a user-safe message otherwise."""
+    parsed = httpx.URL(url)
+    if parsed.scheme != "https":
+        raise FetchHardError("URL must be https")
+    await _resolve_public_ip(parsed.host)
--- a/repositories/webhook_delivery_repository.py
+++ b/repositories/webhook_delivery_repository.py
@@ -3,7 +3,7 @@
 Owns the executor's claim semantics: ``claim_due`` is an atomic
 ``find_one_and_update`` (claim = lease via ``claimed_until``), which is
 what makes N concurrent executors safe with no scheduler infrastructure
-beyond Mongo itself (TRD §14). The claim query is stateless over
+beyond Mongo itself. The claim query is stateless over
 ``next_attempt_at``, so process restarts self-heal on the first loop.
 """
 
@@ -87,7 +87,7 @@ class WebhookDeliveryRepository(BaseRepository[WebhookDeliveryDoc]):
         return WebhookDeliveryDoc.from_mongo(doc)
 
     async def set_rendered_body(self, delivery_id: ObjectId, body: str) -> None:
-        """First attempt only — the body is frozen across retries (D15)."""
+        """First attempt only — the body is frozen across retries."""
         await self._update(
             {"_id": delivery_id, "rendered_body": None},
             {"$set": {"rendered_body": body}},
--- a/repositories/webhook_delivery_repository.py
+++ b/repositories/webhook_delivery_repository.py
@@ -1,0 +1,143 @@
+"""Repository for the ``webhook-deliveries`` collection.
+
+Owns the executor's claim semantics: ``claim_due`` is an atomic
+``find_one_and_update`` (claim = lease via ``claimed_until``), which is
+what makes N concurrent executors safe with no scheduler infrastructure
+beyond Mongo itself (TRD §14). The claim query is stateless over
+``next_attempt_at``, so process restarts self-heal on the first loop.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime, timedelta, timezone
+
+from bson import ObjectId
+from pymongo import ReturnDocument
+
+from repositories.base import BaseRepository
+from schemas.enums.webhook import DeliveryStatus
+from schemas.models.webhook import DeliveryAttempt, WebhookDeliveryDoc
+
+
+class WebhookDeliveryRepository(BaseRepository[WebhookDeliveryDoc]):
+    async def insert_many_rows(self, rows: list[dict]) -> None:
+        if not rows:
+            return
+        await self._col.insert_many(rows)
+
+    async def insert_row(self, row: dict) -> ObjectId:
+        return await self._insert(row)
+
+    async def find_by_id(self, delivery_id: ObjectId) -> WebhookDeliveryDoc | None:
+        return await self._find_one({"_id": delivery_id})
+
+    async def find_owned(
+        self, delivery_id: ObjectId, user_id: ObjectId
+    ) -> WebhookDeliveryDoc | None:
+        return await self._find_one({"_id": delivery_id, "user_id": user_id})
+
+    async def count_pending(self, endpoint_id: ObjectId) -> int:
+        return await self._count(
+            {"endpoint_id": endpoint_id, "status": DeliveryStatus.PENDING.value}
+        )
+
+    async def list_by_endpoint(
+        self,
+        endpoint_id: ObjectId,
+        *,
+        page: int,
+        page_size: int,
+        status: DeliveryStatus | None = None,
+    ) -> tuple[list[WebhookDeliveryDoc], int]:
+        query: dict = {"endpoint_id": endpoint_id}
+        if status is not None:
+            query["status"] = status.value
+        total = await self._count(query)
+        cursor = (
+            self._col.find(query)
+            .sort("created_at", -1)
+            .skip((page - 1) * page_size)
+            .limit(page_size)
+        )
+        docs = await cursor.to_list(length=page_size)
+        return [WebhookDeliveryDoc.from_mongo(d) for d in docs], total
+
+    # ── Executor surface ─────────────────────────────────────────────────
+
+    async def claim_due(self, *, lease_seconds: int = 60) -> WebhookDeliveryDoc | None:
+        """Atomically claim one due delivery, or None when nothing is due.
+
+        A crashed executor's claim expires with its lease — rows are never
+        stranded.
+        """
+        now = datetime.now(timezone.utc)
+        doc = await self._col.find_one_and_update(
+            {
+                "status": DeliveryStatus.PENDING.value,
+                "next_attempt_at": {"$lte": now},
+                "$or": [
+                    {"claimed_until": None},
+                    {"claimed_until": {"$lte": now}},
+                ],
+            },
+            {"$set": {"claimed_until": now + timedelta(seconds=lease_seconds)}},
+            sort=[("next_attempt_at", 1)],
+            return_document=ReturnDocument.AFTER,
+        )
+        return WebhookDeliveryDoc.from_mongo(doc)
+
+    async def set_rendered_body(self, delivery_id: ObjectId, body: str) -> None:
+        """First attempt only — the body is frozen across retries (D15)."""
+        await self._update(
+            {"_id": delivery_id, "rendered_body": None},
+            {"$set": {"rendered_body": body}},
+        )
+
+    async def record_attempt_and_reschedule(
+        self, delivery_id: ObjectId, attempt: DeliveryAttempt, next_attempt_at: datetime
+    ) -> None:
+        await self._update(
+            {"_id": delivery_id},
+            {
+                "$push": {"attempts": attempt.model_dump()},
+                "$inc": {"attempt_count": 1},
+                "$set": {"next_attempt_at": next_attempt_at, "claimed_until": None},
+            },
+        )
+
+    async def record_attempt_and_finish(
+        self, delivery_id: ObjectId, attempt: DeliveryAttempt, status: DeliveryStatus
+    ) -> None:
+        await self._update(
+            {"_id": delivery_id},
+            {
+                "$push": {"attempts": attempt.model_dump()},
+                "$inc": {"attempt_count": 1},
+                "$set": {
+                    "status": status.value,
+                    "completed_at": datetime.now(timezone.utc),
+                    "next_attempt_at": None,
+                    "claimed_until": None,
+                },
+            },
+        )
+
+    async def mark_failed(self, delivery_id: ObjectId, reason: str) -> None:
+        """Terminal failure without an HTTP attempt (endpoint inactive,
+        event row TTL-expired)."""
+        await self._update(
+            {"_id": delivery_id},
+            {
+                "$set": {
+                    "status": DeliveryStatus.FAILED.value,
+                    "completed_at": datetime.now(timezone.utc),
+                    "next_attempt_at": None,
+                    "claimed_until": None,
+                },
+                "$push": {
+                    "attempts": DeliveryAttempt(
+                        attempted_at=datetime.now(timezone.utc), error=reason
+                    ).model_dump()
+                },
+            },
+        )
--- a/repositories/webhook_delivery_repository.py
+++ b/repositories/webhook_delivery_repository.py
@@ -122,12 +122,27 @@ class WebhookDeliveryRepository(BaseRepository[WebhookDeliveryDoc]):
             },
         )
 
+    async def defer(self, delivery_id: ObjectId, *, delay_seconds: int) -> None:
+        """Push a delivery back without recording an attempt — used while
+        its endpoint is paused. Not a retry: the ladder is untouched."""
+        await self._update(
+            {"_id": delivery_id},
+            {
+                "$set": {
+                    "next_attempt_at": datetime.now(timezone.utc)
+                    + timedelta(seconds=delay_seconds),
+                    "claimed_until": None,
+                }
+            },
+        )
+
     async def mark_failed(self, delivery_id: ObjectId, reason: str) -> None:
         """Terminal failure without an HTTP attempt (endpoint inactive,
         event row TTL-expired)."""
         await self._update(
             {"_id": delivery_id},
             {
+                "$inc": {"attempt_count": 1},
                 "$set": {
                     "status": DeliveryStatus.FAILED.value,
                     "completed_at": datetime.now(timezone.utc),
--- a/services/webhooks/consumers.py
+++ b/services/webhooks/consumers.py
@@ -1,0 +1,98 @@
+"""Webhooks consumer-group handlers — framework-free like the click
+consumers; the worker's FastStream subscribers delegate here.
+
+Two feeds, one dispatcher:
+- ``events:clicks`` (existing stream, new ``webhooks`` group): ClickEvent
+  → link.clicked adaptation happens here, at read time — the producer
+  side of the click pipeline is untouched.
+- ``events:domain`` (new stream): the envelope is already a DomainEvent.
+
+Malformed payloads are dropped (never raised) — a payload that cannot
+parse today can never parse, and raising would leave it pending forever
+(the claim path's DLQ guard is the backstop for transient poison).
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from infrastructure.geoip import GeoIPService
+from infrastructure.logging import get_logger
+from services.click.events import click_event_from_payload
+from services.events.contract import domain_event_from_payload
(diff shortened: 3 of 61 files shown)


Zingzy (commented):
**Verdict: mergeable, and merging is safe — but nothing here should be enabled for a real user until the tier-1 list below is closed.** This is a genuinely strong piece of architecture: `WEBHOOKS_ENABLED` defaults false → `NullDomainEventSink`, producers stay unconditional, and I verified the production redirect path (stream rung) is byte-for-byte unchanged — webhook work lives in the worker, never in the GET response. So the risk gate is flag-enable time, not merge. I ran the full webhooks suite (65 pass) plus five adversarial deep-dives (SSRF, crypto, executor concurrency, hot-path cost, authz); the findings are what to finish before you seed the first `webhooks` flag, framed that way throughout.

## Architecture read — the load-bearing decisions are right

- The `services/events/` backbone as a **separate package with webhooks as one consumer** is the correct seam: alerts/audit slot in as sibling consumer groups without touching producers. Not speculative — it's already carrying two feeds (clicks + domain) in this PR.
- **Feature-off is genuinely free**, and I checked rather than trusted: NullSink `emit` is a no-op, no worker groups mount, the redirect path's click sink is only wrapped on the inline (no-queue-Redis) rung. In prod with queue Redis, GET /{code} is untouched.
- **Mongo-only executor, embedded OR worker, atomic lease-claim** (`find_one_and_update` on `ix_claim`) — elegant, needs no new infra, self-heals on restart, and N concurrent executors are safe. The claim/backoff/render-once/crash-window semantics are all correct (verified line by line).
- **Signing is spec-correct** — the published Standard Webhooks reference vector is pinned and passes; secret storage is authenticated AES-GCM with a random per-encryption nonce; the secret is returned exactly once and never logged. Confidentiality and integrity hold.
- **Registry-as-code** driving validation + catalog + samples + test sends + wildcard expansion (can't drift), with disciplined event governance (causer-not-field; `status_changed`/`bot.detected` cut with reasons). Payload sanitization holds across every event type — no `creation_ip`, no `owner_id`, no password hash; the earlier `updated_ip` leak fix survives.
- **IDOR: clean.** Every reachable by-id path enforces ownership in the Mongo query. Test send / delivery log / manual retry is Stripe-grade DX.

## Tier 1 — close before enabling the flag (correctness/security that will bite the first real subscriber or the operator)

**1. `SECRET_KEY` rotation turns every endpoint into an immortal poison delivery (live).** `_headers` calls `decrypt_secret` unguarded (`executor.py:223`). Rotating `SECRET_KEY` re-keys `_derive_aes_key`, so GCM auth fails with `InvalidTag` on every stored secret. That throws *before* any attempt is recorded → the loop's generic `except` logs `webhook_executor_tick_failed` and sleeps → the 60s lease expires → the same row is re-claimed → throws again, forever. The endpoint never disables and never exhausts. `SECRET_KEY` rotation is a routine security op (it's how JWTs get invalidated in this very ecosystem), and it silently converts webhooks into an infinite retry storm with no operator signal beyond a log line. Wrap decrypt: on failure, mark the delivery terminally failed and disable the endpoint with a distinct reason, so rotation degrades loudly and recoverably.

**2. `dropped_since_last` is dead code — the flagship event loses data silently.** `make_delivery_row` hardcodes `dropped_since_last: 0` (`dispatcher.py:41`); `record_success` drains the counter and *returns* it, but the caller discards the return (`executor.py:138`); nothing ever reads `endpoint.dropped_count` when building a row. So the pending-cap drop signal increments a counter that's zeroed on the next success and **never reaches a consumer**. For `link.clicked` feeding an analytics pipeline — the whole "turn spoo into infrastructure" use case — a subscriber who falls behind the 1000-cap loses clicks with zero indication, and the mechanism designed to tell them (D13) is unwired. Wire it (carry the endpoint's current dropped_count onto the next row and reset on that row's success) or cut it and log the drop loudly. The `dropped_since_last=42` test builds a state the pipeline can never produce, so it green-lights the dead path.

**3. Test-send and manual-retry mutate real endpoint health and enroll in the background ladder.** `is_test` is used only for logging (`executor.py:147`); nothing guards the health counters or the reschedule. Consequences: a transient-failed **test** send calls `record_attempt_and_reschedule`, setting `next_attempt_at` → the row becomes claimable and silently becomes a 10-hour background delivery (contradicting the "only executor this row meets" contract); exhausted test sends call `record_exhausted`, so a user clicking "test" on a broken endpoint ~10 times **auto-disables their own endpoint**; a passing test calls `record_success`, resetting a real failure streak and masking a failing endpoint. Test/retry should be single-shot and must not touch `consecutive_failures`/`total_*` or the retry schedule.

## Tier 2 — hardening before GA (abuse/reputation/trust, given spoo's phishing-flag history)

**4. Endpoints auto-disable silently.** The `on_disabled` hook is `None` in both wiring sites, so an endpoint dying after 10 failures notifies the user only via a log event. For infrastructure positioning, a user's integration going dark with no email is a trust-breaker — wire the notification before GA.

**5. Abuse/reputation cluster — decide the kill-switch.** `POST /{id}/test` runs the real pipeline synchronously and returns `status_code` + 256-byte `response_body` + latency for any user-supplied public HTTPS URL. That's an authenticated "fetch the first 256 bytes + status + timing of any HTTPS host, from spoo's egress IP" oracle/amplifier (~600 probes/hr/credential). Internal SSRF is blocked (the guard is tight — see below), so this is external-only, but it's still spoo's IP hitting third parties on demand. Compounding it: per-click fanout POSTs to user URLs make spoo a potential reflector, there's no per-user *delivery-rate* cap (only a 5-endpoint count cap), and flipping the `webhooks` flag off does **not** stop `/test` or existing deliveries — only CREATE is gated. So the flag is an onboarding gate, not an abuse lever. Decide the actual abuse response (per-endpoint disable is the only current one) and consider a global delivery-rate cap.

**6. The delivery-time SSRF guard has zero direct tests.** The guard itself is solid — I tried hard: https-only + resolve-then-pin-to-IP + no-redirects blocks loopback/RFC1918/link-local/CGNAT/metadata-IP/userinfo/DNS-rebind/redirect-to-Redis. But `post_public` has **no** direct test asserting it rejects http, rejects a private IP, or refuses redirects — `test_executor` and `test_webhooks` patch it out entirely. A future refactor silently regresses SSRF. Add the tests. One real gap in the guard: `64:ff9b::/96` (NAT64) returns `is_global=True`, so `https://[64:ff9b::7f00:1]/` passes `_is_public` (`safe_fetch.py:69`) and on a NAT64/DNS64 host routes to an internal HTTPS service — reject the prefix explicitly.

**7. Bulk ops add per-item awaited work even when the feature is off.** Bulk delete/update emit one event per item in an awaited loop, each doing a `UrlV2Doc.from_mongo` re-parse (`bulk_url_service.py:634`) with no `enabled` guard, plus a sequential XADD per item in stream mode. A 1000-link bulk op adds ~1000 re-parses (feature off) or ~1000 serial XADDs (on) to that request. Batch the writes and guard the re-parse behind the sink being non-null.

## Nits

- **Signing docstring lies about the security model.** `signing.py:6` says the secret is "stored hash-only (SHA-256)" — it's AES-GCM encrypted (it has to be; the server signs with it). `hash_signing_secret` is dead in production (test-only). Fix the docstring and drop the function — a maintainer trusting it would assume secrets are irreversible at rest.
- **Rotation scaffolding is inert.** `previous_secret_enc`/`previous_secret_expires_at` and the dual-signing branch exist but no route ever writes them, so rotation grace can't happen. Fine as a Phase-2 seam — but note that when you *do* wire rotation, the grace-window comparison at `executor.py:229` compares a naive Mongo datetime against an aware `now()` and will `TypeError` (the loop-poisoning kind); use `as_aware_utc()` (the repo already has it). Cheap to fix now while it's dead.
- **Pending-cap and quota are both racy** (count-then-write): concurrent dispatch can breach the 1000 cap, concurrent creates can exceed 5 endpoints. The quota race mirrors the existing api-key pattern, so it's a consistent stance, not a regression — noting for completeness.
- **`update_fields`/`list_by_endpoint` enforce ownership via a preceding fetch, not in their own filter** — safe today, but a latent regression trap; fold `user_id` into the query like `delete_endpoint` does.

## Product sense

The catalog scope is right: five events (created/updated/deleted/clicked/expired), fat snapshots so consumers don't need a follow-up call, additive-only contract, raw-flavor first. `link.clicked` is correctly the flagship — it's what makes real-time analytics/Zapier pipelines possible — which is exactly why Tier-1 #2 (silent drop, dead drop-signal) matters most: the highest-value event is the one that loses data quietly under load. Deferring Discord/Slack flavors to Phase 2 is defensible, though "point a webhook at your Discord" is the single most-requested easy win for this indie audience and the scaffolding's already there. The debuggability triad (test send + full delivery log with rendered body + manual retry) is the right DX bet and genuinely best-in-class — it just needs Tier-1 #3 fixed so the test button can't disable the thing it's testing. Net: the product thesis is sound and the surface is well-chosen; the gap is reliability-under-load and silent-failure trust, which is precisely what an "infrastructure" positioning is judged on.

## Merge-safety summary

Merge is safe: dark by default, hot path verified untouched, CI green across the matrix, feature-off cost negligible. Nothing above blocks the merge. Tier 1 blocks **enabling the flag for a real user** (data loss + operator-facing crash loop + self-disabling test button); Tier 2 blocks **GA** (silent disable, abuse levers, SSRF test coverage). Sequence the tail as: land the events backbone + management API now (dark) → close Tier 1 → seed your own flag and dogfood `link.clicked` at volume → wire disable-email + abuse cap → open the flag.

services/webhooks/executor.py
```diff
@@ -0,0 +1,258 @@
...
+        await self._deliveries.set_rendered_body(row.id, body)
+        return body
+
+    def _headers(
+        self, webhook_id: str, body: str, endpoint: WebhookEndpointDoc
+    ) -> dict[str, str]:
+        """Fresh timestamp per attempt (replay protection); dual signatures
+        during a rotation grace window, space-delimited per the spec."""
+        ts = int(time.time())
+        secret = decrypt_secret(
```
Zingzy, line 309: **Tier 1 (live).** `decrypt_secret` is unguarded here, and it runs before any attempt is recorded. A `SECRET_KEY` rotation re-keys `_derive_aes_key`, so every stored secret fails GCM auth with `InvalidTag` → this throws → the loop's generic `except` logs and sleeps → the 60s lease expires → the same row is re-claimed → throws again, forever. No disable, no exhaustion, no recovery, one log line. `SECRET_KEY` rotation is a routine op (it's how JWTs get invalidated), so this converts webhooks into a silent infinite retry storm. Catch decrypt failure: mark the delivery terminally failed and disable the endpoint with a distinct reason so rotation degrades loudly and recoverably.

services/webhooks/dispatcher.py
```diff
@@ -0,0 +1,98 @@
...
+) -> dict:
+    return {
+        "endpoint_id": endpoint.id,
+        "user_id": endpoint.user_id,
+        "event_oid": event_oid,
+        "event_type": event.type,
+        "webhook_id": new_webhook_id(),
+        "is_test": False,
+        "rendered_body": None,
+        "dropped_since_last": 0,
```
Zingzy: **Tier 1.** `dropped_since_last` is dead. This hardcodes 0; `record_success` drains `dropped_count` and returns it but the caller (`executor.py:138`) discards the return; nothing reads `endpoint.dropped_count` when building a row. So the pending-cap drop signal increments a counter that's zeroed on the next success and never reaches a consumer. For `link.clicked` feeding a pipeline, a subscriber past the 1000-cap loses clicks silently and the D13 mechanism meant to tell them is unwired. Either carry the endpoint's current dropped_count onto the next row (reset on that row's success) or cut the field and log the drop loudly. The `dropped_since_last=42` test builds a state the pipeline can't produce, so it masks this.

services/webhooks/executor.py
```diff
@@ -0,0 +1,258 @@
...
+            await self._endpoints.record_success(endpoint.id)
+            log.info(
+                "webhook_delivered",
+                endpoint_id=str(endpoint.id),
+                event_type=row.event_type,
+                webhook_id=row.webhook_id,
+                status_code=result.status_code,
+                duration_ms=duration_ms,
+                attempt=row.attempt_count + 1,
+                is_test=row.is_test,
```
Zingzy, line 185: **Tier 1.** `is_test` is used only for logging — nothing guards the health counters or the reschedule below. So: a transient-failed **test** send hits `record_attempt_and_reschedule`, sets `next_attempt_at`, and the row becomes claimable → a test silently becomes a 10-hour background delivery; exhausted test sends call `record_exhausted`, so testing a broken endpoint ~10× **auto-disables it**; a passing test calls `record_success`, resetting a real failure streak. Test and manual-retry deliveries should be single-shot and must not touch `consecutive_failures`/`total_*` or the retry ladder.

infrastructure/safe_fetch.py
```diff
@@ -228,3 +228,74 @@ async def fetch_public_image(
...
+
+    No redirects are followed — a redirect defeats IP pinning, so it is
+    reported as an error outcome. Never raises: webhook delivery failure
+    is a recorded fact, not an exception path.
+    """
+    try:
+        parsed = httpx.URL(url)
+        if parsed.scheme != "https":
+            return PostResult(None, "non-https URL", None)
+        ip = await _resolve_public_ip(parsed.host)
```
Zingzy, line 285: The guard is genuinely tight — I couldn't get an internal-SSRF path through (https-only + pin-to-IP + no-redirects blocks loopback/RFC1918/link-local/metadata/userinfo/DNS-rebind/redirect-to-Redis). Two gaps: (1) `64:ff9b::/96` (NAT64) returns `is_global=True`, so `https://[64:ff9b::7f00:1]/` passes here and on a NAT64/DNS64 host reaches an internal HTTPS service — reject the prefix explicitly. (2) `post_public` has **zero** direct tests — both test files patch it out, so nothing pins http-rejection / private-IP-rejection / no-redirect at the delivery layer. A refactor silently regresses SSRF; add the tests.

services/webhooks/executor.py
```diff
@@ -0,0 +1,258 @@
...
+        """Fresh timestamp per attempt (replay protection); dual signatures
+        during a rotation grace window, space-delimited per the spec."""
+        ts = int(time.time())
+        secret = decrypt_secret(
+            endpoint.signing_secret_enc, self._master_secret, domain=SECRET_ENC_DOMAIN
+        )
+        signatures = [sign(webhook_id, ts, body, secret)]
+        if endpoint.previous_secret_enc and (
+            endpoint.previous_secret_expires_at
+            and endpoint.previous_secret_expires_at > datetime.now(timezone.utc)
```
Zingzy: Nit / latent: `previous_secret_expires_at` is read from Mongo (naive, the client isn't tz_aware) and compared against an aware `now()` → `TypeError`, the loop-poisoning kind. It's unreachable today (no route writes `previous_secret_enc`), but the moment Phase 2 wires rotation this poisons every delivery to a rotating endpoint for the whole grace window. Use `as_aware_utc()` (already in the repo) now while it's a free fix. Relatedly, `signing.py:6` claims secrets are 'stored hash-only' — they're AES-GCM encrypted; fix that docstring and drop the dead `hash_signing_secret`.

services/webhooks/executor.py
Zingzy: The header build is now guarded: a decrypt failure terminates the row (secret_unreadable) and disables the endpoint with its own reason (SECRET_UNREADABLE), so a SECRET_KEY rotation degrades loudly and recoverably instead of livelocking the claim loop. Pinned by test_unreadable_secret_terminates_and_disables. Fixed in ed562e6.

services/webhooks/dispatcher.py
Zingzy: Wired end to end in ed562e6: dispatch snapshots endpoint.dropped_count onto each new row, the renderer already surfaces it, and record_success drains the counter. Consecutive rows can repeat a count until one lands, which is the right at-least-once semantic for a reconcile signal. The hand-built 42 test is now a state the pipeline produces, and a dispatcher test pins the carry.

services/webhooks/executor.py
Zingzy: Single-shot semantics in ed562e6: is_test rows and manual retries (status not pending) record exactly one attempt and finish. No reschedule, no record_exhausted, no record_success, no disable, 410 included. Three tests pin it: failing test sends stop dead, passing ones do not reset a streak, retried rows never enter the ladder.

services/webhooks/executor.py
Zingzy: Both fixed in ed562e6: the grace comparison now goes through as_aware_utc so Phase 2 rotation cannot poison deliveries, the signing docstring states AES-GCM storage, and the dead hash_signing_secret is gone.

infrastructure/safe_fetch.py
Zingzy: Both closed in ed562e6: the 64:ff9b::/96 prefix is rejected outright in _is_public (a translation literal is never a legitimate delivery target, so no unwrap-and-recheck), and post_public now has direct tests in test_safe_fetch_post.py covering the address policy table (loopback, RFC1918, link-local metadata, CGNAT, ULA, mapped v4, NAT64, multicast), http rejection with zero network, private and transient DNS resolutions as recorded outcomes, and refusal to follow redirects.

Zingzy:
Review batch addressed in ed562e6. Disposition per finding:

**Fixed**

- Startup now raises when `WEBHOOKS_ENABLED=true` with an empty `SECRET_KEY` (hard fail rather than the suggested warning: a predictable encryption key is not a degraded mode worth booting into). Test added.
- `post_public` converts transient DNS failures (and invalid URLs) into delivery outcomes, so they land in the retry ladder as recorded attempts instead of escaping as exceptions.
- TTL index recreation guards the concurrent `drop_index` race (code 27 tolerated), matching the existing `alias_1` guard.
- `mark_failed` increments `attempt_count` alongside the appended attempt.
- Read endpoints now require a verified email, matching the manage side.
- Test sends and manual retries carry the feature-flag gate, since both initiate outbound calls. List/get/patch/delete deliberately stay ungated: a user who loses the flag keeps managing and pausing what they already created, and creation was always gated. The module docstring now states this split.
- `UpdateWebhookEndpointRequest.description` enforces the 256-char bound.
- The queue-Redis-present + `CLICK_EVENTS_SINK=inline` combination no longer leaves click webhooks silent: the fanout wrapper now keys on the click sink being inline rather than on the domain sink, so that topology works instead of warning.
- Bounded the embedded executor shutdown wait (10s).
- Paused endpoints now defer pending deliveries (recheck in 5 minutes, no attempt recorded) instead of terminally failing them; disabled and deleted endpoints stay terminal. Tests updated.

**Not changed, with reasons**

- `ref_policy` on `xadd`: the lockfile pins redis-py 7.4.1, which supports the kwarg (verified against the installed signature), and the same call has been shipping in the click pipeline. The 6.2.0 premise does not match this repository.
- Pending-cap atomicity: count-then-insert can overshoot by one concurrent batch, and that is acceptable for a protective backstop. An exact cap would add a reservation counter write to every dispatch on the hot consumer path; the cap bounds pathology, it is not an SLA. Now documented at the call site.
- Exception containment in the stream consumers: deliberate. An escaping exception leaves the message pending, which is the designed at-least-once path: the claimer retries it and the DLQ guard bounds poison. Swallowing a transient Mongo error there would silently lose delivery intent. The fanout sink swallows because the redirect path must never fail; the consumers must not, because redelivery is exactly what we want.

repositories/webhook_delivery_repository.py
```diff
@@ -0,0 +1,143 @@
...
+
+    async def find_owned(
+        self, delivery_id: ObjectId, user_id: ObjectId
+    ) -> WebhookDeliveryDoc | None:
+        return await self._find_one({"_id": delivery_id, "user_id": user_id})
+
+    async def count_pending(self, endpoint_id: ObjectId) -> int:
+        return await self._count(
+            {"endpoint_id": endpoint_id, "status": DeliveryStatus.PENDING.value}
+        )
```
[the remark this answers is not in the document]
Zingzy: Keeping count-then-insert: the cap is a protective backstop, not an SLA, and worst-case overshoot is one concurrent batch. An exact cap would add a reservation write to every dispatch plus release bookkeeping on every completion, on the click-consumer path. Trade-off documented at the call site.

services/webhooks/consumers.py
```diff
@@ -0,0 +1,98 @@
...
+    async def consume(self, payload: Any) -> None:
+        click = click_event_from_payload(payload)
+        if click is None:
+            return  # malformed — logged by the decoder, drop
+        event = build_link_clicked(
+            click, self._system_default_domain, geoip=self._geoip
+        )
+        if event is None:
+            return  # unowned link — no possible subscriber
+        await self._dispatcher.dispatch(event)
```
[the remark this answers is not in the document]
Zingzy: Intentional: an exception escaping the handler leaves the message unacked, so the claimer retries it and the DLQ bounds poison. That is the at-least-once path. Catching a transient Mongo error here would ack with no delivery rows written, silently losing intent. The fanout sink swallows only because the redirect must never fail.

Zingzy (commented):
Re-reviewed ed562e6 against the head I first reviewed (dcb21b3) — I read the delta and re-verified each finding at the choke point rather than trusting the changelog, and ran the affected suite (123 pass, incl. the new direct `post_public` SSRF tests and config-guard tests). **All three Tier-1 items are genuinely closed, and several Tier-2/nit items came with them.** My original findings stand as having been real; nothing I flagged turned out to be a false alarm, and every fix lands where the defect was.

**Tier 1 — verified fixed:**
- **Decrypt crash-loop:** `_headers` is now wrapped (`executor.py`); a decrypt failure marks the delivery `secret_unreadable` and disables the endpoint with the new `SECRET_UNREADABLE` reason (single-shot rows skip the disable). The infinite reclaim livelock is dead — a `SECRET_KEY` rotation now degrades loudly and recoverably. Confirmed the exception can no longer escape before an attempt is recorded.
- **`dropped_since_last`:** `make_delivery_row` now snapshots `endpoint.dropped_count` onto the row, the renderer surfaces it, `record_success` drains it. The drop signal reaches the subscriber. The at-least-once repeat-until-landed behavior you noted in the comment is the right tradeoff (consumers dedup on `webhook-id`, the number is informational).
- **Test/retry single-shot:** `single_shot = is_test or status != PENDING` cleanly excludes them from `record_success`/`record_exhausted`/the retry ladder and 410-disable. A failing test can no longer auto-disable a real endpoint and a passing one can no longer mask a failure streak. Bonus: PAUSED now `defer`s (5-min recheck) instead of terminally failing the queued delivery — that also closes the pause-destroys-deliveries semantic.

**Also closed:** NAT64 `64:ff9b::/96` rejected outright in `_is_public` (rejecting the whole translation prefix is cleaner than unwrap-and-recheck — agreed); `post_public` now catches `FetchTransientError`/`InvalidURL` → returns data not an exception, so a slow DNS name can't leave a row leased; config hard-fails `WEBHOOKS_ENABLED` without `SECRET_KEY`; the grace-window comparison goes through `as_aware_utc` so Phase-2 rotation can't poison; signing docstring corrected to AES-GCM and dead `hash_signing_secret` removed; `mark_failed` now `$inc`s `attempt_count`; reads require verified email. And flag-gating `test`+`retry` is the right call — it means flipping the flag off now stops the on-demand outbound amplifier, which was the sharp half of the abuse concern; keeping passive management ungated is a coherent doctrine.

**The two you skipped with reasons — I agree with both:** the pending-cap count-then-insert is a protective backstop, not an SLA, and a one-batch overshoot doesn't justify a reservation-counter write on every dispatch. And letting a dispatch exception escape the stream consumer is correct — it leaves the message unacked for the claimer/DLQ at-least-once path; the asymmetry with the inline sink (which swallows, because there's no message to leave pending) is right, not an oversight.

**Merge:** clear. Nothing outstanding blocks it, and it's still dark by default.

**Before GA (the flag-enable tail, all explicitly deferred and fine to carry):** wire the on-disabled email (→ the notifications framework) so endpoints don't die silently; add a per-user delivery-rate cap (the endpoint-count cap doesn't bound the reflector surface once `link.clicked` subscribers exist); and decide whether `/test` should keep echoing the 256-byte response body (it's a content-peek of any public HTTPS URL from spoo's egress, now gated behind the flag + 10/min, so bounded — but worth a conscious call at GA). None of these gate the merge or the first dogfood.

Good, fast turnaround — the fixes are precise and the reasoning on the two declines is sound.

PR #272, opened 2026-07-24
feat(webhooks): discord and slack delivery flavors

Zingzy:
Stacked on #270. Adds the two prebuilt flavors so an endpoint can point straight at a Discord or Slack incoming webhook, no middleware in between.

## What's in here

**Flavors**
- `WebhookFlavor` gains `discord` and `slack`; endpoint create/update accepts them and the signature keeps covering the rendered body, so verification works the same for every flavor.
- A shared copy layer (`renderers/copy.py`) decides wording and which fields earn space, so the two flavors never drift apart. Density follows event frequency: `link.clicked` renders as a few compact lines that read well as a stream, while the rare lifecycle events get field grids.
- `link.updated` on Discord renders the change set as a `diff` code block, so old/new values come out red/green. Slack code blocks don't colorize, so it gets `old → new` pairs instead.
- Receiver limits are enforced locally (Discord: title 256, description 4096, 6000 across the embed; Slack: 3000 per section, fields chunked at 10) with values fenced so they can't break out of code blocks.
- Unknown event types degrade to a JSON code block instead of failing, so adding a new event later can never break a flavored endpoint. `webhook.test` renders its message in both flavors.

**429 handling**
- `post_public` now surfaces an integer `Retry-After` when the receiver sends one.
- On 429 the executor defers the row (Retry-After honored, capped at 15 minutes, 60s fallback) instead of burning retry-ladder attempts or counting toward the auto-disable streak. Discord rate limiting is flow control, not endpoint failure, and popularity should never get an endpoint disabled.
- Test sends and manual retries still report a terminal outcome synchronously.

## Tests

35 new: full-shape pins per flavor per event, receiver size limits under pathological inputs, code-block escape safety, Slack mrkdwn escaping, the unknown-type fallback, a registry/enum drift guard (every flavor an endpoint can be created with must have a renderer), Retry-After parsing, and the 429 defer semantics including the single-shot exception. Full suite: 2879 green.

### DIFF
--- a/routes/api_v1/webhooks.py
+++ b/routes/api_v1/webhooks.py
@@ -30,6 +30,7 @@ from dependencies.auth import (
     WEBHOOKS_MANAGE_SCOPES,
     WEBHOOKS_READ_SCOPES,
     CurrentUser,
+    JwtVerifiedUser,
     require_scopes_verified,
 )
 from errors import ValidationError
@@ -184,11 +185,14 @@ async def get_endpoint(
 async def reveal_secret(
     request: Request,
     endpoint_id: Annotated[str, Path()],
-    user: ManageUser,
+    user: JwtVerifiedUser,
     webhook_service: WebhookSvc,
 ) -> WebhookSecretResponse:
     """The full signing secret, for the endpoint owner. Secrets are stored
-    encrypted and readable on demand; treat the response like a password."""
+    encrypted and readable on demand; treat the response like a password.
+
+    **Authentication**: Interactive session only — API keys are refused,
+    like every operation that exposes or mints credential material."""
     secret = await webhook_service.reveal_secret(
         _oid(endpoint_id, "endpoint"), user.user_id
     )
--- a/routes/api_v1/webhooks.py
+++ b/routes/api_v1/webhooks.py
@@ -48,6 +48,7 @@ from schemas.dto.responses.webhook import (
     WebhookEndpointCreatedResponse,
     WebhookEndpointResponse,
     WebhookEndpointsListResponse,
+    WebhookSecretResponse,
 )
 from schemas.enums.webhook import DeliveryStatus
 from services.feature_flag_service import WEBHOOKS_FLAG
@@ -171,6 +172,27 @@ async def get_endpoint(
         _oid(endpoint_id, "endpoint"), user.user_id
     )
     return WebhookEndpointResponse.from_doc(doc)
+
+
+@router.get(
+    "/webhooks/{endpoint_id}/secret",
+    responses={**AUTH_RESPONSES, **ERROR_RESPONSES},
+    operation_id="revealWebhookSecret",
+    summary="Reveal Signing Secret",
+)
+@limiter.limit(Limits.WEBHOOK_READ)
+async def reveal_secret(
+    request: Request,
+    endpoint_id: Annotated[str, Path()],
+    user: ManageUser,
+    webhook_service: WebhookSvc,
+) -> WebhookSecretResponse:
+    """The full signing secret, for the endpoint owner. Secrets are stored
+    encrypted and readable on demand; treat the response like a password."""
+    secret = await webhook_service.reveal_secret(
+        _oid(endpoint_id, "endpoint"), user.user_id
+    )
+    return WebhookSecretResponse(signing_secret=secret)
 
 
 @router.patch(
--- a/services/webhooks/renderers/copy.py
+++ b/services/webhooks/renderers/copy.py
@@ -97,7 +97,7 @@ def _title(event_type: str, link: dict[str, Any]) -> str:
     label = EVENT_LABELS.get(event_type, event_type)
     alias, domain = link.get("alias"), link.get("domain")
     if _present(alias) and _present(domain):
-        return f"{label} · {domain}/{alias}"
+        return f"{label}: {domain}/{alias}"
     return label
 
 
@@ -128,29 +128,40 @@ def _age_days(created_at: Any, occurred_at: str) -> int | None:
         return None
 
 
+def _country_display(code: Any) -> str:
+    """ISO alpha-2 → flag emoji + code; anything else passes through."""
+    text = str(code)
+    if len(text) == 2 and text.isascii() and text.isalpha():
+        upper = text.upper()
+        flag = "".join(chr(0x1F1E6 + ord(c) - 65) for c in upper)
+        return f"{flag} {upper}"
+    return text
+
+
 def _clicked_pairs(payload: dict[str, Any]) -> list[tuple[str, str]]:
-    """A click always renders at least Device and From — absent analytics
-    dimensions are skipped, but a referrer-less click IS a direct visit
-    and renders as one, never as an empty card."""
+    """One field per dimension — the airy grid IS the point. Absent
+    analytics dimensions are skipped, but a referrer-less click IS a
+    direct visit and the running count is always stated, so a card is
+    never empty."""
     pairs: list[tuple[str, str]] = []
 
-    place = [
-        str(v) for v in (payload.get("city"), payload.get("country")) if _present(v)
-    ]
-    if place:
-        pairs.append(("Location", ", ".join(place)))
-
-    agent = " · ".join(
-        str(payload[k]) for k in ("browser", "os", "device") if _present(payload.get(k))
-    )
-    pairs.append(("Device", agent or "not detected"))
+    if _present(payload.get("country")):
+        pairs.append(("Country", _country_display(payload["country"])))
+    if _present(payload.get("city")):
+        pairs.append(("City", str(payload["city"])))
+    if _present(payload.get("browser")):
+        pairs.append(("Browser", str(payload["browser"])))
+    if _present(payload.get("os")):
+        pairs.append(("OS", str(payload["os"])))
+    if _present(payload.get("device")):
+        pairs.append(("Device", str(payload["device"])))
 
     referrer = payload.get("referrer")
     pairs.append(("From", _referrer_host(referrer) if _present(referrer) else "direct"))
 
     utm = payload.get("utm")
     if isinstance(utm, dict):
-        campaign = " · ".join(str(v) for v in utm.values() if _present(v))
+        campaign = " / ".join(str(v) for v in utm.values() if _present(v))
         if campaign:
             pairs.append(("UTM", campaign))
 
@@ -159,8 +170,9 @@ def _clicked_pairs(payload: dict[str, Any]) -> list[tuple[str, str]]:
         pairs.append(("Bot", str(bot) if _present(bot) else "detected"))
 
     total = payload.get("total_clicks")
-    if isinstance(total, int) and total > 0:
-        pairs.append(("Clicks", f"{total:,}"))
+    if isinstance(total, int):
+        # Count at click time — the payload snapshot precedes the increment.
+        pairs.append(("Clicks so far", f"{total:,}"))
 
     return [(name, clip(value, _VALUE_MAX)) for name, value in pairs]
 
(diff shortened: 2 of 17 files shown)


services/webhooks/renderers/discord.py
```diff
@@ -0,0 +1,106 @@
...
+                diff_lines.append(f"+ {name}: {_fence(new)}")
+            embed["description"] = clip(
+                "```diff\n" + "\n".join(diff_lines) + "\n```", _DESCRIPTION_MAX
+            )
+        elif copy.code:
+            embed["description"] = clip(
+                "```json\n" + _fence(copy.code) + "\n```", _DESCRIPTION_MAX
+            )
+        elif copy.lines:
+            embed["description"] = clip("\n".join(copy.lines), _DESCRIPTION_MAX)
```
Zingzy: Should-fix (security). This joins `copy.lines` into the embed description with no markdown escaping — only code blocks get `_fence`d. For `link.clicked`, the lines include the visitor-controlled `referrer` (and `_referrer_host` falls back to the raw string when it isn't URL-shaped). Verified: `Referer: [x](https://evil.phish)` → description `from [x](https://evil.phish)`, which is Discord masked-link syntax and renders clickable. So a click on a subscribed owner's link plants an attacker-chosen link into the owner's Discord channel under the spoo.me embed. Slack escapes (`_escape`) and `[]()` isn't its syntax, so this is Discord-only — the fix is to escape Discord markdown on user-derived values or wrap them in inline code, matching Slack's posture. `browser`/`os`/`city` are the same class; `pairs` values (long_url) are owner-controlled so lower risk but same fix.

services/webhooks/renderers/copy.py
```diff
@@ -0,0 +1,205 @@
...
+    if isinstance(dropped, int) and dropped > 0:
+        plural = "delivery" if dropped == 1 else "deliveries"
+        return f"{dropped} earlier {plural} dropped"
+    return None
+
+
+def _referrer_host(referrer: Any) -> str:
+    text = str(referrer)
+    host = urlparse(text).netloc
+    return host or text
```
Zingzy, line 121: This is the raw-text fallback that feeds the Discord injection: a referrer that doesn't parse to a netloc (any non-URL string an attacker sends in `Referer:`) returns verbatim, then rides into the Discord description unescaped. Host-extraction isn't a sanitizer here — the escaping has to happen at the Discord render layer (see the discord.py comment).

Zingzy (commented):
Mergeable, and the shape is right — one should-fix security asymmetry before the Discord flavor ships to anyone, plus a couple of small notes. Base is `feat/webhooks-system`, so only formatting ran in CI; I ran the renderer/executor/safe_fetch/integration suite locally (86 pass) and probed the injection path directly.

## Architecture read

The shared `copy.py` layer deciding wording + which fields earn space, with the two flavors as thin presentation over it, is the correct factoring — it's why `link.clicked` reads as a compact stream while lifecycle events get field grids, without the two flavors drifting. The frequency-aware density is a genuinely nice touch. The unknown-event-type → JSON code-block fallback is the right call: it means a future registry addition can never terminal-fail a flavored endpoint, and the enum↔renderer drift-guard test pins that a flavor you can create with always has a renderer. The 429-as-flow-control handling (defer, honor Retry-After capped at 15m, don't burn the ladder or the auto-disable streak, single-shot stays terminal) is exactly right for Discord/Slack receivers — popularity should never disable an endpoint. Signing still covers the rendered body for every flavor; it's dead weight for Discord/Slack (those receivers don't verify our headers) but harmless and consistent.

## Should-fix — Discord renders visitor-controlled `referrer` as a live masked link (content injection into the subscriber's channel)

The Discord embed description is built from `copy.lines` joined verbatim — only code blocks get `_fence`d, nothing escapes inline markdown. `link.clicked`'s lines include `from {referrer_host}`, and `_referrer_host` falls back to the **raw** referrer string whenever it doesn't parse as a URL (`urlparse(...).netloc or text`). I verified it end to end:

```
Referer: [Free-Nitro](https://evil.phish)
→ Discord embed description: "from [Free-Nitro](https://evil.phish) · click #1"
```

`[text](url)` is exactly Discord's masked-link syntax in embed descriptions, so that renders as a clickable link. Concretely: an attacker who clicks any link a Discord-subscribed owner owns, with a crafted `Referer`, plants a clickable link of their choosing into the owner's Discord channel — under the `username: "spoo.me"` embed. `@everyone` won't ping (embeds don't, and there's no `content`/`allowed_mentions`), but a masked phishing link absolutely renders. Given spoo's phishing-flag history, spoo.me-branded embeds carrying attacker links into subscribers' channels is precisely the reputational surface to avoid — and `link.clicked → Discord` is the marquee use case for this PR.

The asymmetry is the tell: Slack got `_escape` (neutralizes `<>&`, and `[]()` isn't Slack link syntax anyway, so it's safe on both counts), but Discord only got `_fence` for code-block breakout and nothing for inline markdown in the description/field values. **Fix:** escape Discord markdown on user-derived values — or wrap them in inline code (backticks) which disables markdown inside — matching the posture Slack already has. `referrer` is the confirmed cross-actor vector; `browser`/`os` (ua-parser output) and `city`/`country` (CF-set, spoofable only on a self-host-without-CF deployment) are the same class and the same fix covers them. The lifecycle `pairs` values (e.g. `Destination` = long_url) are owner-controlled → own-channel only, so lower priority, but the same escape closes them too.

## Notes (not blocking)

- **429 defer never terminates.** A receiver that 429s forever leaves the row PENDING, re-deferred every ≤15m until the 30-day delivery-log TTL reaps it — and deferred rows still count toward `count_pending`, so a chronically rate-limited Discord endpoint eventually fills its pending cap and new events drop. The comment acknowledges TTL as the backstop; it's defensible (real Discord 429s clear in seconds), just noting the deferred row records no attempt, so the delivery log shows it silently parked. A max-defer-count → exhaust would bound it, but not worth it now.
- **`_embed_size` trim loop** pops all fields before touching the description; for a pathological lifecycle event that's mostly description this is fine, but worth knowing the description clip is the last resort and can cut mid-word (the `…` handles it). Bounded and correct.

Nothing else stood out — receiver size limits are enforced locally under pathological inputs (tested), the fallback is safe, and the Retry-After parser correctly ignores the HTTP-date form and negatives. Close the Discord escaping and this is good to go behind the flag.

routes/api_v1/webhooks.py
```diff
@@ -173,6 +174,27 @@ async def get_endpoint(
...
+    "/webhooks/{endpoint_id}/secret",
+    responses={**AUTH_RESPONSES, **ERROR_RESPONSES},
+    operation_id="revealWebhookSecret",
+    summary="Reveal Signing Secret",
+)
+@limiter.limit(Limits.WEBHOOK_READ)
+async def reveal_secret(
+    request: Request,
+    endpoint_id: Annotated[str, Path()],
+    user: ManageUser,
```
Zingzy: Should-fix (security). `ManageUser` is `require_scopes_verified(WEBHOOKS_MANAGE_SCOPES)`, which a `webhooks:manage` API key satisfies (ALLOWED_SCOPES mints everything but keys:manage). So a leaked/over-scoped API key can reveal the signing secret and forge validly-signed events to the owner's existing consumer — worse than the data-exfil a leaked key already allowed. This is the forward-looking constraint from the #270 review now made real. The codebase's own doctrine: key creation is `JwtVerifiedUser` (rejects API keys) and keys:manage isn't mintable, precisely because minting/exposing credential material must be interactive-only. Gate reveal on `require_jwt` too. (Separately: always-on reveal walks back the shown-once posture — a wider disclosure window than create-time-only; fine as a DX call, but it's why interactive-only gating matters here.)

services/webhooks/renderers/copy.py
```diff
@@ -0,0 +1,341 @@
...
+    ]
+    device = " on ".join(device_bits)
+    if _present(payload.get("device")):
+        device = f"{device}, {payload['device']}" if device else str(payload["device"])
+    if device:
+        lines.append(device)
+
+    referrer = payload.get("referrer")
+    source = (
+        f"from **{_referrer_host(referrer)}**" if _present(referrer) else "direct visit"
```
Zingzy: Still unescaped after the V2 rewrite. `_referrer_host(referrer)` returns the raw referrer for any non-URL string, and it lands in a text display as `from **[x](https://evil.phish)**` — verified at head. `allowed_mentions:{parse:[]}` closed the mention-ping vector (good, and necessary for V2 text displays), but not masked links or arbitrary markdown. Escape it here (or inline-code it at the Discord layer) the way Slack's `_escape` already does; referrer is the confirmed visitor-controlled vector into the subscriber's channel.

Zingzy (commented):
Re-reviewed 869b660 against the head I first reviewed (096d163) — read the full delta and re-verified at the choke point, ran the suite (104 pass). This push does more than fix the escaping: Discord moved to Components V2 and a **new secret-reveal endpoint** appeared. Net: my original finding is half-fixed, and the reveal endpoint introduces a new should-fix. Correcting my own verdict on the escaping below, since it's not fully closed.

## Original finding (Discord injection) — half fixed

The V2 rewrite added `allowed_mentions: {"parse": []}`, which **correctly closes the mention-ping vector** — and that addition was necessary, because V2 text displays (type 10) *can* ping where embeds never could, so the rewrite would otherwise have *opened* an @everyone/@here ping vector. Good catch to hard-disable it.

But the **masked-link / markdown half is still open**: the visitor-controlled `referrer` still reaches a text display unescaped. I verified at head:

```
Referer: [Free-Nitro](https://evil.phish)
→ text display: 'from **[Free-Nitro](https://evil.phish)**'
```

`clicked_summary` (`copy.py:270`) embeds `_referrer_host(referrer)` raw, and `_referrer_host` still falls back to the raw string for any non-URL referrer. Text displays render markdown including masked links, so this plants an attacker-chosen clickable link into the subscriber's channel; even if a given client didn't render the masked link, arbitrary markdown (headings, bold, small-text `-#`) still injects and can deface the card. `allowed_mentions` does not touch either. Slack still escapes via `_escape`; Discord still doesn't. Same fix as before: escape Discord markdown on user-derived values, or wrap them in inline code. `referrer` is the confirmed cross-actor vector.

## New should-fix — secret-reveal endpoint is reachable by an API key

`GET /webhooks/{id}/secret` returns the full signing secret and is gated on `ManageUser` = `require_scopes_verified(WEBHOOKS_MANAGE_SCOPES)`, which a `webhooks:manage` **API key** satisfies (`ALLOWED_SCOPES` mints everything except `keys:manage`). This is the exact escalation I flagged as a forward-looking constraint on #270: a leaked/over-scoped API key can now pull the signing secret and **forge validly-signed events to the owner's existing consumer** — an integrity attack on their Zapier/server pipeline, strictly worse than the data-exfil a leaked key already allowed (spin up a new endpoint on the attacker's URL).

The codebase already has the doctrine for this: minting/managing credential material is interactive-only — key *creation* uses `JwtVerifiedUser` (rejects API keys), and `keys:manage` is deliberately not even mintable ("a key that can mint keys defeats revocation"). Revealing a long-lived signing secret is the same class. **Fix:** gate reveal on `require_jwt` (interactive session), like key creation — not a scope an API key can carry.

Second, a product note on the reveal itself: it walks back the "secret shown once at creation" posture (the Stripe/Svix model this system otherwise follows). The secret is now perpetually retrievable by anything with manage access, a materially wider disclosure window than create-time-only. That's a defensible DX call (people lose secrets), but it's the reason the gating above matters more now — if reveal exists, it should be the interactive-only path.

## The rest of the delta — verified fine

- **`_delivery_url` appending `?with_components=true`:** SSRF-safe — host/scheme unchanged, `post_public` re-resolves and pins the same host; query-append handles an existing `?` correctly.
- **`as_aware_utc(event.occurred_at).isoformat()`:** correct fix — Mongo hands back naive datetimes, and the wire timestamp (and Discord's `<t:epoch>`) needs the offset.
- **`increment_deliveries` at enqueue** (counting deliveries when queued, not at terminal): reasonable — "0 of 2 mid-ladder" beats "0 of 0" contradicting the log. One note tying into the parked perf work: it's a per-row awaited `$inc` in a loop after `insert_many`, so it adds one more Mongo write per delivery per event on the dispatch path — worth folding into the same O(1)-counter/batch pass you already parked, since `link.clicked` fan-out is exactly where it compounds.
- Receiver-limit trimming reworked for V2 (clip the largest block) — bounded, tested under pathological input.

Close the referrer escaping and the reveal gating and this is good behind the flag. Everything else here is sound.

services/webhooks/renderers/copy.py
Zingzy: Fixed in 844ae5c. Payload values are backslash-escaped before they reach any text display: clicked_summary takes an escape hook so visitor values are neutralized before the copy layer adds its own markup, and masked link targets encode parentheses so a short URL cannot close the link early. Regression tests cover the masked-link referrer, styled dimension values, and the paren case.

services/webhooks/renderers/copy.py
Zingzy: Addressed in 844ae5c exactly as suggested: host extraction stays a display helper and the escaping happens at the render layer, applied to every payload-derived value including this fallback path.

services/webhooks/renderers/discord.py
Zingzy: The embed path this pointed at was replaced by the components rewrite, and the same class of fix landed in 844ae5c: all user-derived values are markdown-escaped at the discord layer, matching the Slack posture.

routes/api_v1/webhooks.py
Zingzy: Fixed in 844ae5c. Reveal now requires an interactive session via the same dependency key creation uses, so API keys and delegated app tokens are refused outright. Added an integration test asserting a webhooks:manage key gets a 403.

PR #273, opened 2026-07-24
feat: send rate limit headers on limited responses

Zingzy:
Adds X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, and Retry-After to every rate-limited response, including 429s.

## Why not slowapi's built-ins

- The decorator-side injection raises unless the endpoint returns a Response object or declares a `response` parameter, which would mean touching every route signature.
- The bundled SlowAPIASGIMiddleware exempts decorated routes entirely (it only enforces default limits), so it never injects headers for our routes.

Instead a small ASGI middleware reads the window state the decorator already records (`request.state.view_rate_limit`) and injects headers at response start. A Limiter subclass makes the decorator's own injection a no-op when there is no Response to mutate.

## Tests

- Two new integration tests: headers on a 200, headers + Retry-After on a 429.
- Full suite: 941 passed, 1 failed. The failure (`test_app_token_scopes.py::TestScopeEnforcement::test_in_scope_route_200`) also fails on current main without this change, so it is pre-existing and unrelated.

## Summary by CodeRabbit

- **New Features**
  - Added/standardized `X-RateLimit-*` headers on applicable HTTP responses, including health endpoints.
  - `X-RateLimit-Reset` is now returned as numeric epoch seconds; rate-limited (`429`) responses include `Retry-After`.

- **Bug Fixes**
  - Prevented duplicate `X-RateLimit-*` headers for endpoints that return explicit responses.

- **Tests**
  - Expanded integration coverage to verify header presence, correctness, and single-header behavior across main and health endpoints.

### DIFF
--- a/app.py
+++ b/app.py
@@ -276,8 +276,9 @@ def create_app(settings: AppSettings | None = None) -> FastAPI:
     app.add_middleware(TenantMiddleware)
     # 6. Request logging — logs all requests with request_id
     app.add_middleware(RequestLoggingMiddleware)
-    # 7. Rate limit headers — innermost so view_rate_limit is populated by
-    #    the route decorator before the response passes back through it
+    # 7. Rate limit headers — placement is order-independent: the decorator
+    #    writes view_rate_limit into shared scope state during endpoint
+    #    execution, before any response starts flowing outward
     app.add_middleware(RateLimitHeadersMiddleware)
 
     # ── Error handlers + rate limiter ────────────────────────────────────
--- a/app.py
+++ b/app.py
@@ -39,7 +39,7 @@ from middleware.openapi import (
     OPENAPI_TAGS,
     configure_openapi,
 )
-from middleware.rate_limiter import limiter
+from middleware.rate_limiter import RateLimitHeadersMiddleware, limiter
 from middleware.security import (
     MaxContentLengthMiddleware,
     SecurityHeadersMiddleware,
@@ -274,8 +274,11 @@ def create_app(settings: AppSettings | None = None) -> FastAPI:
     )
     # 5. Tenant resolution — populates request.state.tenant from Host
     app.add_middleware(TenantMiddleware)
-    # 6. Request logging — innermost, logs all requests with request_id
+    # 6. Request logging — logs all requests with request_id
     app.add_middleware(RequestLoggingMiddleware)
+    # 7. Rate limit headers — innermost so view_rate_limit is populated by
+    #    the route decorator before the response passes back through it
+    app.add_middleware(RateLimitHeadersMiddleware)
 
     # ── Error handlers + rate limiter ────────────────────────────────────
     app.state.limiter = limiter
--- a/middleware/rate_limiter.py
+++ b/middleware/rate_limiter.py
@@ -206,12 +206,59 @@ def rate_limit_key(request: Request) -> str:
 _redis_uri = os.environ.get("REDIS_URI")
 _storage_uri = _redis_uri if _redis_uri else "memory://"
 
-limiter = Limiter(
+
+class _HeaderSafeLimiter(Limiter):
+    """Limiter whose decorator tolerates endpoints that return plain data.
+
+    With ``headers_enabled`` slowapi's decorator raises unless the endpoint
+    returns a Response or declares a ``response`` parameter. Header injection
+    happens in ``RateLimitHeadersMiddleware`` at the ASGI layer instead, so
+    the decorator-side injection can safely no-op when there is nothing to
+    mutate.
+    """
+
+    def _inject_headers(self, response, current_limit):
+        if response is None:
+            return None
+        return super()._inject_headers(response, current_limit)
+
+
+limiter = _HeaderSafeLimiter(
     key_func=rate_limit_key,
     default_limits=[Limits.DEFAULT_MINUTE, Limits.DEFAULT_HOUR, Limits.DEFAULT_DAY],
     storage_uri=_storage_uri,
     strategy="fixed-window",
+    headers_enabled=True,
 )
+
+
+class RateLimitHeadersMiddleware:
+    """Inject X-RateLimit-* and Retry-After headers at the ASGI layer.
+
+    The rate-limit decorator records the evaluated window on
+    ``request.state.view_rate_limit`` for every limited route. Reading it
+    from the scope state here covers all response shapes, including plain
+    dict returns and 429s produced by the exception handler.
+    """
+
+    def __init__(self, app):
+        self.app = app
+
+    async def __call__(self, scope, receive, send):
+        if scope["type"] != "http":
+            return await self.app(scope, receive, send)
+
+        async def send_wrapper(message):
+            if message["type"] == "http.response.start":
+                view_limit = scope.get("state", {}).get("view_rate_limit")
+                if view_limit is not None:
+                    from starlette.datastructures import MutableHeaders
+
+                    headers = MutableHeaders(raw=message["headers"])
+                    limiter._inject_asgi_headers(headers, view_limit)
+            await send(message)
+
+        await self.app(scope, receive, send_wrapper)
 
 
 # ── Dynamic limits ───────────────────────────────────────────────────────────
--- a/middleware/rate_limiter.py
+++ b/middleware/rate_limiter.py
@@ -8,6 +8,7 @@ Key resolution: API key hash → JWT token hash → client IP.
 from __future__ import annotations
 
 import hashlib
+import math
 import os
 
 from fastapi import Request
@@ -263,7 +264,7 @@ class RateLimitHeadersMiddleware:
                     else:
                         reset = headers.get("X-RateLimit-Reset")
                         if reset is not None:
-                            headers["X-RateLimit-Reset"] = str(int(float(reset)))
+                            headers["X-RateLimit-Reset"] = str(math.ceil(float(reset)))
                         if message["status"] != 429 and not had_retry_after:
                             del headers["Retry-After"]
             await send(message)
--- a/middleware/rate_limiter.py
+++ b/middleware/rate_limiter.py
@@ -12,8 +12,12 @@ import os
 
 from fastapi import Request
 from slowapi import Limiter
+from starlette.datastructures import MutableHeaders
 
+from infrastructure.logging import get_logger
 from shared.ip_utils import get_client_ip
+
+log = get_logger(__name__)
 
 # ── Limits ───────────────────────────────────────────────────────────────────
 
@@ -24,11 +28,6 @@ class Limits:
     Ported from blueprints/limits.py. All values use slowapi's "N per period"
     format. Semicolons combine multiple limits into one decorator.
     """
-
-    # Global defaults (applied to every route unless overridden)
-    DEFAULT_MINUTE = "10 per minute"
-    DEFAULT_HOUR = "100 per hour"
-    DEFAULT_DAY = "500 per day"
 
     # API v1 — authenticated vs anonymous tiers
     API_AUTHED = "60 per minute; 5000 per day"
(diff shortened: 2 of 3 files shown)


middleware/rate_limiter.py
```diff
@@ -206,14 +206,61 @@ def rate_limit_key(request: Request) -> str:
...
+            return await self.app(scope, receive, send)
+
+        async def send_wrapper(message):
+            if message["type"] == "http.response.start":
+                view_limit = scope.get("state", {}).get("view_rate_limit")
+                if view_limit is not None:
+                    from starlette.datastructures import MutableHeaders
+
+                    headers = MutableHeaders(raw=message["headers"])
+                    limiter._inject_asgi_headers(headers, view_limit)
```
Zingzy: Two things on this call.

`_inject_asgi_headers` sets `Retry-After` unconditionally, so every 200 from a limited route now carries a retry delay. The header only means something on a 429 (or 503), and generic HTTP retry layers that key on its presence will back off on successful responses. Keeping the X-RateLimit trio everywhere and the retry hint where it belongs is one line:

```python
limiter._inject_asgi_headers(headers, view_limit)
if message["status"] != 429 and "Retry-After" in headers:
    del headers["Retry-After"]
```

Second: this does a fresh `get_window_stats`, and with `swallow_errors` at its default False a transient Redis failure here raises before `http.response.start` goes out, so a request whose work already completed dies as a 500. Headers are cosmetic; wrap the injection in try/except and log, matching how the other Redis-backed extras degrade.

middleware/rate_limiter.py
```diff
@@ -206,14 +206,61 @@ def rate_limit_key(request: Request) -> str:
...
+    returns a Response or declares a ``response`` parameter. Header injection
+    happens in ``RateLimitHeadersMiddleware`` at the ASGI layer instead, so
+    the decorator-side injection can safely no-op when there is nothing to
+    mutate.
+    """
+
+    def _inject_headers(self, response, current_limit):
+        if response is None:
+            return None
+        return super()._inject_headers(response, current_limit)
```
Zingzy: Consider no-opping for every response shape, not just `None`. Endpoints that return a real `Response` (exports, the oauth redirects) still take this decorator injection path, then the ASGI middleware injects the same headers again: two `get_window_stats` storage reads per request and two writers for the same values, with the middleware silently winning through `MutableHeaders.__setitem__` de-duplication (verified against starlette 1.3.1, no duplicate headers reach the wire). Since the design here is that the middleware is the single injection point, `return response` unconditionally makes that actually true and drops the double read.

middleware/rate_limiter.py
```diff
@@ -206,14 +206,61 @@ def rate_limit_key(request: Request) -> str:
...
+
+    async def __call__(self, scope, receive, send):
+        if scope["type"] != "http":
+            return await self.app(scope, receive, send)
+
+        async def send_wrapper(message):
+            if message["type"] == "http.response.start":
+                view_limit = scope.get("state", {}).get("view_rate_limit")
+                if view_limit is not None:
+                    from starlette.datastructures import MutableHeaders
```
Zingzy: Nit: no import cycle prevents this living at module top, and inside the closure it re-executes per response.

app.py
```diff
@@ -274,8 +274,11 @@ async def docs(request: Request):
     )
     # 5. Tenant resolution — populates request.state.tenant from Host
     app.add_middleware(TenantMiddleware)
-    # 6. Request logging — innermost, logs all requests with request_id
+    # 6. Request logging — logs all requests with request_id
     app.add_middleware(RequestLoggingMiddleware)
+    # 7. Rate limit headers — innermost so view_rate_limit is populated by
```
Zingzy: This comment is inverted: `add_middleware` inserts at index 0 and the stack wraps in reverse, so the last middleware added sees the request first. `RateLimitHeadersMiddleware` is the outermost user middleware here, not the innermost (verified empirically against starlette 1.3.1 with a two-middleware probe). It still works from any layer, because the decorator writes `view_rate_limit` into the shared `scope["state"]` dict during endpoint execution, before any `http.response.start` flows outward, so the ordering dependency this comment documents does not exist. Worth rewording to say the placement is order-independent.

Side note, pre-existing and not this PR's problem: the same inversion runs through the whole numbered block. Session (added first, labeled outermost) is actually innermost, and SplitCORS actually runs outside SecurityHeaders, so preflight 204s skip HSTS/CSP despite the claim on step 2. Deserves a small follow-up sweep.

Zingzy (commented):
No blockers. The approach is right, and the "why not slowapi's built-ins" analysis checks out against the installed slowapi 0.1.10: the raise lives inside `Limiter._inject_headers`, so the subclass intercepts exactly the right seam, and `SlowAPIASGIMiddleware` really does skip decorated routes. I also verified the mechanism end to end: `scope["state"]` is the same dict through the whole BaseHTTPMiddleware stack (probed empirically against starlette 1.3.1), so the middleware sees `view_rate_limit` no matter where it sits, and `MutableHeaders.__setitem__` de-duplicates anything the decorator already appended. The redirect hot path stays untouched (`@limiter.exempt`), and the rate-limiting test file passes 12/12 in a clean worktree.

Three should-fixes, all cheap, plus nits inline.

1. **`Retry-After` now ships on 200s.** `_inject_asgi_headers` sets it unconditionally, so every successful response from a limited route advertises a retry delay. It belongs on 429s; generic client backoff layers that key on the header's presence will throttle for no reason. One line after the injection fixes it.

2. **A storage blip at send time turns a finished response into a 500.** The injection does a fresh `get_window_stats` (a second synchronous storage read), and with `swallow_errors` at its default a transient Redis failure raises inside `send_wrapper` before `http.response.start` goes out. When Redis is hard down the decorator raises first, so the new exposure is only the transient window, but headers are cosmetic and should never kill a completed request: try/except and log, same degradation posture as the Redis-backed caches.

3. **Make the decorator override an unconditional no-op.** Endpoints that return a real `Response` (exports, the oauth redirects) still take the decorator injection path, then the middleware injects again: two `get_window_stats` reads per request and two writers for the same headers, with the middleware silently winning. The PR's own design argument is that the ASGI middleware is the single injection point; `return response` unconditionally makes that true.

**Cost:** each response from a limited route pays one extra synchronous storage read on the event loop, the same blocking pattern slowapi's decorator already uses, and the exempt redirect path pays nothing. Fine at current API volume.

**Merge frame:** nothing gates this, so it ships with the next release. That is fine (the headers are purely additive), and the public API docs should document the new headers when they land.

Zingzy:
Follow-up to my review: I ran the branch against the real stack (uvicorn from a clean worktree on the dockerized Mongo and Redis, isolated db index and database) and checked slowapi's docs and source more carefully. Everything below is observed on the wire, not inferred.

**Confirmed from the review:**
- `Retry-After` ships on success responses: `GET /api/v1/emoji-set` returns `200` with `retry-after: 60`, and `POST /api/v1/shorten` returns `201` with `retry-after: 60`. A created resource telling the client to back off for a minute is exactly the misread generic retry layers will make.
- No duplicate headers reach the wire, even on `emoji-set`, which declares a `response` param and therefore takes the decorator injection path too. The double `get_window_stats` read is still there, but the output is clean.
- The redirect path is clean: `302` with zero rate-limit headers. `/health` and `/docs` clean too.
- The 429 is right: `x-ratelimit-remaining: 0`, `retry-after: 40` counting down to the window edge, headers present on both the JSON and browser Accept variants.

**New findings from the live run:**

1. `X-RateLimit-Reset` is a float epoch: `1785443523.660372`. It also jitters within a single fixed window (observed `.452`, `.461`, `.660` across three responses of the same window, TTL-derived noise). Every mainstream implementation of this header sends integer epoch seconds, and strict client parsers will reject or mangle the float. Worth an `int()` in the middleware while you are already post-processing (same spot as the Retry-After status gate).

2. Multi-window limits only ever report the shortest window. `limits.RateLimitItem.__lt__` compares window length alone and ignores the amount, so slowapi's "pick the smallest limit for the header" always selects the minute window: `"60 per minute; 2000 per day"` reports `limit: 60` on every success, and a client whose day budget is nearly gone sees nothing until the 429 (which does correctly report the failed window). No cheap fix inside slowapi; fine to accept, but the API docs should say the headers describe the shortest window only.

3. `default_limits` is dead config. It only takes effect through `SlowAPIMiddleware`, which this app never registers, and every decorated route keeps slowapi's `override_defaults=True`. Proven live: 15 rapid hits on `/docs` all returned 200 against a `10 per minute` default. Blast radius is near zero because coverage is genuinely excellent (96 routes explicitly limited, 14 exempt, only `/docs` and `/openapi.json` uncovered), but the three `DEFAULT_*` constants and the constructor arg protect nothing and read as if they do. Either drop them or leave a comment saying they are aspirational.

On the docs check: slowapi's documentation does confirm the PR body's rationale (the decorator needs a `Response` return or a `response` param when `headers_enabled` is set). The header-selection rule and the defaults-do-not-stack behavior above are undocumented; both verified from the installed 0.1.10 source.

middleware/rate_limiter.py
Zingzy: Fixed in a789b66. Retry-After is stripped unless the status is 429 (values an endpoint set on other statuses survive), the reset value is cast to integer epoch seconds, and the injection is wrapped in try/except log so a storage blip cannot fail a response whose work already completed. Verified live: 200 and 201 no longer carry Retry-After, the 429 still does.

middleware/rate_limiter.py
Zingzy: Done in a789b66. The override returns the response untouched, so the middleware is the single injection point and the second get_window_stats read on Response-returning endpoints is gone. Added a regression test asserting such endpoints get exactly one header set.

middleware/rate_limiter.py
Zingzy: Moved to module top in a789b66.

app.py
Zingzy: Reworded in a789b66: the comment now says placement is order-independent, since the decorator writes view_rate_limit into shared scope state before any response starts. The rest of the block's ordering comments are left for the follow-up sweep.

middleware/rate_limiter.py
```diff
@@ -206,14 +206,61 @@ def rate_limit_key(request: Request) -> str:
...
+
+                    headers = MutableHeaders(raw=message["headers"])
+                    limiter._inject_asgi_headers(headers, view_limit)
+            await send(message)
```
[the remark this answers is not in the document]
Zingzy: Checked against the installed slowapi (0.1.10, not 0.1.9) and on the wire: _inject_asgi_headers sets Retry-After unconditionally, the existing-header branch only feeds the max() on the reset time. A live 429 on this branch carried retry-after: 59 with nothing seeding it, so no seeding is needed. The real problem was the opposite one, Retry-After leaking onto 200s and 201s, fixed in a789b66 by stripping it from non-429 responses.

tests/integration/test_rate_limiting.py
```diff
@@ -269,3 +275,50 @@ async def _ip_bucket_test(request: Request):
...
+        return {"ok": True}
+
+    app = _build_test_app(extra_routers=[r])
+    with TestClient(app, raise_server_exceptions=False) as c:
+        c.get("/test-headers-429")
+        resp = c.get("/test-headers-429", headers={"Accept": "application/json"})
+        assert resp.status_code == 429
+        assert resp.json()["code"] == "rate_limit_exceeded"
+        assert "Retry-After" in resp.headers
+        assert resp.headers.get("X-RateLimit-Remaining") == "0"
```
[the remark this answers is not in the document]
Zingzy: Added in 176eac8: the 429 test now asserts X-RateLimit-Limit and an integer X-RateLimit-Reset alongside Retry-After and Remaining.

middleware/rate_limiter.py
```diff
@@ -206,14 +205,72 @@ def rate_limit_key(request: Request) -> str:
...
+                    headers = MutableHeaders(raw=message["headers"])
+                    had_retry_after = "Retry-After" in headers
+                    try:
+                        limiter._inject_asgi_headers(headers, view_limit)
+                    except Exception:
+                        log.warning("rate_limit_header_injection_failed", exc_info=True)
+                    else:
+                        reset = headers.get("X-RateLimit-Reset")
+                        if reset is not None:
+                            headers["X-RateLimit-Reset"] = str(int(float(reset)))
```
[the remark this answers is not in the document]
Zingzy: Applied in 4e68aa3. slowapi already pads the reset with a full second before this post-process (reset_in = 1 + window_stats[0]), so truncation could not land before the window edge, but ceil drops the dependency on that internal detail.

tests/integration/test_rate_limiting.py
```diff
@@ -269,3 +276,77 @@ async def _ip_bucket_test(request: Request):
...
+        assert resp.headers["X-RateLimit-Reset"].isdigit()
```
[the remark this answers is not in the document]
Zingzy: Added in 4e68aa3: Retry-After asserted numeric on the 429, and the Response-endpoint test asserts exactly one numeric X-RateLimit-Reset value.

PR #277, opened 2026-08-01
feat: claimable anonymous links via one-time claim tokens

### LINKED ISSUE #133: anonymous users can't reclaim their links

Zingzy:
## What

Links shortened anonymously can now be claimed into an account later. The anonymous shorten response carries a one-time `claim_token`, a bearer proof of creation, and a new authenticated endpoint attaches the link to whoever presents it. Claimed links join the account like any owned link: management, dashboard, and their full click history.

## How it works

**Issuance.** Anonymous `POST /api/v1/shorten` responses gain a `claim_token` field (null on authenticated creates). Only its sha256 hash is stored on the URL document, the same recipe API keys use; the plaintext exists exactly once, in the creation response.

```json
{
  "id": "688c9f...",
  "alias": "abc1234",
  "short_url": "https://spoo.me/abc1234",
  "claim_token": "kJ8vQ2m...one-time...",
  "owner_id": null
}
```

**Claiming.** `POST /api/v1/urls/claim` (auth required, `urls:manage` or `admin:all`) takes up to 16 pairs and resolves each independently; the batch never hard-fails:

```json
{ "claims": [ { "url_id": "688c9f...", "token": "kJ8vQ2m..." } ] }
```
```json
{ "results": [ { "url_id": "688c9f...", "status": "claimed" } ], "claimed": 1 }
```

Statuses are `claimed`, `already_yours` (idempotent repeat), and `invalid`. Unknown ids, foreign-owned links, and wrong tokens all answer `invalid`, so the endpoint cannot be used to probe whether a URL exists or who owns it.

A successful claim is a single compare-and-swap on the URL document: the stored hash and the anonymous owner are part of the filter, so concurrent claims lose cleanly. The update stamps `owner_id` and `claimed_at` and burns the hash (single use). The redirect cache entry is evicted so the next click is attributed to the new owner.

**Stats.** Clicks recorded before a claim keep their anonymous owner stamp, so `scope=all` queries add a `meta.url_id` arm for the account's claimed links. The set comes from a new partial index (`owner_claimed`) that only contains claimed URLs; accounts that never claimed resolve an empty set and run the exact same query as before.

## Blast radius

- `UrlService.create` now returns `(doc, claim_token)`; single caller updated.
- `UrlResponse` gains the nullable `claim_token` field (shorten endpoint only).
- `UrlV2Doc` gains `claim_token_hash` and `claimed_at`, both stripped when unset so the partial index can filter on `$exists`.
- New rate limit tier for the claim endpoint: 30/min, 500/day per credential. Token guessing is not a realistic vector at 256 bits; the limiter is a backstop.
- `claim_token` is masked by the existing log redaction substring rule (test added).
- Legacy `POST /` does not issue tokens; v1-generation links are out of scope.

## Testing

- 26 new unit and integration tests: issuance and hashing, the claim state machine (wrong token, unknown id, idempotent repeat, foreign owner, race loss, mixed batches, pre-feature links), stats query composition including `$and` nesting with the null-sentinel filters, the partial index options, request validation bounds, response contracts, and redaction. Full suite: 2,883 passing.
- Verified end to end against a live app with real Mongo and Redis: register, shorten anonymously, claim (wrong token first, then right, then repeat), confirm ownership and burned hash in the database, and confirm the dashboard counts pre-claim clicks while a second account sees none of it.

## Summary by CodeRabbit

- **New Features**
  - Anonymous shortened URLs now include a one-time claim token.
  - Authenticated users can claim anonymous URLs individually or in batches of up to 16.
  - Claim results identify successfully claimed, already-owned, and invalid URLs.
  - Account claim limits and request rate limits help protect the feature.
  - Statistics now include historical activity from URLs claimed by the account.
- **Security**
  - Claim tokens are treated as sensitive data and are securely protected.

Closes #133.

### DIFF
--- a/repositories/url_repository.py
+++ b/repositories/url_repository.py
@@ -105,20 +105,48 @@ class UrlRepository(BaseRepository[UrlV2Doc]):
             },
         )
 
+    async def count_claimed(self, owner_id: ObjectId) -> int:
+        """How many links the owner has claimed in (``claimed_at`` present).
+
+        Backs the per-account claim ceiling; served by the owner_claimed
+        partial index.
+        """
+        return await self._count(
+            {"owner_id": owner_id, "claimed_at": {"$exists": True}}
+        )
+
     async def list_claimed_ids(
-        self, owner_id: ObjectId, *, limit: int = 128
+        self, owner_id: ObjectId, *, limit: int = 1024
     ) -> list[ObjectId]:
         """ids of the owner's claimed-in links (``claimed_at`` present).
 
         Served by the owner_claimed partial index (holds only claimed
-        links platform-wide — sub-ms). Cap is defensive; real sets are
-        dozens.
+        links platform-wide — sub-ms). The cap is a backstop above the
+        write-side claim ceiling: unreachable by construction, so a full
+        cursor means something bypassed the ceiling and scope=all stats
+        are silently undercounting — exactly what the warning is for.
         """
-        cursor = self._col.find(
-            {"owner_id": owner_id, "claimed_at": {"$exists": True}},
-            {"_id": 1},
-        ).limit(limit)
-        return [d["_id"] async for d in cursor]
+        try:
+            cursor = self._col.find(
+                {"owner_id": owner_id, "claimed_at": {"$exists": True}},
+                {"_id": 1},
+            ).limit(limit)
+            ids = [d["_id"] async for d in cursor]
+        except PyMongoError as exc:
+            log.error(
+                "repo_list_claimed_ids_failed",
+                collection=self._collection_name,
+                error=str(exc),
+                error_type=type(exc).__name__,
+            )
+            raise
+        if len(ids) >= limit:
+            log.warning(
+                "claimed_ids_truncated",
+                owner_id=str(owner_id),
+                limit=limit,
+            )
+        return ids
 
     async def record_meta_image_validation(
         self, url_id: ObjectId, image_url: str, meta: dict
--- a/repositories/url_repository.py
+++ b/repositories/url_repository.py
@@ -16,6 +16,7 @@ from pymongo.errors import DuplicateKeyError, PyMongoError
 
 from infrastructure.logging import get_logger
 from repositories.base import BaseRepository
+from schemas.models.base import ANONYMOUS_OWNER_ID
 from schemas.models.url import UrlStatus, UrlV2Doc
 
 log = get_logger(__name__)
@@ -75,6 +76,49 @@ class UrlRepository(BaseRepository[UrlV2Doc]):
     async def delete(self, url_id: ObjectId) -> bool:
         """Hard-delete a URL document. Returns True if a document was deleted."""
         return await self._delete({"_id": url_id})
+
+    async def claim_by_token_hash(
+        self,
+        url_id: ObjectId,
+        token_hash: str,
+        new_owner_id: ObjectId,
+        claimed_at: datetime,
+    ) -> bool:
+        """CAS-transfer an anonymous URL to *new_owner_id*.
+
+        Hash + anonymous owner live in the filter, so races and burned
+        tokens match nothing. Hash unset on success — single use.
+        """
+        return await self._update(
+            {
+                "_id": url_id,
+                "claim_token_hash": token_hash,
+                "owner_id": ANONYMOUS_OWNER_ID,
+            },
+            {
+                "$set": {
+                    "owner_id": new_owner_id,
+                    "claimed_at": claimed_at,
+                    "updated_at": claimed_at,
+                },
+                "$unset": {"claim_token_hash": ""},
+            },
+        )
+
+    async def list_claimed_ids(
+        self, owner_id: ObjectId, *, limit: int = 128
+    ) -> list[ObjectId]:
+        """ids of the owner's claimed-in links (``claimed_at`` present).
+
+        Served by the owner_claimed partial index (holds only claimed
+        links platform-wide — sub-ms). Cap is defensive; real sets are
+        dozens.
+        """
+        cursor = self._col.find(
+            {"owner_id": owner_id, "claimed_at": {"$exists": True}},
+            {"_id": 1},
+        ).limit(limit)
+        return [d["_id"] async for d in cursor]
 
     async def record_meta_image_validation(
         self, url_id: ObjectId, image_url: str, meta: dict
--- a/routes/api_v1/management.py
+++ b/routes/api_v1/management.py
@@ -70,6 +70,8 @@ async def claim_urls_v1(
     **Authentication**: Required.
 
     **API Key Scope**: `urls:manage` or `admin:all`
+
+    **Rate Limits**: 30/min, 500/day
     """
     results = await url_service.claim(body.claims, user.user_id)
     return ClaimUrlsResponse(
--- a/routes/api_v1/management.py
+++ b/routes/api_v1/management.py
@@ -64,8 +64,9 @@ async def claim_urls_v1(
     repeat), or `invalid` (unknown id, wrong token, or a link that is not
     claimable — deliberately indistinguishable).
 
-    Claimed links join the account like any owned link: dashboard,
-    management, and full stats history included.
+    The claim_token is single-use and is invalidated immediately on
+    success. Claimed links join the account like any owned link:
+    dashboard, management, and full stats history included.
 
     **Authentication**: Required.
 
--- a/routes/api_v1/management.py
+++ b/routes/api_v1/management.py
@@ -25,12 +25,57 @@ from dependencies import (
 from middleware.openapi import AUTH_RESPONSES, ERROR_RESPONSES
 from middleware.rate_limiter import Limits, limiter
 from routes.api_v1._helpers import parse_url_id
-from schemas.dto.requests.url import UpdateUrlRequest, UpdateUrlStatusRequest
-from schemas.dto.responses.url import DeleteUrlResponse, UpdateUrlResponse
+from schemas.dto.requests.url import (
+    ClaimUrlsRequest,
+    UpdateUrlRequest,
+    UpdateUrlStatusRequest,
+)
+from schemas.dto.responses.url import (
+    ClaimResultItem,
+    ClaimUrlsResponse,
+    DeleteUrlResponse,
+    UpdateUrlResponse,
+)
 from services.feature_flag_service import GEO_TARGETING_FLAG, META_TAGS_FLAG
 from shared.ip_utils import get_client_ip
 
 router = APIRouter(tags=["Link Management"])
+
+
+@router.post(
+    "/urls/claim",
+    responses=AUTH_RESPONSES,
+    operation_id="claimUrls",
+    summary="Claim anonymous URLs",
+)
+@limiter.limit(Limits.URL_CLAIM)
+async def claim_urls_v1(
+    request: Request,
+    body: ClaimUrlsRequest,
+    url_service: UrlSvc,
+    user: CurrentUser = Depends(require_scopes(URL_MANAGEMENT_SCOPES)),  # noqa: B008
+) -> ClaimUrlsResponse:
+    """Claim anonymously-created URLs into your account.
+
+    Each item pairs a URL id with the one-time `claim_token` returned by
+    the anonymous shorten call — the bearer proof of creation. Items
+    resolve independently and the batch never hard-fails: `claimed`
+    (ownership transferred, token burned), `already_yours` (idempotent
+    repeat), or `invalid` (unknown id, wrong token, or a link that is not
+    claimable — deliberately indistinguishable).
+
+    Claimed links join the account like any owned link: dashboard,
+    management, and full stats history included.
+
+    **Authentication**: Required.
+
+    **API Key Scope**: `urls:manage` or `admin:all`
+    """
+    results = await url_service.claim(body.claims, user.user_id)
+    return ClaimUrlsResponse(
+        results=[ClaimResultItem(url_id=r.url_id, status=r.status) for r in results],
+        claimed=sum(1 for r in results if r.status == "claimed"),
+    )
 
 
 @router.patch(
--- a/schemas/dto/requests/url.py
+++ b/schemas/dto/requests/url.py
@@ -535,7 +535,9 @@ class ClaimItemRequest(RequestBase):
             "The one-time claim_token returned by the anonymous shorten "
             "call — the bearer proof of creation."
         ),
-        min_length=16,
+        # Issued tokens are 43 chars (32 urlsafe bytes); anything shorter
+        # is garbage worth rejecting before it costs a lookup.
+        min_length=43,
         max_length=256,
     )
 
--- a/schemas/dto/requests/url.py
+++ b/schemas/dto/requests/url.py
@@ -518,3 +518,33 @@ class AliasCheckQuery(RequestBase):
         if v is None or v == "":
             return None
         return normalise_fqdn(v)
+
+
+class ClaimItemRequest(RequestBase):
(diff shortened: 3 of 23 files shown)


repositories/url_repository.py
```diff
@@ -76,6 +77,49 @@ async def delete(self, url_id: ObjectId) -> bool:
...
+                    "owner_id": new_owner_id,
+                    "claimed_at": claimed_at,
+                    "updated_at": claimed_at,
+                },
+                "$unset": {"claim_token_hash": ""},
+            },
+        )
+
+    async def list_claimed_ids(
+        self, owner_id: ObjectId, *, limit: int = 128
```
Zingzy: The docstring says real sets are dozens, but the endpoint budget (16 per request, 500/day) lets an API consumer legitimately exceed 128 claimed links, after which dashboard totals silently drop history with no log and no signal. Suggest `limit: int = 1024` (the `$in` stays cheap at that size) plus a `log.warning` when `len(result) == limit`, so truncation is at least loud.

repositories/url_repository.py
```diff
@@ -76,6 +77,49 @@ async def delete(self, url_id: ObjectId) -> bool:
...
+    ) -> list[ObjectId]:
+        """ids of the owner's claimed-in links (``claimed_at`` present).
+
+        Served by the owner_claimed partial index (holds only claimed
+        links platform-wide — sub-ms). Cap is defensive; real sets are
+        dozens.
+        """
+        cursor = self._col.find(
```
Zingzy: Pattern nit: the sibling `list_aliases_by_owner_and_domain` wraps its cursor in try/except PyMongoError with a context log before propagating. This one goes bare, so a Mongo hiccup here becomes an uncontextualized 500 on every scope=all stats call.

Zingzy (commented):
Mergeable, nothing blocking. One should-fix worth doing before merge, two nits. CI is green on both Python versions and the full suite passes at HEAD.

## What I traced

- **Webhook payload leak**: `link.created` fires at create time with the doc now carrying `claim_token_hash`. Verified `link_snapshot` is an explicit field projection, so the hash never reaches webhook receivers. Clean.
- **Existence-oracle resistance**: unknown id, foreign owner, wrong token, and pre-feature links all collapse to `invalid`, and the scope=all short_code dimension guard still composes conjunctively with the new ownership `$or` group, so a foreign short_code filter still yields empty rather than leaking.
- **CAS correctness**: the stored hash and the anonymous sentinel both live in the update filter, so concurrent claims of the same link resolve to exactly one winner. The eviction race (a click served from the stale cache entry between the CAS and the invalidate) is self-healing: such a click still matches the claimed-links arm by url_id, so no history is lost.
- **Index coexistence**: the new partial index shares its key pattern with the existing plain owner_id index; distinct name plus partialFilterExpression is legal and boot-time creation on the existing collection was exercised against a live instance.
- **Redaction**: `claim_token` is masked by the existing substring rule; the new test pins that.

## Should-fix

**Silent cap on the claimed-set lookup contradicts the claim budget.** `list_claimed_ids` truncates at 128 with no log, while the endpoint permits 16 per request at 30/min, 500/day. An API-heavy user who claims more than 128 links gets silently wrong dashboard totals, which is exactly the numbers-disagree failure this feature is designed to avoid. Raise the cap (the query is comfortably fast at four-digit set sizes) or log when truncation happens; silent is the only wrong option. Inline comment with specifics.

## Nits

- `routes/api_v1/shorten.py` docstring still says anonymous URLs "cannot be managed later" and are "not linked to any account" under Anonymous Usage Consequences, while the response now documents `claim_token` as the way to do exactly that. The OpenAPI page will contradict itself. Reword the two bullets.
- `list_claimed_ids` skips the try/except-with-context-log convention its sibling `list_aliases_by_owner_and_domain` follows, so a Mongo error there surfaces as a bare 500 on the dashboard path with no repo-level context.

## Merge safety

Merging ships nothing until a release is published. Note the endpoint is not flag-gated, which is fine: issuance is an additive response field and the claim route is new surface. Until the frontend stores tokens, links created through the site UI will discard their tokens and stay unclaimable, same as today; API and CLI consumers get the feature immediately on release. Worth shipping the frontend localStorage plumbing in the same release window so site-minted deeds stop evaporating.

## Good

The per-item result contract mirrors the reports intake shape, the token discipline matches API keys exactly (hash-only storage, single-use burn), claim fields are stripped when unset so the partial index can filter on `$exists`, and the test matrix covers the claim state machine including race loss and mixed-batch ordering. The stats arm collapsing to the byte-identical legacy query for accounts with no claims is the right zero-cost default.

repositories/url_repository.py
Zingzy: Should-fix: if an account crosses 128 claimed links, the excess silently drops out of scope=all stats. Dashboards undercount with no signal anywhere, and nothing bounds claims today (16 per call, 500 calls per day), so a motivated account can get there. Cheapest fix: log a warning when the cursor fills the cap, so an undercount report is diagnosable in one search. Dropping the cap until link quotas exist would also be defensible; the partial index keeps the read cheap either way.

services/url_service.py
```diff
@@ -931,7 +958,57 @@ async def create(
...
+                url_id, token_hash, owner_id, now
+            )
+            if not swapped:
+                # Lost a concurrent race — whoever won owns it now.
+                results.append(ClaimResult(item.url_id, "invalid"))
+                continue
+            # Evict so the next click stamps the new owner.
+            await self._url_cache.invalidate(existing.alias, existing.domain)
+            log.info(
+                "url_claimed",
```
Zingzy, line 1033: Should-decide: `update()` and the bulk paths emit `link.updated` per item and `delete()` emits `link.deleted`, but a claim, the one moment a link enters an account, emits nothing. A webhook subscriber never learns the link exists until some later edit touches it. Worth deciding now whether this should emit `link.updated` (or a dedicated `link.claimed`) after the CAS, because adding it later means integrations built in the meantime silently miss claims. Non-blocking either way given webhook adoption is young.

routes/api_v1/management.py
```diff
@@ -25,14 +25,59 @@
...
+    (ownership transferred, token burned), `already_yours` (idempotent
+    repeat), or `invalid` (unknown id, wrong token, or a link that is not
+    claimable — deliberately indistinguishable).
+
+    Claimed links join the account like any owned link: dashboard,
+    management, and full stats history included.
+
+    **Authentication**: Required.
+
+    **API Key Scope**: `urls:manage` or `admin:all`
```
Zingzy, line 73: Nit: the sibling endpoints all carry a `**Rate Limits**` section in the docstring and it lands in the OpenAPI description; this one omits it. Suggest `**Rate Limits**: 30/min, 500/day`.

Zingzy (commented):
Mergeable, nothing blocking. I traced the probe resistance, the ownership transfer, token handling end to end, and the stats query composition; all of it holds.

What holds up especially well:

- The token recipe is byte for byte the API key one (`generate_secure_token` + `hash_token`), plus `hmac.compare_digest` on the hashes. No new crypto surface.
- The claim state machine collapses unknown id, foreign owner, wrong token, and burned token into one `invalid`, and the CAS filter (hash + anonymous owner in the filter, hash unset in the same write) makes concurrent claims lose cleanly with no window where a token works twice.
- The absent-not-null discipline for `claim_token_hash` / `claimed_at` is enforced at the single insert path, documented at both the model and the index, and pinned by a dedicated test. That is exactly the null shape that caused the clicks bucket churn, handled correctly this time.
- An empty claimed set keeps the scope=all query byte-identical to the old stamp-only one, with a test asserting no `$or` leaks in, and the `$and` nesting with the null-sentinel groups is covered too.
- Service returns a frozen dataclass, route maps it to the DTO. Clean layering.

Findings are inline: one should-fix (silent truncation in `list_claimed_ids`), one should-decide (no domain event on claim), one docstring nit.

Merge vs deploy: merging ships nothing, but unlike most features this rides the next release with no flag, which is the right call here since any anonymous link created before issuance is permanently unclaimable. The tail before that release: API docs for the new endpoint and the `claim_token` response field, and the client-side storage plumbing on the frontend.

repositories/url_repository.py
Zingzy: Concrete numbers, reasoned from the failure modes rather than expected set size. The two sides fail in opposite ways: exceeding a read-side cap lies (dashboards silently undercount), exceeding a write-side cap refuses, loudly, while the user is watching. Refusing beats lying, so bound at write time and keep the read cap as a backstop.

1. Per-account claim ceiling at claim time: **500**, enforced with a `count` on the same `owner_claimed` partial index (sub-ms, one count per batch). That is roughly 10x any plausible organic user, since the UI surface is the onboarding wizard draining one browser's localStorage, dozens at most, while staying far below where the `$in` arm gets expensive (benchmarked at only +5-19ms per widget even at ~2.3k ids). When plan quotas land this constant becomes the free-tier number and the mechanism already exists.
2. Read-side cap raised **128 -> 1024**, plus a warning log when the cursor fills it. With the write ceiling at 500 it is unreachable by construction; if the log ever fires, something bypassed the write path, which is exactly what should be paged about.

If touching the write path is out of scope for this PR, the one-line version is just the second item: 1024 plus the warning log moves the silent-wrongness threshold from reachable-in-a-day to pathological, at a query cost that is noise.

The one value not to keep is 128 itself: it is calibrated to neither constraint, too high to matter for performance, too low to be unreachable.

Zingzy: Applied in d3a2774 exactly as prescribed: CLAIM_LIMIT_PER_ACCOUNT = 500 enforced at the top of claim() with a count on the owner_claimed partial index (conservative: counts the whole batch), refusing with a 403 and an url_claim_ceiling_hit log. Read cap raised to 1024 with a claimed_ids_truncated warning when the cursor fills; with the write ceiling in place, that log firing means bypass, as designed.

services/url_service.py
Zingzy: Decided and shipped in d3a2774: dedicated link.claimed. The registry doctrine settles it: an event is a fact someone would act on without having been there, and this is the account's first sight of the link, so link.updated would be a changes map for a link the subscriber never saw. Emitted after the CAS and cache eviction with the new owner stamped for routing; registry entry drives catalog, docs sample, and test sends. Frontend event copy for the webhook dialog rides the frontend PR.

routes/api_v1/management.py
Zingzy: Added in d3a2774.

repositories/url_repository.py
Zingzy: Wrapped in d3a2774 with the sibling's try/except and a repo_list_claimed_ids_failed context log.

schemas/dto/requests/url.py
```diff
@@ -518,3 +518,33 @@ def _norm_domain(cls, v: str | None) -> str | None:
...
+        max_length=24,
+        pattern=r"^[0-9a-f]{24}$",
+        examples=["507f1f77bcf86cd799439011"],
+    )
+    token: str = Field(
+        description=(
+            "The one-time claim_token returned by the anonymous shorten "
+            "call — the bearer proof of creation."
+        ),
+        min_length=16,
```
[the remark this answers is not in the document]
Zingzy: Tightened to min_length=43 in d3a2774, matching the issued size.

services/stats_service.py
```diff
@@ -595,8 +609,22 @@ async def query(
                 raise AuthenticationError("authentication required for scope=all")
 
         # ── Execute + format ──────────────────────────────────────────────────
+        # Claimed links carry pre-claim history under the anonymous stamp;
+        # sub-ms partial-index lookup, empty for almost every account.
+        claimed_url_ids: list[ObjectId] = []
+        if scope == StatsScope.ALL and owner_id:
+            claimed_url_ids = await self._url_repo.list_claimed_ids(
+                ObjectId(owner_id) if isinstance(owner_id, str) else owner_id
+            )
```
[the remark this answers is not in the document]
Zingzy: Resolved in d3a2774, stricter than the suggestion: a write-side per-account ceiling of 500 enforced at claim time (counted on the owner_claimed partial index) makes the read cap unreachable by construction, and the cap itself rose to 1024 with a warning log if a cursor ever fills it. The 129-id regression test is superseded by ceiling tests: a batch that would cross the limit is refused before any lookup, and one that exactly fills it succeeds.

services/url_service.py
Zingzy: Correction to my earlier reply: reverted in the latest push. Weighing the full doctrine, the causer rule says actor-caused changes ride link.updated, and a claim cannot be expressed honestly there because ownership is not a payload field. When the contract cannot express a thing, the answer is not to grow a forever-contract, it is to wait for an integration that proves the need. Claim emits nothing for now; the decision and reasoning are recorded in the claim flow itself.

repositories/url_repository.py
Zingzy: Verified at e4c7e06: write-side ceiling of 500 enforced on the partial-index count before the loop, read backstop raised to 1024 with the truncation warning. The batch-conservative refusal (counting items before outcomes are known) is the right kind of conservative. Resolved.

services/url_service.py
Zingzy: Decision made and documented at the emit point in e4c7e06 (event catalog is forever, ownership is not a payload field, revisit on demonstrated integration need). That is all this asked for. Resolved.

routes/api_v1/management.py
Zingzy: Fixed in e4c7e06.

services/url_service.py
```diff
@@ -931,7 +965,81 @@ async def create(
...
+            log.info(
+                "url_claim_ceiling_hit",
+                user_id=str(owner_id),
+                claimed=already_claimed,
+                batch=len(claims),
+            )
+            raise ForbiddenError(
+                f"Claim limit reached: an account can claim up to "
+                f"{CLAIM_LIMIT_PER_ACCOUNT} links"
+            )
```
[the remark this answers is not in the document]
Zingzy: Real race, deliberately tolerated rather than engineered away: the ceiling is a soft product bound, not a security invariant. Worst case is one extra batch (16) per concurrent request inside the read window, the endpoint runs at 30/min per credential, and the read-side backstop sits at 1024, far above any realistic overshoot. Atomic reservation with release-on-invalid accounting is more machinery than a soft cap is worth at this scale; the docstring now records the trade explicitly. When plan quotas land this constant becomes a billed entitlement and the enforcement moves wherever quotas live.

PR #283, opened 2026-08-08
fix(metadata): gate the destination fetcher on the meta tags flag

Zingzy:
## The inconsistency

`GET /api/v1/metadata` fetches a destination page and returns its parsed meta tags. It is the prefill companion to custom meta tags: the client calls it, then sends the result back through `POST /api/v1/shorten` or `PATCH /api/v1/urls/{id}`.

Both of those write paths call `flag_svc.require(META_TAGS_FLAG, user)`. The fetcher did not. It only required auth and `urls:read`, so writing meta tags was gated while the endpoint that exists to support that writing was open to every authenticated caller, whether or not the feature was enabled for their account.

The consequence is a wider surface than intended. This endpoint makes server side requests to arbitrary third party URLs. It is SSRF guarded, rate limited (20/min, 500/day) and cached, so the exposure is bounded, but an ungated authed fetch proxy is available to far more accounts than the feature it serves.

## The fix

Adds `await flag_svc.require(META_TAGS_FLAG, user)` as the first statement of the handler, through the same `FeatureFlagSvc` dependency alias the sibling routes use.

403, not a hiding 404. The custom domains routes hide behind 404 because the existence of that feature is itself concealed. `custom_meta_tags` is not concealed: the write paths already answer 403 for it, `GET /api/v1/me/features` reports its state, and this endpoint is published in the OpenAPI spec. A 404 here would conceal nothing and would disagree with its own siblings. 403 is already in the endpoint's declared responses, so the documented contract does not change shape.

## Audit of the other routes

Checked every route under `routes/api_v1/` against the five registered flags. No other gaps:

- `custom_domains`: create is gated (404); reads, verify and delete stay open on purpose so a de-allowlisted owner can still see and remove domains.
- `webhooks`: create, test send and retry are gated, which are the paths that make outbound calls. Passive management of existing endpoints is intentionally not gated.
- `geo_targeting`: gated on both write paths. The redirect path applies already stored rules and is not an authoring path.
- `ab_testing`: registered as a flag, no endpoints yet.

## Tests

`test_metadata_requires_flag` asserts 403 when the flag is off and that no fetch is attempted, using the same `_flag_svc` mock shape as the existing meta tags and custom domain route tests. The rest of the metadata tests now build their app with the flag on explicitly rather than leaning on the default mock.

Full suite: 3002 passed.

## Summary by CodeRabbit

* **New Features**
  * Metadata access is now controlled by the `custom_meta_tags` feature flag.
  * Requests without access receive a forbidden response before any metadata processing occurs.

* **Bug Fixes**
  * Prevented unauthorized metadata requests from triggering URL validation, caching, fetching, or parsing.

### DIFF
--- a/tests/integration/api_v1/test_metadata.py
+++ b/tests/integration/api_v1/test_metadata.py
@@ -6,7 +6,8 @@ from unittest.mock import AsyncMock, patch
 
 from fastapi.testclient import TestClient
 
-from dependencies import get_current_user
+from dependencies import get_current_user, get_feature_flag_service
+from errors import ForbiddenError
 from infrastructure.cache.meta_fetch_cache import MetaFetchCache
 from infrastructure.safe_fetch import FetchedBody, FetchHardError, FetchTransientError
 
@@ -18,8 +19,27 @@ HTML = b"""<html><head>
 </head><body></body></html>"""
 
 
-def _app(user=None):
-    app = _build_test_app({get_current_user: lambda: user or _make_user()})
+def _flag_svc(enabled: bool) -> AsyncMock:
+    svc = AsyncMock()
+    svc.is_enabled = AsyncMock(return_value=enabled)
+
+    async def _require(name: str, user, **_kw) -> None:
+        # Mirrors FeatureFlagService.require: raise 403 when disabled.
+        if not enabled:
+            feature = name.replace("_", " ").capitalize()
+            raise ForbiddenError(f"{feature} is not enabled for this account")
+
+    svc.require = AsyncMock(side_effect=_require)
+    return svc
+
+
+def _app(user=None, flag_enabled: bool = True):
+    app = _build_test_app(
+        {
+            get_current_user: lambda: user or _make_user(),
+            get_feature_flag_service: lambda: _flag_svc(flag_enabled),
+        }
+    )
     app.state.meta_fetch_cache = MetaFetchCache(None)  # no-op cache
     return app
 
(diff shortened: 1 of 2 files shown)


tests/integration/api_v1/test_metadata.py
```diff
@@ -18,8 +19,27 @@
 </head><body></body></html>"""
 
 
-def _app(user=None):
-    app = _build_test_app({get_current_user: lambda: user or _make_user()})
+def _flag_svc(enabled: bool) -> AsyncMock:
```
Zingzy, line 22: This is now the second verbatim copy of the helper in `test_meta_tags_api.py`, and the repo carries two more variants (`test_shorten.py`, `test_custom_domain_routes.py`). Each hand-mirrors `FeatureFlagService.require`, which is the kind of copy that drifts silently when the real signature or error shape changes. Worth promoting one require-faithful version to a shared test helper next time someone is in here; not worth blocking a two-file fix on.

Zingzy (commented):
Mergeable, nothing blocking. The gate matches its siblings exactly (`shorten.py:106` and `management.py:144` use the same `flag_svc.require(META_TAGS_FLAG, user)` through the same dependency alias), it sits ahead of any validation or fetch work, the new test proves no fetch is attempted on 403, and the docstring carries the gate into the OpenAPI description so the contract change is documented where the endpoint is published. The 403-over-404 reasoning is correct: the write paths already answer 403 for this flag and `/me/features` reports its state, so a hiding 404 here would conceal nothing and disagree with the endpoint's own siblings.

Also traced the interaction with the frontend fix that ships separately, in both deploy orders: an unentitled account on a stale bundle still fires this fetch and now gets a 403, and renders nothing differently, because every consumer of that query already sat inside the feature-gated block; the error notice included. Invisible either way.

One nit inline on the test helper. The audit section is what makes this a complete fix rather than a spot patch, and writing down which routes stay open on purpose (custom-domain reads for de-allowlisted owners, passive webhook management) is worth as much as the gate itself.

tests/integration/api_v1/test_metadata.py
Zingzy: Agreed, leaving as is here. The shared require-faithful helper is worth doing when someone is next in these files; four hand-mirrored copies is already past the point where a signature change would go unnoticed.

PR #287, opened 2026-08-12
fix: scope the self-link check to the destination host

Zingzy:
`validate_url` treated any URL containing the string `spoo.me` as self-referential, including matches in the path, query or fragment. Shortening an analytics dashboard URL filtered on `spoo.me` was rejected with "URL is not allowed or invalid", and the dashboard surfaced it as "Short links can't point back at spoo.me."

A destination only loops back to us when the request would actually arrive here, which the host decides on its own. This matches on the parsed host instead, or a subdomain of it.

```
https://eu.posthog.com/project/1?filter=spoo.me   accepted (was rejected)
https://example.com/spoo.me/guide                 accepted (was rejected)
https://spoo.me/abc                               rejected
https://www.spoo.me/abc                           rejected
https://notspoo.me/abc                            accepted
```

`urlparse().hostname` drops userinfo and the port, so `https://example.com@spoo.me/` is still blocked and `https://spoo.me@example.com/` is not.

The frontend mirrors this validator to pre-validate the field, so spoo-me/frontend#42 carries the matching change. This one needs to deploy first, otherwise the form accepts a URL the API still rejects.

## Testing

Existing cases still pass, including the `www.spoo.me` one that now relies on the subdomain clause. Added coverage for path, query and fragment mentions, the lookalike host, and both userinfo directions. 58 passed.

## Summary by CodeRabbit

* **New Features**
  * Added a utility to preview and replay failed click events, with configurable limits and cleanup behavior.
  * Added configurable limits for dead-letter queues.
* **Bug Fixes**
  * Improved URL validation to accurately detect blocked hosts and subdomains while avoiding false positives in paths, queries, and fragments.
  * Click records now omit empty optional fields instead of storing null values.
* **Improvements**
  * Updated click time-series storage to use hourly granularity, improving long-term scalability.

### DIFF
--- a/repositories/indexes.py
+++ b/repositories/indexes.py
@@ -64,7 +64,11 @@ async def ensure_indexes(
             timeseries={
                 "timeField": "clicked_at",
                 "metaField": "meta",
-                "granularity": "seconds",
+                # "hours" spans buckets over 30 days. Most links get sparse
+                # clicks, so finer granularity degenerates into one bucket
+                # per click. Existing deploys need a one-time collMod to
+                # match (the transition is one-way, coarser only).
+                "granularity": "hours",
             },
         )
     except (CollectionInvalid, OperationFailure) as e:
--- a/scripts/replay_dlq.py
+++ b/scripts/replay_dlq.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env -S uv run --script
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#     "redis>=5",
+# ]
+# ///
+"""Replay dead-lettered click events back onto the main stream.
+
+Standalone — reads ``CLICK_EVENTS_QUEUE_REDIS_URI`` from the environment.
+Replayed events fan out to every consumer group again (streams have no
+per-group publish); replayed entries leave the DLQ unless ``--keep``.
+
+Usage::
+
+    uv run --env-file .env.production scripts/replay_dlq.py --dry-run
+    uv run --env-file .env.production scripts/replay_dlq.py --limit 500 --keep
+"""
+
+from __future__ import annotations
+
+import argparse
+import os
+import sys
+
+from redis import Redis
+
+STREAM_FIELD_DATA = "__data__"  # payload field the workers' parser consumes
+DLQ_FIELD_SOURCE_ID = "dlq_source_id"
+DLQ_FIELD_GROUP = "dlq_group"
+DLQ_FIELD_REASON = "dlq_reason"
+
+_BATCH = 200
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--dry-run",
+        action="store_true",
+        help="Show what would be replayed without writing.",
+    )
+    parser.add_argument(
+        "--limit", type=int, default=0, help="Replay at most N entries (0 = all)."
+    )
+    parser.add_argument(
+        "--keep",
+        action="store_true",
+        help="Do not delete replayed entries from the DLQ.",
+    )
+    parser.add_argument("--stream", default="events:clicks")
+    parser.add_argument("--dlq-stream", default="events:clicks:dlq")
+    args = parser.parse_args()
+
+    uri = os.environ.get("CLICK_EVENTS_QUEUE_REDIS_URI")
+    if not uri:
+        sys.exit("CLICK_EVENTS_QUEUE_REDIS_URI not set in environment.")
+
+    redis = Redis.from_url(uri, decode_responses=True)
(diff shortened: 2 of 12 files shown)


shared/validators.py
```diff
@@ -27,18 +44,22 @@ def validate_url(
...
+        blocked_self_domains: Bare hostnames that mark the URL as
+            self-referential when they are the destination's HOST (or a
+            parent of it). Defaults to ``("spoo.me",)`` to prevent redirect
+            loops. Matching is host-scoped: a destination that merely
+            mentions the name in its path or query is not a loop.
     """
     # Scheme allowlist defends against ftp/file/data/etc even if the
     # validators package widens its accepted schemes upstream.
-    if urlparse(url).scheme not in _ALLOWED_URL_SCHEMES:
+    parsed = urlparse(url)
```
Zingzy: `urlparse` itself raises on a malformed IPv6 authority: `urlparse("https://[::1")` throws `ValueError: Invalid IPv6 URL`. `long_url` is a plain `str` in the DTO, so the raw string reaches this line and the exception escapes as a 500 on the shorten endpoints. Pre-existing (the old code called `urlparse(url).scheme` too), but since this PR touches the exact line: wrap the parse in `try/except ValueError: return False` and add `https://[::1` to the parametrized cases. Anonymous rate-limited surface, so the cost today is Sentry noise and an ugly 500 instead of a clean 400.

scripts/replay_dlq.py
```diff
@@ -0,0 +1,105 @@
...
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#     "redis>=5",
+# ]
+# ///
+"""Replay dead-lettered click events back onto the main stream.
+
+Standalone — reads ``CLICK_EVENTS_QUEUE_REDIS_URI`` from the environment.
+Replayed events fan out to every consumer group again (streams have no
```
Zingzy, line 11: The fan-out note stops one sentence short of the consequence. A message dead-lettered by one group was already processed and acked by the others, so replaying re-delivers to every group: duplicate hotness increments, duplicate click inserts when stats had succeeded (no dedupe key on a time-series insert), and duplicate webhook fires under a fresh stream id, which consumer-side idempotency will not catch. After a global outage the replay is clean; after a partial-group failure it is not, and the per-group breakdown this script prints is the signal to read first. Worth saying that here so the operator runs `--dry-run`, checks the group counts, and decides with eyes open.

Zingzy (commented):
Reviewed the full diff and ran the five affected test files in a clean worktree: 95 passed. Mergeable, nothing blocking; inline notes below.

The host-scoped check is the right semantic: the host alone decides whether a request loops back, and `urlparse().hostname` already strips userinfo and the port, so both smuggling directions stay closed. The normalization matches `system_default_domain` (lowercase, strip trailing dots), so the derived blocklist compares cleanly. The `to_mongo` override preserves the base contract too: `exclude_none=True` drops a None `_id` the same way the base's explicit pop does.

Deploy ordering: this goes out before spoo-me/frontend#42, otherwise the form accepts URLs the API still rejects. Merging ships nothing either way.

One trade worth stating out loud: the old substring check accidentally rejected impersonation URLs like `https://phisher.example/spoo.me/login`, and those now validate. Loop prevention was never the right place to catch them; if they should be caught, the regex blocklist is the honest home.

workers/dlq.py
```diff
@@ -42,11 +42,13 @@ def __init__(
         group: str,
         dlq_stream: str,
         max_deliveries: int,
+        dlq_maxlen: int = DLQ_MAXLEN,
```
Zingzy, line 45: Two bounds exist now: this constant (10k, still the constructor default) and the config default (100k). Prod always takes the config path, so the default here is a fallback nobody uses, and the comment above `DLQ_MAXLEN` tells a different sizing story than the one in config.py. Drop the parameter default and let config be the single source; the tests already pass it explicitly.

repositories/indexes.py
```diff
@@ -64,7 +64,11 @@ async def ensure_indexes(
             timeseries={
                 "timeField": "clicked_at",
                 "metaField": "meta",
-                "granularity": "seconds",
+                # "hours" spans buckets over 30 days. Most links get sparse
+                # clicks, so finer granularity degenerates into one bucket
+                # per click. Existing deploys need a one-time collMod to
+                # match (the transition is one-way, coarser only).
+                "granularity": "hours",
```
Zingzy, line 71: The collMod note is right but nothing operational carries it; add the one-time collMod to the release checklist so it does not evaporate between merge and release. Also worth remembering that this and the exclude_none change only help buckets written afterwards. The existing fragmented backlog stays as it is unless the collection is rewritten; letting it age is probably fine, but that is a decision, not a default.

PR #288, opened 2026-08-14
feat(stats): domain dimension on the stats API

Zingzy:
Adds **domain** as a stats dimension: per-domain breakdowns and filtering on `/api/v1/stats` and `/api/v1/export`, completing the analytics side of custom domains.

## What changed

**Stats API:** `domain` is now a valid `group_by` dimension and filter (both the `filters` JSON and a dedicated query param). Grouping runs on `meta.domain`, which every click already carries, and the sparse `(meta.domain, clicked_at)` index already exists, so this is read-side only: no hot-path, schema, or migration work.

**Unknown bucket:** clicks recorded before domain stamping existed have a null or absent `meta.domain` (old time-series buckets keep their shape forever). Those aggregate under `unknown`, and filtering by `unknown` matches the same null/missing documents via the existing null-sentinel machinery, so filter counts always agree with group-by. `unknown` can never collide with a real value since domains are always dotted fqdns.

**Field path mapping:** domain is the first filterable dimension living under the time-series metaField (`short_code` also does, but it has its own branch for scope-bypass prevention). The query builder now maps dimensions to their Mongo field paths instead of assuming top-level names; a filter keyed on bare `domain` would have silently matched nothing.

Exports pick the dimension up automatically. The public per-link stats payload is deliberately unchanged: a single link lives on exactly one domain, so a domain breakdown there is meaningless.

## Tests

- DTO tests: `group_by=domain`, the dedicated param, and the `filters` JSON form
- Query builder tests: meta path targeting, the `unknown` sentinel matching null/missing, and the group-by/filter agreement invariant
- Aggregation strategy tests: nested field grouping, sentinel fallback, limit (50)
- Full suite green: 3016 passed
- Live smoke on a real Mongo 8 time-series collection: seeded clicks with a real domain, an explicit null, and an absent key; verified bucket counts, both filter shapes, and group-by/filter agreement end to end

## Summary by CodeRabbit

* **New Features**
  * Added domain-based grouping and filtering to statistics.
  * Supports comma-separated domain filters and JSON filter input.
  * Domain values use lowercase FQDN format.
  * Missing, empty, or unavailable domain data is grouped and filterable as `unknown`.
  * Domain groupings return up to 50 results.

* **Tests**
  * Added coverage for domain filters, grouping, fallback handling, and result formatting.

### DIFF
--- a/schemas/dto/requests/stats.py
+++ b/schemas/dto/requests/stats.py
@@ -254,6 +254,15 @@ class StatsQuery(RequestBase):
                     continue
                 parsed_filters[dim] = _parse_comma_separated(raw)
 
+        # Domains are case-insensitive by definition and stored lowercase
+        # (model normalizer), so fold caller casing — "Spoo.me" must not
+        # silently match nothing. The only normalized dimension; all others
+        # filter on case-sensitive stored values.
+        if StatsDimension.DOMAIN in parsed_filters:
+            parsed_filters[StatsDimension.DOMAIN] = [
+                v.lower() for v in parsed_filters[StatsDimension.DOMAIN]
+            ]
+
         self._parsed_filters = parsed_filters
 
         return self
--- a/schemas/dto/requests/stats.py
+++ b/schemas/dto/requests/stats.py
@@ -26,6 +26,7 @@ from schemas.dto.requests._descriptions import (
     STATS_CITY_DESC,
     STATS_COUNTRY_DESC,
     STATS_DEVICE_DESC,
+    STATS_DOMAIN_DESC,
     STATS_END_DATE_DESC,
     STATS_FILTERS_DESC,
     STATS_GROUP_BY_DESC,
@@ -154,6 +155,12 @@ class StatsQuery(RequestBase):
         description=STATS_REFERRER_DESC,
         examples=["https://google.com,https://twitter.com"],
     )
+    domain: str | None = Field(
+        default=None,
+        max_length=1000,
+        description=STATS_DOMAIN_DESC,
+        examples=["spoo.me,links.example.com"],
+    )
     utm_source: str | None = Field(
         default=None,
         max_length=1000,
@@ -234,6 +241,7 @@ class StatsQuery(RequestBase):
             "country",
             "city",
             "referrer",
+            "domain",
             "short_code",
             "utm_source",
             "utm_medium",
--- a/services/stats_service.py
+++ b/services/stats_service.py
@@ -71,9 +71,17 @@ _KNOWN_TIMEZONES: frozenset[str] = frozenset(available_timezones())
 _NULL_SENTINEL_FILTERS: dict[StatsDimension, str] = {
     StatsDimension.REFERRER: "Direct",
     StatsDimension.DEVICE: "unknown",
+    StatsDimension.DOMAIN: "unknown",
     StatsDimension.UTM_SOURCE: "(none)",
     StatsDimension.UTM_MEDIUM: "(none)",
     StatsDimension.UTM_CAMPAIGN: "(none)",
+}
+
+# Dimensions whose clicks field lives under the time-series metaField rather
+# than at the top level. short_code is not listed — it has its own branch
+# (scope-bypass prevention).
+_FILTER_FIELD_PATHS: dict[StatsDimension, str] = {
+    StatsDimension.DOMAIN: "meta.domain",
 }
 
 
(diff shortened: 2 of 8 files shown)


services/stats_service.py
```diff
@@ -71,11 +71,19 @@
...
+    StatsDimension.DOMAIN: "unknown",
     StatsDimension.UTM_SOURCE: "(none)",
     StatsDimension.UTM_MEDIUM: "(none)",
     StatsDimension.UTM_CAMPAIGN: "(none)",
 }
 
+# Dimensions whose clicks field lives under the time-series metaField rather
+# than at the top level. short_code is not listed — it has its own branch
+# (scope-bypass prevention).
+_FILTER_FIELD_PATHS: dict[StatsDimension, str] = {
```
Zingzy, line 83: **Nit (drift guard for the next dimension):** the mapping here and the strategy registry's `"$meta.domain"` are two hand-kept copies of the same fact. For `domain` they're locked together by the new tests, but the next meta-path dimension can update the registry and forget this map, which reproduces exactly the silent-empty-filter bug this dict exists to prevent, and nothing would fail.

A small parametrized invariant test would close it for good: for every dimension in `ALLOWED_FILTERS` except `short_code`, assert the strategy's mongo field with the `$` stripped equals `_FILTER_FIELD_PATHS.get(dim, dim)`. Cheap now, and it turns the next dimension's field-path wiring into a test failure instead of a wrong-numbers report.

schemas/dto/requests/stats.py
```diff
@@ -154,6 +155,12 @@ class StatsQuery(RequestBase):
         description=STATS_REFERRER_DESC,
         examples=["https://google.com,https://twitter.com"],
     )
+    domain: str | None = Field(
```
Zingzy, line 158: **Nit (conscious call, flagging for the record):** domain is the one dimension whose values are case-insensitive by definition, and the stored side is guaranteed lowercase by the model normalizer. A caller sending `Spoo.me` gets silently empty results. The docs do say exact lowercase fqdn, and no other dimension normalizes, so leaving it verbatim is defensible for uniformity; but a `.lower()` on just this dimension's parsed values would erase the footgun at the cost of one special case. Fine either way, just deciding it on purpose beats inheriting it.

Zingzy (commented):
Mergeable, nothing blocking. Two nits inline; everything load-bearing checks out.

**Verified, not just read:**
- `meta.domain` really is stamped on every current click path. v2 docs carry a required, lowercased fqdn (`UrlDocV2._normalise_domain`), and the legacy/emoji cache projections stamp `system_default_domain` explicitly, so `unknown` genuinely means pre-stamping buckets (plus at most a brief tail of stale cached entries from the stamping deploy, which the `or None` coercion in the click handler already folds in). Filtering `spoo.me` matches default-domain traffic as documented.
- The tricky intersection this PR is first to hit, a dimension that is both meta-path and null-sentinel, is handled correctly: all three arms of the sentinel `$or` group go through the field-path mapping, and the group-by/filter agreement invariant test follows the device precedent. Ran the three affected test files at the head commit in a clean worktree: 137 passed.
- Export flows through with zero changes (formatters iterate `metrics` keys generically), and `public_stats_service._V2_GROUP_BY` is untouched, so nothing bleeds into the public per-link payload.

**Architecture read:** introducing `_FILTER_FIELD_PATHS` instead of a fourth hardcoded branch is the right call. `short_code` staying out of the map (its branch does scope-bypass prevention, not just path mapping) is correct and the comment says exactly that. Sentinel choice is symmetric on both sides and the no-collision argument (domains are always dotted) holds; `device` already proves `unknown` is safe even when it doubles as a stored value.

**One perf footnote, not a finding:** the sparse `(meta.domain, clicked_at)` index cannot serve the `unknown` sentinel arms (`$exists: false` / null), so that filter shape leans on the owner/time match alone. At dashboard read volume that's fine; just don't expect the sparse index to help the unknown bucket if it ever shows up in a slow-query report.

Merge is safe and this can ride the next release un-flagged: read-side only, additive query surface, no hot-path or write-path changes. This also closes the custom-domains stats gap.

shared/aggregation_strategies.py
```diff
@@ -274,6 +274,12 @@ class AggregationStrategyFactory:
         "short_code": lambda: FieldAggregationStrategy(
             "$meta.short_code", "short_code", 100
         ),
+        # "unknown" = clicks recorded before domain stamping existed
+        # (time-series buckets can't be backfilled). It can never collide
+        # with a real value: domains are always dotted fqdns.
+        "domain": lambda: FieldAggregationStrategy(
+            "$meta.domain", "domain", 50, default="unknown"
+        ),
```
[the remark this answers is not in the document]
Zingzy: Fixed in a15f615. Applied at the FieldAggregationStrategy level for every sentinel dimension, not just domain: the filter side's sentinel arm already matched "" for device/referrer/utm too, so the same group-by disagreement was latent there. Group expression now folds "" into null before the $ifNull. Regression test pins the expression shape, and a live smoke against a Mongo 8 time-series collection verified null, missing-key, and empty-string clicks all land in the unknown bucket for both group-by and filter.

services/stats_service.py
Zingzy: Added in a15f615: test_filter_field_paths_agree_with_strategy_registry iterates ALLOWED_FILTERS (short_code exempt, its branch is scope-bypass prevention) and pins the strategy's mongo field against the map. The next meta-path dimension now fails a test instead of silently filtering on a wrong field.

schemas/dto/requests/stats.py
Zingzy: Decided for folding, in a15f615: domain filter values are lowercased at parse time (both the param and the filters JSON path), and the docs now say case-insensitive instead of demanding exact lowercase. Rationale: domains are case-insensitive by definition and the stored side is guaranteed lowercase by the model normalizer, so verbatim uniformity here preserves a footgun with no upside. It stays the only normalized dimension; the test asserts browser values pass through untouched.

PR #289, opened 2026-08-14
feat: tighten anonymous and legacy rate limits

Zingzy:
## What

- v1 anonymous tier daily cap: 1000 to 200 (burst stays 20/min)
- Legacy v0 shorten endpoints (POST /, POST /emoji): 5/min + 50/day
- Legacy /{code}+ preview now shares the v1 public preview budget (30/min + 2000/day)
- Anonymous and legacy JSON 429s carry a `hint` field pointing at the authenticated v1 tiers

## Why

The legacy endpoints have run at 100/min with no daily cap since the FastAPI migration, while the docs promise 10/min + 100/hour + 500/day. That made v0 the loosest anonymous minting path on the service. Real traffic over the last week: legacy POST / has had no successful callers since the frontend cutover, POST /emoji serves about 20 requests a week, and 90% of anonymous v1 creator IPs never exceed 10 links a day. The new caps sit far above human usage while cutting per-IP campaign minting capacity, and the 429 hints give script authors the migration path instead of a bare error.

Test suite green apart from the pre-existing app-token-scopes and click-sink smoke failures on main.

A docs PR updating the published numbers follows and should merge once this deploys.

## Summary by CodeRabbit

* **New Features**
  * Rate-limit responses now provide helpful guidance tailored to the affected API route and access method.
  * Legacy shortener rate-limit messages direct users toward the current shortening endpoint.

* **Changes**
  * Anonymous API usage is now limited to 200 requests per day.
  * Legacy URL shortening is limited to 5 requests per minute and 50 per day.
  * Preview requests use the standard preview limit instead of the legacy shortener limit.

* **Bug Fixes**
  * Improved consistency of rate-limit guidance for anonymous, authenticated, and legacy requests.

### DIFF
--- a/middleware/error_handler.py
+++ b/middleware/error_handler.py
@@ -17,6 +17,7 @@ from starlette.exceptions import HTTPException as StarletteHTTPException
 from errors import AppError
 from infrastructure.logging import get_logger
 from infrastructure.templates import templates
+from middleware.rate_limiter import rate_limit_key
 
 log = get_logger(__name__)
 
@@ -146,13 +147,28 @@ def register_error_handlers(app: FastAPI) -> None:
             content=_validation_error_response(exc.errors()),
         )
 
+    def _rate_limit_hint(request: Request) -> str | None:
+        if request.url.path.startswith("/api/v1"):
+            if rate_limit_key(request).startswith(("apikey:", "jwt:")):
+                return None
+            return (
+                "Anonymous requests get the lowest limits. Authenticate with "
+                "a free account or an API key for 60 per minute and 5000 per day."
+            )
+        if request.method == "POST" and request.url.path in ("/", "/emoji"):
+            return (
+                "The legacy API is strictly rate limited. Use POST /api/v1/shorten "
+                "for higher limits: https://docs.spoo.me/rate-limits"
+            )
+        return None
+
     @app.exception_handler(RateLimitExceeded)
     async def rate_limit_handler(request: Request, exc: RateLimitExceeded) -> Response:
         if _wants_json(request):
-            return JSONResponse(
-                status_code=429,
-                content={"error": "Too many requests", "code": "rate_limit_exceeded"},
-            )
+            content = {"error": "Too many requests", "code": "rate_limit_exceeded"}
+            if hint := _rate_limit_hint(request):
+                content["hint"] = hint
+            return JSONResponse(status_code=429, content=content)
         return _error_html(request, 429, "TOO MANY REQUESTS", "rate_limit_exceeded")
 
     @app.exception_handler(StarletteHTTPException)
--- a/middleware/error_handler.py
+++ b/middleware/error_handler.py
@@ -70,6 +70,24 @@ def _wants_json(request: Request) -> bool:
     )
 
 
+def _rate_limit_hint(request: Request) -> str | None:
+    # Mirrors rate_limit_key on purpose: the hint must match the tier the
+    # limiter actually applied, validated or not.
+    if request.url.path.startswith("/api/v1"):
+        if rate_limit_key(request).startswith(("apikey:", "jwt:")):
+            return None
+        return (
+            "Anonymous requests get the lowest limits. Authenticate with "
+            "a free account or an API key for 60 per minute and 5000 per day."
+        )
+    if request.method == "POST" and request.url.path in ("/", "/emoji"):
+        return (
+            "The legacy API is strictly rate limited. Use POST /api/v1/shorten "
+            "for higher limits: https://docs.spoo.me/rate-limits"
+        )
+    return None
+
+
 # Statuses that return EMPTY bodies under EDGE_COMPOSED_ERRORS — must stay
 # a subset of what Caddy's handle_response matcher composes, or visitors
 # get blank pages (verify at cutover; redirect ⊆ app-level is test-pinned).
@@ -147,29 +165,22 @@ def register_error_handlers(app: FastAPI) -> None:
             content=_validation_error_response(exc.errors()),
         )
 
-    def _rate_limit_hint(request: Request) -> str | None:
-        if request.url.path.startswith("/api/v1"):
-            if rate_limit_key(request).startswith(("apikey:", "jwt:")):
-                return None
-            return (
-                "Anonymous requests get the lowest limits. Authenticate with "
-                "a free account or an API key for 60 per minute and 5000 per day."
-            )
-        if request.method == "POST" and request.url.path in ("/", "/emoji"):
-            return (
-                "The legacy API is strictly rate limited. Use POST /api/v1/shorten "
-                "for higher limits: https://docs.spoo.me/rate-limits"
-            )
-        return None
-
     @app.exception_handler(RateLimitExceeded)
     async def rate_limit_handler(request: Request, exc: RateLimitExceeded) -> Response:
+        hint = _rate_limit_hint(request)
         if _wants_json(request):
             content = {"error": "Too many requests", "code": "rate_limit_exceeded"}
-            if hint := _rate_limit_hint(request):
+            if hint:
                 content["hint"] = hint
-            return JSONResponse(status_code=429, content=content)
-        return _error_html(request, 429, "TOO MANY REQUESTS", "rate_limit_exceeded")
+            return JSONResponse(
+                status_code=429,
+                content=content,
+                headers={"X-Spoo-Hint": hint} if hint else None,
+            )
+        response = _error_html(request, 429, "TOO MANY REQUESTS", "rate_limit_exceeded")
+        if hint:
+            response.headers["X-Spoo-Hint"] = hint
+        return response
 
     @app.exception_handler(StarletteHTTPException)
     async def http_exception_handler(
--- a/middleware/rate_limiter.py
+++ b/middleware/rate_limiter.py
@@ -32,7 +32,7 @@ class Limits:
 
     # API v1 — authenticated vs anonymous tiers
     API_AUTHED = "60 per minute; 5000 per day"
-    API_ANON = "20 per minute; 1000 per day"
+    API_ANON = "20 per minute; 200 per day"
 
     # Alias availability check — cheap read, UI debounces on each keystroke
     API_CHECK_AUTHED = "180 per minute; 10000 per day"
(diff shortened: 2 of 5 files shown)


middleware/rate_limiter.py
```diff
@@ -32,7 +32,7 @@ class Limits:
 
     # API v1 — authenticated vs anonymous tiers
     API_AUTHED = "60 per minute; 5000 per day"
-    API_ANON = "20 per minute; 1000 per day"
+    API_ANON = "20 per minute; 200 per day"
```
Zingzy, line 35: **Should-decide (collateral scope):** `API_ANON` is not just the minting tier — it also backs anonymous `GET /api/v1/stats` (stats.py:27). So the 1000 → 200 daily cut applies to anonymous *reads* of public stats as well as creates, and the PR's evidence (90% of anonymous v1 **creator** IPs never exceed 10 links a day) only covers the mint side.

The shape that gets squeezed is a shared or NAT'd egress IP doing read-heavy polling of public stats — a small dashboard or script hitting a handful of codes on a loop reaches 200/day sooner than you'd expect, and the limiter keys anonymous callers purely by IP, so a whole office or campus shares one 200/day bucket. Rate limits are per-IP not per-user by design, but the read path is where the reuse actually stings.

Not a blocker, and probably still the right number, but worth either (a) confirming from Axiom that anonymous `/api/v1/stats` daily-per-IP volume also sits under ~200, or (b) splitting the read tier so stats keeps a higher daily cap than shorten. If you keep them shared, the 429 volume for anonymous v1 is the metric to watch for a day or two after the release.

Worth noting the docs PR inherits this too: it advertises 20/min + 200/day on the `GET /api/v1/stats` row, so whatever you decide here needs to stay in step there.

middleware/error_handler.py
```diff
@@ -146,13 +147,28 @@ async def pydantic_validation_error_handler(
             content=_validation_error_response(exc.errors()),
         )
 
+    def _rate_limit_hint(request: Request) -> str | None:
```
Zingzy: **Nit (drift):** the hint hardcodes "60 per minute and 5000 per day" as prose while `Limits.API_AUTHED` holds the same numbers a module away — and this PR is itself the proof that these numbers move. When `API_AUTHED` next changes, this string silently starts lying to exactly the callers you're trying to convert.

Either interpolate from the constant, or if the prose reads better hand-written, a one-line assertion in the rate-limit tests that `API_AUTHED == "60 per minute; 5000 per day"` makes the next editor notice this string. Same category as the field-path drift guard on #288, and same cheap fix.

Minor placement note while you're here: `_rate_limit_hint` is a pure request-shape predicate with no closure over `register_error_handlers`, so it could sit at module level beside `_wants_json` (error_handler.py:59) rather than nested in the registrar. That's where its sibling helpers live.

Zingzy (commented):
Mergeable, nothing blocking. One should-decide about collateral scope, one nit; the abuse-reduction core is sound and the numbers are right.

**Verified rather than assumed:**
- `SHORTEN_LEGACY` really did guard three routes, and swapping the preview to `PUBLIC_PREVIEW` is a genuine loosening of a read path (100/min bare → 30/min + 2000/day) at the same time the two mint paths tighten. Both mint decorators (`POST /` at url_shortener.py:79, `POST /emoji` at :252) pick up 5/min + 50/day, so the tightening lands where the PR body claims.
- The hint's authenticated-suppression logic is correct and cheap: `rate_limit_key` is pure header/cookie inspection, no DB or JWT verify, so calling it inside an exception handler adds no I/O. A revoked or expired API key resolves to `apikey:` on the limiter key while `get_current_user` returns None, so such a caller is throttled at the anon tier but gets no hint. That's the pre-existing tier/identity split, not something this PR introduces, and the silence is arguably right (they had a key, they know about auth).
- No import cycle from `error_handler` importing `middleware.rate_limiter` — the dependency is one-directional.
- Ran `tests/integration/test_rate_limiting.py` at the head commit in a clean worktree: 16 passed. Full CI matrix is green, so the pre-existing failures mentioned in the body aren't hiding here.
- Cross-checked every number against docs#10: the v1 anon pair, both v0 mint rows, and the legacy stats/export rows all match the constants. That pair agrees.

**Product read:** the funnel-not-wall shape is right — a 429 that names the migration path is worth more than a lower number alone, and it fits the abuse posture (per-IP campaign minting capacity drops ~10x on v0 while a human minting a few links a day never notices). Deploy ordering in the body is also the right way round: ship this, then the docs.

**Merge safety:** no flag gates this, so unlike most changes here it takes effect the moment the release publishes. Nothing to seed, but also no dark period — worth a glance at 429 volume for anonymous v1 after deploy, since the daily cap is the one number with real collateral surface (see inline).

Zingzy (commented):
Follow-up finding from reviewing the docs PR alongside this one — it belongs here rather than there, since the fix is on this side.

**Should-fix: the v0 hint almost never reaches a real v0 caller.** The JSON branch is gated on `_wants_json` (middleware/error_handler.py:59), which needs `application/json` in `Accept`, a JSON content-type, or an `/api/`, `/auth/`, `/oauth/` path. A typical legacy caller posts form-encoded data to `/` or `/emoji` with curl's default `Accept: */*` — that satisfies none of the four, so it takes the HTML branch and gets `_error_html`, which carries no hint at all.

The new `test_rate_limit_hint_on_legacy_shorten` passes `Accept: application/json` explicitly, which is why it's green while the common path is uncovered. The anonymous v1 hint is unaffected (the `/api/` path check always fires there), so this is v0-only — precisely the audience the hint exists for.

Blast radius is soft: nobody breaks, the conversion nudge just doesn't land, and v0 minting is already tiny (~20 req/week on `/emoji`, no successful callers on `POST /`). So should-fix, not blocker.

Options, cheapest first: put the hint in the HTML 429 for the two legacy mint paths; or add a `Retry-After`-adjacent response header (something like `X-Spoo-Hint`) that both branches set, which also reaches callers that ignore the body entirely. If you'd rather not touch the HTML page, that's a legitimate call — but then docs#10's sentence about v0 responses carrying a hint needs narrowing to JSON callers, which I've flagged there.

Either way a test for the no-`Accept` form-post case would stop this from silently regressing.

Zingzy:
Cross-PR note while reviewing #289 and #290 together: both branch off current main and both rewrite the **same lines** of `rate_limit_handler` in `middleware/error_handler.py` — #289 replaces the single `JSONResponse(...)` with a content dict plus the `hint`, #290 replaces it with `_json_error(429, {...})`. Whichever merges second will hit a textual conflict there (the new import lines won't clash, just that one hunk).

The resolution you want is both changes composed rather than either side winning:

```python
content = {"error": "Too many requests", "code": "rate_limit_exceeded"}
if hint := _rate_limit_hint(request):
    content["hint"] = hint
return _json_error(429, content)
```

That keeps the hint in the body while the header still derives from `content["code"]`, so #290's no-drift property holds. Worth resolving deliberately since a careless "take theirs" silently drops the hint, and the #289 hint tests would still pass on the anonymous v1 path if only the legacy branch broke.

Zingzy:
Addressed in 41ce740:

- **v0 hint reach (should-fix):** the hint now also rides an `X-Spoo-Hint` response header on both branches, so form posts with curl's default `Accept` get it despite landing on the HTML page, and it survives edge-composed error bodies too. Added `test_rate_limit_hint_header_on_legacy_form_post` covering the no-Accept form-post path.
- **Stats collateral (should-decide):** checked 14 days of anonymous `GET /api/v1/stats`: 4 IPs exceed 200/day, peaking at 1,054/day, so the shared cap would clip real readers. Split it: `API_ANON_READ` keeps 20/min + 1000/day for stats reads, minting stays at 200/day. Docs PR updated to match.
- **Prose drift (nit):** kept the hand-written prose, added `test_hint_prose_matches_authed_tier` pinning `API_AUTHED` so the string cannot silently lie when the tier moves.
- **Placement (nit):** `_rate_limit_hint` moved to module level beside `_wants_json`.
- **CodeRabbit's validated-auth suggestion:** declining. The hint deliberately mirrors `rate_limit_key`, which is the same unvalidated tiering `dynamic_limit` applies, so an invalid bearer is bucketed on the authed tier and the anonymous hint would name limits that caller does not actually get. Validating tokens in the error handler would also put auth work on the 429 path.
- **#290 overlap:** agreed, whichever merges second should compose the hint dict with `_json_error` as sketched.

Zingzy (commented):
All three addressed, and the two substantive ones are fixed at the right choke points. Nothing blocking; mergeable.

**Verified each fix rather than the commit message:**
- **Collateral scope** — `API_ANON_READ` (rate_limiter.py:37) keeps anonymous `GET /api/v1/stats` at the pre-tightening 1000/day, and `API_ANON` is now reached only by `shorten.py:38`. Splitting the tier is the better half of the option I raised: the read path stops paying for a mint-abuse fix. Nice side effect — a future anonymous endpoint that grabs `API_ANON` inherits the tight budget, so the default is the safe one.
- **Hint reach** — `X-Spoo-Hint` is set on both branches, so the form-post legacy caller now gets it. I traced the edge interaction too: for a `POST` the Caddy `@html` gate fails on method, so `copy_response` preserves the header — the script audience gets it. Under `EDGE_COMPOSED_ERRORS` a browser `GET` 429 does lose it (Caddy replaces the response with the Next page), but that's the human case where the branded page is the message anyway. `test_rate_limit_hint_header_on_legacy_form_post` pins exactly the shape that was broken, asserting `text/html` **and** the header, which is the right assertion pair.
- **Drift nit** — `test_hint_prose_matches_authed_tier` pins `API_AUTHED`, and promoting `_rate_limit_hint` to module scope beside `_wants_json` picked up the placement note.
- Ran `tests/integration/test_rate_limiting.py` at 41ce740 in a clean worktree: 18 passed. Full CI matrix green.
- Cross-checked docs#10 at its new head: the stats row was updated to 1000/day and `X-Spoo-Hint` is documented, so the pair still agrees. I also verified `Retry-After` really is 429-only (`RateLimitHeadersMiddleware` deletes it otherwise), which is what the revised docs now claim.

One observation, no action needed: `X-Spoo-Hint` inherits the same limitation as `X-Error-Code` on #290 — no `Access-Control-Expose-Headers`, so cross-origin JS can't read it. Irrelevant for this header's audience (server-side scripts and curl), just worth knowing it's now a second header waiting on that same follow-up.

Merge ordering still holds: this deploys, then docs#10.

PR #290, opened 2026-08-14
fix(errors): set X-Error-Code header on JSON error responses

Zingzy:
Only HTML errors carried the X-Error-Code header. JSON clients had to parse the response body to get the machine-readable code, so the same error self-described differently depending on content negotiation.

Every negotiated error response now sends X-Error-Code with the exact slug that goes in the body's code field. A single _json_error() helper derives the header from the body, so the two cannot drift, including the EMAIL_NOT_VERIFIED casing and the http_<status> fallback slugs. The body shape is now a typed ErrorBody (TypedDict with required code), so a future call site that omits code fails where the body is built instead of raising KeyError inside an exception handler.

Covered branches:

- AppError handler
- RequestValidationError and PydanticValidationError (422)
- slowapi RateLimitExceeded (429)
- StarletteHTTPException fallback (merges with exc.headers)
- unhandled exception handler (500)
- the 413 body-size rejection in MaxContentLengthMiddleware, which short-circuits before the exception handlers run

Edge interaction, kept deliberately: with EDGE_COMPOSED_ERRORS on, Caddy's @err matcher keys on status plus the X-Error-Code header. JSON errors could never match before this PR; now they can. The inner @html gate still requires Accept: text/html on GET/HEAD, so real API clients (curl, SDKs, fetch) keep raw JSON. The only behavior change is a browser navigating directly to an API URL: an intercepted status (404, 410, 429, 451, 500) now renders the branded error page instead of raw JSON, which is the better outcome for a human at the address bar. Noted in the cutover checklist.

Not changed: response bodies, status codes, the HTML template path, and legacy route error shapes (no code field to mirror). One existing test pinned the header's absence on the JSON path; it now asserts the slug instead, and the body remains byte-identical.

Note: SplitCORSMiddleware has no Access-Control-Expose-Headers policy, so cross-origin browser JS still cannot read this header (same as X-Request-ID today). Until that lands, the header helps server-side clients and the edge; the dashboard still reads code from the body. Left out of this PR since it is a new mechanism rather than an addition to an existing set.

### DIFF
--- a/errors.py
+++ b/errors.py
@@ -6,25 +6,29 @@ AppError is the base for all typed errors.
 
 from __future__ import annotations
 
-from typing import Any, TypedDict
+from typing import Any
 
-from typing_extensions import Required
+from typing_extensions import NotRequired, TypedDict
 
-
-class ErrorBody(TypedDict, total=False):
-    """Wire shape of a JSON error body.
-
-    ``code`` is the machine-readable slug that also feeds the
-    ``X-Error-Code`` response header, so a body without it is a type
-    error at the call site rather than a KeyError inside an exception
-    handler.
-    """
-
-    error: Required[str]
-    code: Required[str]
-    field: str
-    details: Any
-    message: str
+# Wire shape of a JSON error body. ``code`` is the machine-readable slug
+# that also feeds the ``X-Error-Code`` response header, so a body without
+# it is a type error at the call site rather than a KeyError inside an
+# exception handler.
+#
+# Functional syntax on purpose: under ``from __future__ import annotations``
+# the class form stringifies its annotations and the (Not)Required markers
+# are silently dropped at class creation — ``__required_keys__`` comes out
+# empty. Dict-literal values are real objects, so the markers survive.
+ErrorBody = TypedDict(  # noqa: UP013 — class syntax loses the markers here
+    "ErrorBody",
+    {
+        "error": str,
+        "code": str,
+        "field": NotRequired[str],
+        "details": NotRequired[Any],
+        "message": NotRequired[str],
+    },
+)
 
 
 class AppError(Exception):
@@ -204,7 +208,7 @@ class CloudflareAPIError(AppError):
     status_code = 502
     error_code = "cloudflare_api_error"
 
-    def to_dict(self) -> dict:
+    def to_dict(self) -> ErrorBody:
         return {
             "error": (
                 "Upstream certificate service is temporarily unavailable. "
--- a/errors.py
+++ b/errors.py
@@ -6,7 +6,25 @@ AppError is the base for all typed errors.
 
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, TypedDict
+
+from typing_extensions import Required
+
+
+class ErrorBody(TypedDict, total=False):
+    """Wire shape of a JSON error body.
+
+    ``code`` is the machine-readable slug that also feeds the
+    ``X-Error-Code`` response header, so a body without it is a type
+    error at the call site rather than a KeyError inside an exception
+    handler.
+    """
+
+    error: Required[str]
+    code: Required[str]
+    field: str
+    details: Any
+    message: str
 
 
 class AppError(Exception):
@@ -27,8 +45,8 @@ class AppError(Exception):
         self.field = field
         self.details = details
 
-    def to_dict(self) -> dict:
-        payload: dict = {"error": self.message, "code": self.error_code}
+    def to_dict(self) -> ErrorBody:
+        payload: ErrorBody = {"error": self.message, "code": self.error_code}
         if self.field is not None:
             payload["field"] = self.field
         if self.details is not None:
@@ -68,7 +86,7 @@ class EmailNotVerifiedError(ForbiddenError):
 
     error_code = "EMAIL_NOT_VERIFIED"
 
-    def to_dict(self) -> dict:
+    def to_dict(self) -> ErrorBody:
         d = super().to_dict()
         d["message"] = (
             "You must verify your email address before creating resources. "
--- a/middleware/error_handler.py
+++ b/middleware/error_handler.py
@@ -82,6 +82,22 @@ REDIRECT_EDGE_INTERCEPTED_STATUSES = {404, 410, 451}
 _HTTP_STATUS_SLUGS = {404: "not_found", 405: "method_not_allowed"}
 
 
+def _json_error(
+    status_code: int, content: dict, headers: dict | None = None
+) -> JSONResponse:
+    """Build a JSON error response that self-describes via ``X-Error-Code``.
+
+    The header always carries the exact slug from the body's ``code`` field,
+    so clients and the edge can route on the header without parsing the body
+    — same contract HTML errors already honour via ``_error_html``.
+    """
+    return JSONResponse(
+        status_code=status_code,
+        content=content,
+        headers={**(headers or {}), "X-Error-Code": content["code"]},
+    )
+
+
 def _error_html(
     request: Request, status_code: int, message: str, slug: str
 ) -> Response:
@@ -122,7 +138,7 @@ def register_error_handlers(app: FastAPI) -> None:
     @app.exception_handler(AppError)
     async def app_error_handler(request: Request, exc: AppError) -> Response:
         if _wants_json(request):
-            return JSONResponse(status_code=exc.status_code, content=exc.to_dict())
+            return _json_error(exc.status_code, exc.to_dict())
         return _error_html(
             request, exc.status_code, exc.message.upper(), exc.error_code
         )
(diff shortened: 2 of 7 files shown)


Zingzy (commented):
Mergeable, nothing blocking. One should-decide (an edge-behavior change the PR body says doesn't happen), one nit, and a merge-order note about #289.

**Verified:**
- `content["code"]` is safe at all six call sites: `AppError.to_dict` always emits `code`, both overrides (`EmailNotVerifiedError`, `CloudflareApiError`) keep it, and both `_validation_error_response` branches set `validation_error`. The EMAIL_NOT_VERIFIED casing rides through unchanged, which is the right call since the body already published that slug.
- The `{**(headers or {}), "X-Error-Code": ...}` merge preserves `exc.headers` (a 401's `WWW-Authenticate` survives) and tolerates `None`.
- 413 sits outside `EDGE_INTERCEPTED_STATUSES`, so the `MaxContentLengthMiddleware` addition can't be edge-composed. Adding the header inline there is right — it short-circuits before the handlers, exactly as the comment says.
- Ran the three affected integration files at the head commit in a clean worktree: 60 passed. Full CI matrix green.
- Your CORS caveat is accurate: there is no `expose_headers` anywhere in the repo, so cross-origin JS genuinely can't read this. Deferring it is the right scope call. Worth being explicit about the consequence though: until that lands, the "don't parse the body" win applies to server-side clients and the edge only — the dashboard still has to read `code` out of the body.

**Architecture read:** deriving the header from `content["code"]` inside one helper is the correct choke point — it makes drift structurally impossible rather than test-enforced, which is a better answer than the parallel-copy pattern used elsewhere in this file. Converting the existing test from pinning the header's absence to asserting the slug, with the body byte-identical, is the right way to retire an obsolete pin.

middleware/error_handler.py
```diff
@@ -82,6 +82,22 @@ def _wants_json(request: Request) -> bool:
...
+    """Build a JSON error response that self-describes via ``X-Error-Code``.
+
+    The header always carries the exact slug from the body's ``code`` field,
+    so clients and the edge can route on the header without parsing the body
+    — same contract HTML errors already honour via ``_error_html``.
+    """
+    return JSONResponse(
+        status_code=status_code,
+        content=content,
+        headers={**(headers or {}), "X-Error-Code": content["code"]},
```
Zingzy, line 116: **Should-decide: this does change the edge-composed flow, contrary to the PR body.** Caddy's `@err{status}` matcher is `status {N}` **plus** `header X-Error-Code *` (caddy/Caddyfile:77-80) for N in {404, 410, 429, 451, 500}. Until now, JSON errors never carried the header, so they could never match. After this PR they always do.

The inner `@html{N}` gate is `Accept: *text/html*` and `GET`/`HEAD`. That's satisfiable on a JSON error, because `_wants_json` returns True for any `/api/`, `/auth/`, `/oauth/` path **regardless of Accept**. So: a browser navigating to `https://spoo.me/api/v1/<nonexistent>` gets a JSON 404 from the app, which Caddy now intercepts and replaces with the Next branded error page (status preserved via `replace_status`). Same for a browser-navigated anonymous 429 on an API GET.

Real API clients are unaffected — curl and every SDK send `*/*` or `application/json`, which fails the `@html` matcher and falls to `copy_response`. So blast radius is browser-navigated API URLs only, and the new behavior is arguably nicer for the human who pasted an endpoint into the address bar. I'd probably keep it.

What I'd change is the claim: the body's "Not changed: … the HTML/edge-composed flow" is what a future bisector will read, and it's inaccurate. Since `EDGE_COMPOSED_ERRORS` is the flag gating this at cutover, please note the interaction in the body (and ideally in the prod-cutover checklist) so it's a decision on the record rather than a surprise.

Minor asymmetry while you're in here: `_error_html` deliberately skips rendering under `EDGE_COMPOSED_ERRORS` because Caddy throws the body away, but the JSON path now serializes a body that gets discarded in the same situation. It's a few microseconds on an error path, not worth code — just noting the two paths no longer reason about composition the same way.

middleware/error_handler.py
```diff
@@ -82,6 +82,22 @@ def _wants_json(request: Request) -> bool:
 _HTTP_STATUS_SLUGS = {404: "not_found", 405: "method_not_allowed"}
 
 
+def _json_error(
```
Zingzy, line 104: **Nit (failure mode):** `content["code"]` raises `KeyError` if a future call site passes a body without `code`, and because this runs *inside* an exception handler the result is a bare 500 that masks the original error — the worst place to discover the mistake. All six current sites are safe (I checked every `to_dict` and both validation branches), so this is about the next one.

Strict indexing is defensible: it fails loudly and a test would catch it. If you want the failure to stay loud without being destructive, typing the parameter as a `TypedDict` with a required `code` moves it to a type error at the call site instead of a runtime error on a live request. Either is fine; the status quo is the only one where the symptom is a mystery 500.

Zingzy:
Cross-PR note while reviewing #289 and #290 together: both branch off current main and both rewrite the **same lines** of `rate_limit_handler` in `middleware/error_handler.py` — #289 replaces the single `JSONResponse(...)` with a content dict plus the `hint`, #290 replaces it with `_json_error(429, {...})`. Whichever merges second will hit a textual conflict there (the new import lines won't clash, just that one hunk).

The resolution you want is both changes composed rather than either side winning:

```python
content = {"error": "Too many requests", "code": "rate_limit_exceeded"}
if hint := _rate_limit_hint(request):
    content["hint"] = hint
return _json_error(429, content)
```

That keeps the hint in the body while the header still derives from `content["code"]`, so #290's no-drift property holds. Worth resolving deliberately since a careless "take theirs" silently drops the hint, and the #289 hint tests would still pass on the anonymous v1 path if only the legacy branch broke.

middleware/error_handler.py
Zingzy: Kept the behavior, agreed it reads better for a human at the address bar. Fixed the PR body (the edge flow claim was wrong as written) and added a verification step to the cutover checklist for the flip: browser-open a dead /api/v1 URL expecting the branded 404, same URL via curl expecting JSON. The render asymmetry is noted, leaving it: a serialized-then-discarded JSON body on an error path is not worth a second composition branch.

middleware/error_handler.py
Zingzy: Took the TypedDict route in 03cb5f3: ErrorBody lives in errors.py next to the taxonomy (required error and code, optional field, details, message), to_dict returns it, and _json_error and _validation_error_response are typed against it. Strict indexing stays as the runtime backstop since nothing in CI runs a type checker yet.

Zingzy:
On the #289 collision: agreed on composing rather than either side winning. If #289 lands first I will rebase this branch and fold the hint dict into _json_error(429, content) exactly as sketched; if this lands first, #289's rebase should do the same. Either way the header keeps deriving from content["code"] so nothing drifts.

errors.py
```diff
@@ -6,7 +6,25 @@
 
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, TypedDict
```
Zingzy: **Nit (the construct doesn't do what it says):** mixing `typing.TypedDict` here with `typing_extensions.Required` on line 11 means the `Required` markers are ignored at class creation. Verified in the project venv at this commit:

```
py (3, 13)
required: []
optional: ['code', 'details', 'error', 'field', 'message']
```

`__required_keys__` is empty and every key is optional at runtime, including `code`. `typing_extensions` documents that `Required`/`NotRequired` need `typing_extensions.TypedDict` on Python < 3.11, and this project supports 3.10.

No impact today — nothing introspects `__required_keys__`, and mypy would still read the annotation correctly if it ran. It matters in two future scenarios: if an error body ever gets validated through this type (pydantic reads `__required_keys__`, so it would silently accept a body with no `code`), and as a trap for whoever later assumes the marker is live. One-line fix: import `TypedDict` from `typing_extensions` alongside `Required`.

Agreed on keeping the strict `content["code"]` indexing as the runtime backstop — with no checker in CI and the markers currently inert, that indexing is doing all the real enforcement.

Zingzy (commented):
Both findings addressed, and the edge one is resolved the right way — keeping the behavior, correcting the body claim, and putting a browser-vs-curl check in the cutover checklist beats changing code for it. `ErrorBody` living in `errors.py` next to the taxonomy is the right home too. Re-ran the four affected integration files at 03cb5f3 in a clean worktree: 72 passed, CI matrix green.

Two leftovers from the TypedDict change, neither blocking.

**Should-fix (small, missed by the refactor):** `CloudflareAPIError.to_dict` at errors.py:207 still returns bare `dict`. It's the one body in the file built by hand instead of derived from `super().to_dict()`, so it's precisely the call site the type gate was meant to cover — the only one that could drop `code` without a checker noticing. It does include `code` today (verified by instantiating it), so no runtime bug, but widening the return type relative to the base is an incompatible override, which becomes a hard error the day mypy lands. One-word fix: `-> ErrorBody`. (Couldn't attach this inline — line 207 isn't in the diff.)

The second one is inline, and it's the interesting one: I probed the new type at runtime and the `Required` markers aren't actually taking effect.

errors.py
Zingzy: Good catch, but the diagnosis is one layer off: the culprit is the module's from __future__ import annotations, not the import origin. PEP 563 stringifies class-syntax annotations, so the markers get dropped at class creation no matter which TypedDict you use. Verified in the venv: typing_extensions.TypedDict with class syntax under the future import still gives required: [].

Fixed properly in 735bf29 with the functional form, whose dict-literal values are never stringified: required is now ['code', 'error'], optional ['details', 'field', 'message']. Flipped polarity to NotRequired on the optionals since total defaults to true. Ruff's UP013 wanted to rewrite it back to class syntax, so it carries a noqa with the reason. CloudflareAPIError.to_dict is typed as ErrorBody in the same commit.

PR #291, opened 2026-08-14
feat(stats): per-link stats and export endpoints, url_id filter

Zingzy:
Adds the per-link analytics surface and an id-based filter, all additive:

- `GET /api/v1/stats/links/{url_id}`: click statistics for one owned URL. Resolves the link first (auth required, foreign ids answer like missing ones), matches clicks on `meta.url_id`, and supports the full grouping surface including the utm dimensions.
- `GET /api/v1/export/links/{url_id}`: export twin. Filenames derive from the alias, with a url_id fallback for non-ASCII aliases since Content-Disposition is latin-1 on the wire.
- `url_id` as a multi-value filter on `GET /api/v1/stats` and `GET /api/v1/export`. Values are validated as ObjectIds; ownership needs no extra check because the owner stamp already scopes the match, so foreign ids simply match nothing.

The `links` segment is typed so future per-entity analytics (domains, groups) can land beside it without route shadowing.

No behavior changes to existing endpoints. 2985 tests pass.

## Summary by CodeRabbit

- **New Features**
  - Added per-link statistics with filtering, grouping, date ranges, and UTM dimensions.
  - Added per-link exports in CSV and other supported formats.
  - Added URL ID filtering for aggregate statistics and exports.
  - Exported files now use safe, descriptive filenames based on link aliases.

- **Documentation**
  - Expanded API documentation for aggregate and per-link statistics and export endpoints.
  - Added guidance on selecting the appropriate endpoint for single-link requests.

### DIFF
--- a/schemas/dto/requests/stats.py
+++ b/schemas/dto/requests/stats.py
@@ -1,18 +1,24 @@
 """
 Request DTOs for statistics and export endpoints.
 
-StatsQuery   — GET /api/v1/stats   (query parameters)
-ExportQuery  — GET /api/v1/export  (query parameters; superset of StatsQuery)
+StatsQuery       — GET /api/v1/stats               (query parameters)
+ExportQuery      — GET /api/v1/export              (superset: adds ``format``)
+LinkStatsQuery   — GET /api/v1/stats/links/{id}    (no link-identity params)
+LinkExportQuery  — GET /api/v1/export/links/{id}   (superset: adds ``format``)
 
-Both models parse comma-separated strings for multi-value fields and validate
+All models parse comma-separated strings for multi-value fields and validate
 IANA timezone names.  The JSON ``filters`` string is parsed into a typed dict.
+The per-link variants drop ``scope``/``short_code``/``url_id`` entirely —
+the path already names the link, so slicing or bucketing by link identity
+is meaningless there.
 """
 
 from __future__ import annotations
 
 import json
-from typing import Any, Literal
+from typing import Any, ClassVar, Literal
 
+from bson import ObjectId
 from pydantic import (
     Field,
     PrivateAttr,
@@ -22,6 +28,8 @@ from pydantic import (
 
 from schemas.dto.base import RequestBase
 from schemas.dto.requests._descriptions import (
+    LINK_STATS_FILTERS_DESC,
+    LINK_STATS_GROUP_BY_DESC,
     STATS_BROWSER_DESC,
     STATS_CITY_DESC,
     STATS_COUNTRY_DESC,
@@ -36,6 +44,7 @@ from schemas.dto.requests._descriptions import (
     STATS_SHORT_CODE_DESC,
     STATS_START_DATE_DESC,
     STATS_TIMEZONE_DESC,
+    STATS_URL_ID_DESC,
     STATS_UTM_DESC,
 )
 from schemas.enums.stats import (
@@ -43,6 +52,8 @@ from schemas.enums.stats import (
     ALLOWED_FILTERS,
     ALLOWED_GROUP_BY,
     ALLOWED_METRICS,
+    LINK_ALLOWED_FILTERS,
+    LINK_ALLOWED_GROUP_BY,
     ExportFormat,
     StatsDimension,
     StatsScope,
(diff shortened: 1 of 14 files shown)


services/stats_service.py
```diff
@@ -200,6 +260,11 @@ def _build_click_query(
                     )
                     continue
                 query["meta.short_code"] = {"$in": values}
+            elif dimension == StatsDimension.URL_ID:
```
Zingzy, line 263: **Should-fix (latent scope bypass in a deliberately shared function):** the `short_code` branch immediately above guards against overwriting an already-locked value, and logs `query_builder_scope_bypass_prevented` when it happens. The `url_id` branch has no equivalent guard — and that asymmetry matters more here than it did for `short_code`.

In `query_link` (line 731) the match is built as `{"meta.url_id": url.id, "clicked_at": {...}}` with **no** `meta.owner_id` arm. The route's `get_owned` is what establishes tenancy, and the `meta.url_id` equality is the only thing carrying that decision into the query. So if a `url_id` filter ever reached this function on the per-link path, `query["meta.url_id"] = {"$in": [...]}` would overwrite the lock and the resulting query would have **no ownership constraint at all** — a cross-tenant read of anyone's click history, since `url_id` filter values are deliberately not ownership-checked.

It is not reachable today. I verified `LINK_ALLOWED_FILTERS` excludes both identity dimensions and that the ClassVar override is honoured, so a `url_id` filter on the per-link endpoint is silently dropped by the DTO. This is one allowlist edit away from being live, in a function whose docstring advertises that it is shared by both callers, and the codebase already contains the exact guard pattern one branch up.

Mirror it:

```python
elif dimension == StatsDimension.URL_ID:
    if "meta.url_id" in query:
        log.warning(
            "query_builder_scope_bypass_prevented",
            dimension="url_id",
            attempted_values=values,
        )
        continue
    query["meta.url_id"] = {"$in": [ObjectId(v) for v in values]}
```

If you want the stronger version, adding `meta.owner_id` to `query_link`'s match makes tenancy explicit in the query rather than inherited from the route — but check the index first, since the existing `(meta.url_id, clicked_at)` index is what makes this query cheap. The guard alone closes the hole.

services/stats_service.py
```diff
@@ -134,6 +135,49 @@ def _fmt_tz(dt: datetime | None, tz_name: str) -> datetime | None:
...
+    # ── Private: date window ──────────────────────────────────────────────────
+
+    def _resolve_window(
+        self,
+        start_raw: str | None,
+        end_raw: str | None,
+    ) -> tuple[datetime, datetime]:
+        """Parse, default, cap, and validate the date window.
+
+        Shared by ``query()`` and ``query_link()``; mirrored by
```
Zingzy, line 147: **Nit (drift guard):** "the three must not drift" is doing real work here — `PublicStatsService._resolve_window` is a genuine second implementation, not a wrapper. I diffed them: identical today (7-day default, both-null and one-null branches, future-capping, `start > end`, and the max-range check), so no bug.

But a comment is the weakest possible enforcement for a rule about two copies of date-window semantics, and the failure mode is quiet — public and owner views of the same link disagreeing about what "last week" means, with every test still green.

A parametrized test that runs both implementations over the same handful of inputs (both-null, start-only, end-only, future dates, inverted range) and asserts equal output would pin it for the price of about fifteen lines. Same shape as the group-by/filter agreement test that already guards the sentinel pairs.

Zingzy (commented):
Mergeable, nothing blocking. One should-fix on defence-in-depth inside the shared filter builder, one nit. This is a clean piece of API design — the per-link endpoints are additive, the DTO hierarchy expresses the constraint rather than documenting it, and the security-relevant paths hold up.

**Verified rather than assumed:**
- `get_owned` enforces ownership *in the query* (`find_by_id_and_owner`), so a foreign id is indistinguishable from a missing one. No existence oracle, and the route resolves before any click work happens.
- `query_link` matches `meta.url_id` on the ObjectId, never `meta.short_code`, so the custom-domain bleed the PR body describes really is closed. The claimed-links note checks out too: claim-time reattribution stamps `meta.url_id` on pre-claim clicks, so history is complete without a second arm.
- The `url_id` filter converts to `ObjectId` before the `$in` (stats_service.py:267). Worth calling out because the obvious version of this feature passes strings and silently matches nothing forever — `meta.url_id` is stored as an ObjectId.
- The per-link allowlist override is real, not decorative: the base validator reads `self._ALLOWED_GROUP_BY` / `self._ALLOWED_FILTERS`, so `LinkStatsQuery`'s ClassVars genuinely reject link-identity dimensions. I checked this specifically because if the validator had closed over the module-level constants instead, the override would be dead and `url_id` would be accepted on the per-link endpoint — see the inline comment for why that would matter.
- Splitting `_ExportFormatMixin` out instead of subclassing `ExportQuery` twice is the right shape, and `ALLOWED_GROUP_BY = frozenset(StatsDimension) - {URL_ID}` keeps the filter-only dimension from leaking into grouping by construction.
- The latin-1 filename fallback is a real constraint, not defensive noise: an emoji alias in `Content-Disposition` would be a wire error, and falling back to the id keeps the download working.
- Ran the stats/export unit and integration files at the stacked head: 142 passed. Full CI matrix green.

**Blast radius:** authed read endpoints at small N, plus one new filter branch on the account aggregate. Nothing touches the redirect path. The `(meta.url_id, clicked_at)` index already exists, so the per-link query is served.

schemas/dto/requests/stats.py
```diff
@@ -234,29 +235,101 @@ def _parse_multi_value_fields(self) -> StatsQuery:
...
+        # the service can convert to ObjectId without a second check. No
+        # ownership check — the owner stamp already scopes the $match, so
+        # foreign ids simply match nothing.
+        invalid = [
+            v
+            for v in parsed_filters.get(StatsDimension.URL_ID, [])
+            if not ObjectId.is_valid(v)
+        ]
+        if invalid:
+            raise ValueError(f"invalid url_id filter values: {', '.join(invalid)}")
```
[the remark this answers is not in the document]
Zingzy: Fixed in 14720a2: url_id filters are rejected with a 422 when scope=anon, covering both the param and the JSON filters path, with tests for each. The rejection retires again in #292 where the scope param itself is removed and every query carries the owner stamp.

services/stats_service.py
Zingzy: Added in 14720a2: the url_id branch now wears the same guard as short_code, skipping and logging query_builder_scope_bypass_prevented when the query already locks meta.url_id, plus a unit test pinning that a locked id cannot be overwritten. Went with the guard alone rather than adding meta.owner_id to query_link's match, to keep the (meta.url_id, clicked_at) index doing the work.

services/stats_service.py
Zingzy: Added in 14720a2: TestResolveWindowParity runs both implementations over explicit, defaulted, future-capped, inverted, oversized, and malformed inputs, asserting equal windows (within now-sampling skew for the defaulted cases) and matching rejections.

PR #292, opened 2026-08-14
feat(stats)!: drop the scope param, stats and export require auth

Zingzy:
Breaking change, stacked on #291. Merge only after the frontend and client migrations ship.

`GET /api/v1/stats` and `GET /api/v1/export` now require authentication and are always scoped to the caller's own URLs. The `scope` parameter is removed; a stray `scope=` in the query string is silently ignored (pinned in tests). `short_code` remains available as a plain filter.

What replaces the old anonymous mode:

- Public per-link stats live at `GET /api/v1/public/stats/{code}`, which has the correct semantics the anonymous scope lacked: privacy answers as a plain 404, passwords travel in a POST body, clicks are matched by url id so same-alias links on other domains cannot bleed in, and utm dimensions stay private to the owner.
- Owner per-link stats live at `GET /api/v1/stats/links/{url_id}` from #291.

Also removed: the alias-based stats privacy check (`check_stats_privacy`), the anonymous export tier, and the anonymous read tier introduced in #289, whose only consumer was the anonymous stats read.

90 days of traffic were reviewed before removing this without a deprecation window: anonymous scope usage came from a single automated client, and the export endpoint had zero requests in any scope. Expect that client to start receiving 401s.

2983 tests pass.

### DIFF
--- a/schemas/dto/requests/stats.py
+++ b/schemas/dto/requests/stats.py
@@ -8,9 +8,13 @@ LinkExportQuery  — GET /api/v1/export/links/{id}   (superset: adds ``format``)
 
 All models parse comma-separated strings for multi-value fields and validate
 IANA timezone names.  The JSON ``filters`` string is parsed into a typed dict.
-The per-link variants drop ``scope``/``short_code``/``url_id`` entirely —
-the path already names the link, so slicing or bucketing by link identity
-is meaningless there.
+The per-link variants drop ``short_code``/``url_id`` entirely — the path
+already names the link, so slicing or bucketing by link identity is
+meaningless there.
+
+There is no ``scope`` parameter: every endpoint in this family requires
+auth and reads the caller's own aggregate. A stray ``scope=`` from an old
+client is silently ignored (RequestBase drops unknown params).
 """
 
 from __future__ import annotations
@@ -40,7 +44,6 @@ from schemas.dto.requests._descriptions import (
     STATS_METRICS_DESC,
     STATS_OS_DESC,
     STATS_REFERRER_DESC,
-    STATS_SCOPE_DESC,
     STATS_SHORT_CODE_DESC,
     STATS_START_DATE_DESC,
     STATS_TIMEZONE_DESC,
@@ -56,7 +59,6 @@ from schemas.enums.stats import (
     LINK_ALLOWED_GROUP_BY,
     ExportFormat,
     StatsDimension,
-    StatsScope,
 )
 
 
@@ -260,9 +262,6 @@ class StatsQuery(_StatsQueryBase):
     The ``filters`` parameter accepts a JSON object string.
     """
 
-    scope: Literal[StatsScope.ALL, StatsScope.ANON] = Field(
-        default=StatsScope.ALL, description=STATS_SCOPE_DESC
-    )
     short_code: str | None = Field(
         default=None,
         max_length=50,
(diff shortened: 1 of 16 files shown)


schemas/dto/requests/stats.py
```diff
@@ -260,9 +262,6 @@ class StatsQuery(_StatsQueryBase):
     The ``filters`` parameter accepts a JSON object string.
     """
 
-    scope: Literal[StatsScope.ALL, StatsScope.ANON] = Field(
-        default=StatsScope.ALL, description=STATS_SCOPE_DESC
-    )
     short_code: str | None = Field(
         default=None,
         max_length=50,
```
Zingzy: **Nit (newly-misleading cap):** the description this PR gives `short_code` now opens with "Comma-separated URL aliases to filter stats to specific URLs you own", but `max_length` is still **50** — while its new sibling `url_id` gets `max_length=1000` for exactly the same comma-separated-list job, and so does every other multi-value filter param on this DTO.

Aliases run up to 50 characters individually, so at this cap the documented list form realistically holds one alias, maybe two. A caller who reads the description and passes five gets a 422 that says nothing about why.

The cap and the comma-splitting were both already here — what changed is that the old description was singular ("URL alias to query stats for"), so the mismatch is new with this PR. Raising it to 1000 to match `url_id` costs nothing; the values are format-checked downstream either way.

Zingzy (commented):
The removal itself is clean and correctly stacked, but **do not merge this until spoo-cli ships a migration** — there is a consumer in the wave's blind spot, and unlike the others it is a distributed binary. Detail below; everything else here is a nit.

**The missing consumer: spoo-cli.** `spoo stats <code>` has a first-class anonymous path that this PR breaks:

- `internal/cmd/stats.go:50-56` — when the credential store returns `ErrNotLoggedIn`, it sets `scope = "anon"` and proceeds with the short code. The error text when no code is given even advertises the feature: *"not logged in — pass a short code for public stats, or run `spoo auth login`"*.
- `internal/api/stats.go:27` — `v.Set("scope", q.Scope)`, so the request goes out as `GET /api/v1/stats?scope=anon&short_code=...` with no auth.
- `internal/tui/stats/switcher.go:28` and `model.go:53` — the TUI carries `scope` as state with anon-specific behaviour ("log in to switch links").

After this merges, that request has `scope` silently ignored and auth mandatory, so it becomes a bare **401** on a path the CLI's own copy tells users to use. There is no spoo-cli PR in the wave.

This is worse than the other clients rather than equivalent, because the CLI is distributed through the Homebrew tap and Scoop bucket. "Deploy the client first" does not apply — installed binaries keep sending `scope=anon` indefinitely, so every stale install is permanently broken rather than broken until a page refresh. The migration target is `GET /api/v1/public/stats/{code}`, which is a straightforward swap since the CLI already has the short code in hand.

**On the traffic evidence.** The body says 90 days showed anonymous scope usage from "a single automated client". Worth re-checking that conclusion against the CLI specifically: you shipped `X-Spoo-Client` telemetry across the client repos, so the anon-scope requests can be attributed by client slug rather than inferred. If that single client *is* spoo-cli, the finding above is confirmed from the other direction; if the CLI's anon usage is genuinely zero, then it is a latent break rather than a live one and you may reasonably decide to ship anyway — but that should be a decision, not an assumption.

**Cheap mitigation regardless:** since `scope=` is now silently ignored, a request arriving with `scope=anon` and no credentials is unambiguously a stale client. Returning the 401 with a `hint` naming `GET /api/v1/public/stats/{code}` would convert an opaque failure into a self-service fix for every stale client at once, not just the CLI — exactly the pattern you just merged for 429s in #289.

**Verified correct:**
- Stacking is right: base is `feat/stats-links-endpoints`, and #289 (which introduced `API_ANON_READ`) merged at 19:55 today, so the tier this PR deletes genuinely exists on the base. `API_ANON_READ` and `API_EXPORT_ANON` are both fully removed with no dangling references.
- `scope` is gone as a request parameter, and keeping `StatsScope` for the response wire with a docstring that says exactly why is the right call — the public endpoint's frozen contract still emits `anon`.
- `check_stats_privacy` and `StatsPrivacyInfo` are removed together; the privacy semantics genuinely moved to the public endpoint, which answers 404 rather than leaking existence.
- The index comments were updated to match the new vocabulary instead of being left describing a parameter that no longer exists — easy to forget, and it is the kind of stale comment that misleads six months later.
- Ran the stats/export unit and integration files at this head: 142 passed. Full CI matrix green.
- An independent audit of docs#11 against this branch confirmed every stats/export path, parameter, group-by enum, auth requirement, and response field in the regenerated spec matches this code exactly — including that the corrected `start_date` default wording landed.

schemas/dto/requests/stats.py
Zingzy: Fixed in bc0409e: raised to 1000 to match url_id and the other multi-value params, with a comment noting it is sized for the documented list form.

PR #294, opened 2026-08-16
fix: document the /health response schema in OpenAPI

Zingzy:
`GET /health` was annotated `-> JSONResponse`, so openapi.json listed the path with no response schema at all.

The handler now returns the existing `HealthResponse` DTO, with the 503 status set on the injected `Response` when MongoDB is unreachable. The DTO was reconciled to what the endpoint actually returns: it gains the `version` field and its `checks` field now uses the already-defined (but previously unused) `HealthChecks` model instead of a bare dict. The wire format is byte-for-byte unchanged, which the existing integration tests confirm.

Both the 200 and 503 responses now reference the `HealthResponse` schema in the spec, and a new smoke test pins that so the schema cannot silently disappear again.

## Summary by CodeRabbit

* **New Features**
  * Health responses now include the application version and structured MongoDB and Redis check results.
  * The health endpoint reports unavailable services with a 503 status while healthy responses remain 200.
  * API documentation now includes schemas for health responses and both success and unavailable outcomes.
  * Statistics alias filters now support up to 1,000 characters.

* **Tests**
  * Added coverage to verify health documentation and version data in serialized responses.
  * Added automated checks to detect API specification drift.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -28,7 +28,19 @@
             "description": "Successful Response",
             "content": {
               "application/json": {
-                "schema": {}
+                "schema": {
+                  "$ref": "#/components/schemas/HealthResponse"
+                }
+              }
+            }
+          },
+          "503": {
+            "description": "Unhealthy \u2014 MongoDB is unreachable",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/HealthResponse"
+                }
               }
             }
           }
@@ -3762,7 +3774,7 @@
               "anyOf": [
                 {
                   "type": "string",
-                  "maxLength": 50
+                  "maxLength": 1000
                 },
                 {
                   "type": "null"
@@ -5001,7 +5013,7 @@
               "anyOf": [
                 {
                   "type": "string",
-                  "maxLength": 50
+                  "maxLength": 1000
                 },
                 {
                   "type": "null"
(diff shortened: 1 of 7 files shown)


openapi.json
```diff
@@ -14195,7 +14249,7 @@
   ],
   "servers": [
     {
-      "url": "http://spoo.local:8000",
+      "url": "https://spoo.me",
```
Zingzy, line 14262: Should-decide: this regen swept in two hunks unrelated to the health work: the `short_code` maxLength going 50 to 1000 (the code has said 1000 for a while) and this servers URL. That means the committed spec had silently drifted from the code and nothing caught it. The new smoke test pins the live app spec, not this artifact, so the file can drift again the same way.

Worth a CI step that diffs `make openapi` output against the committed file. If you add it, pin `APP_URL` inside the target or the workflow env: the servers block comes from `settings.app_url`, so the output currently depends on whoever ran the regen last, which is exactly how `http://spoo.local:8000` ended up committed as "Production".

Zingzy (commented):
Mergeable, nothing blocking. Traced the handler end to end: both check keys are always populated (`mongodb` and `redis`, including the `not_configured` value when Redis is absent), so the typed `HealthChecks` model cannot fail response validation on Redis-less self-host deploys. The `Response` injection plus `responses={503: ...}` is the standard FastAPI shape for a status override with a response model, and the wire format claim holds: ran `tests/integration/test_health.py` plus the smoke and unit suites against the branch, 14 passed.

One should-decide inline about the regenerated `openapi.json`, and one optional nit on the DTO typing. Merge ships nothing on its own (deploys fire on release publish), and this touches no flag-gated behavior, so there is no enable-time tail here either.

schemas/dto/responses/common.py
```diff
@@ -33,7 +33,8 @@ class HealthResponse(ResponseBase):
     """Response body for GET /health."""
 
     status: str
-    checks: dict[str, str]
+    version: str
+    checks: HealthChecks
```
Zingzy, line 37: Nit, optional: `mongodb` and `redis` only ever take `"ok"`, `"error"`, or `"not_configured"`. Typing them as a `Literal` of those three would put the possible values into the OpenAPI schema for free, which is the whole point of this PR. Fine as-is.

openapi.json
Zingzy: Took the CI suggestion in 95e2dcc: a spec-drift job in tests.yaml regenerates the spec and fails on any diff against the committed file. The make target now pins both APP_URL and MONGODB_URI (your spoo.local diagnosis was right, output depended on the invoker's env) and writes through a temp file, since a failed export used to truncate the committed spec via the shell redirect. Kept the two swept-in hunks: they are the code's actual truth and the new job keeps the file honest from here.

schemas/dto/responses/common.py
Zingzy: Done in 95e2dcc, both fields are Literal["ok", "error", "not_configured"] and the enum shows up in the regenerated schema.

PR #295, opened 2026-08-16
feat: expose diagnostic headers on CORS responses

Zingzy:
`SplitCORSMiddleware` never sent `Access-Control-Expose-Headers`, so cross-origin JavaScript could not read any of our diagnostic response headers: `X-Request-ID`, `X-Error-Code`, `X-Spoo-Hint`, the `X-RateLimit-*` trio, or `Retry-After`. Browsers only let scripts read the CORS-safelisted set (Cache-Control, Content-Language, Content-Length, Content-Type, Expires, Last-Modified, Pragma), and `Retry-After` is not on it, so it needs listing explicitly too.

The header is now set on both policies: public routes (wildcard origin) and private routes with an allowlisted origin. Non-CORS requests and disallowed private origins get no CORS headers, same as before.

New integration tests assert the exact exposed list on a public path and an allowlisted private path, and its absence without an Origin header and for a disallowed private origin.

### DIFF
--- a/tests/integration/test_middleware.py
+++ b/tests/integration/test_middleware.py
@@ -245,6 +245,58 @@ def test_cors_private_route_disallowed_origin():
     assert "access-control-allow-origin" not in resp.headers
 
 
+def test_cors_public_route_exposes_headers():
+    app = _build_test_app(include_logging=False)
+    app.include_router(_test_router)
+    with TestClient(app) as c:
+        resp = c.get(
+            "/api/v1/test-404",
+            headers={"Origin": "https://example.com"},
+        )
+    assert resp.headers.get("access-control-expose-headers") == (
+        "X-Request-ID, X-Error-Code, X-Spoo-Hint, X-RateLimit-Limit, "
+        "X-RateLimit-Remaining, X-RateLimit-Reset, Retry-After"
+    )
+
+
+def test_cors_private_route_exposes_headers():
+    app = _build_test_app(include_logging=False)
+    from middleware.security import SplitCORSMiddleware
+
+    for mw in app.user_middleware:
+        if mw.cls is SplitCORSMiddleware:
+            mw.kwargs["private_origins"] = ["https://spoo.me"]
+    app.include_router(_test_router)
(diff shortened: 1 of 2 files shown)


tests/integration/test_middleware.py
```diff
@@ -245,6 +245,58 @@ def test_cors_private_route_disallowed_origin():
...
 
+def test_cors_public_route_exposes_headers():
+    app = _build_test_app(include_logging=False)
+    app.include_router(_test_router)
+    with TestClient(app) as c:
+        resp = c.get(
+            "/api/v1/test-404",
+            headers={"Origin": "https://example.com"},
+        )
+    assert resp.headers.get("access-control-expose-headers") == (
```
Zingzy, line 256: Asserting the literal string here instead of importing `_EXPOSED_HEADERS` is the right choice, since importing the constant would make the test tautological. Leaving a note so a future cleanup pass does not "deduplicate" this into meaninglessness: if the list changes, these two assertions are supposed to fail.

Zingzy (commented):
Mergeable, nothing blocking. Verified every header in the list has a real emitter: `X-Request-ID` from the logging middleware, `X-Error-Code` and `X-Spoo-Hint` from the error handler, and the `X-RateLimit-*` trio plus `Retry-After` from slowapi (`headers_enabled=True`) with the ASGI injection shim. Listing explicit names instead of `*` is the correct call, and not just taste: on the credentialed private policy a wildcard in `Access-Control-Expose-Headers` is treated as the literal header name `*`, so the explicit list is required there anyway.

Preflight responses also pick up the header via `_set_cors_headers`; browsers ignore it on preflights, so that is harmless. The four new tests cover exactly the right matrix (public, allowlisted private, no Origin, disallowed private) and reuse the existing kwargs-mutation pattern from this file. Ran `tests/integration/test_middleware.py` against the branch, 21 passed.

Product-wise this closes the gap left by #273: the rate-limit headers shipped there were unreadable from cross-origin JS until now, which was most of their audience.

Issue #233, opened 2026-07-09
[FEAT] Scam Prevention Framework


Zingzy:
# Explanation

Design and implement a comprehensive abuse prevention framework for **spoo.me** to reduce phishing, scam, and spam links while preserving the platform's privacy-first philosophy and keeping false positives as low as possible.

## Is your feature request related to a problem? Please describe.

As spoo.me has grown, it has increasingly become a target for abuse by scammers and phishing campaigns. Recently, this has resulted in reputation issues, including warnings from external security providers, which negatively affect legitimate users and the platform as a whole.

While individual malicious links can be blocked manually, this approach does not scale. We need a long-term system that can proactively detect, prevent, and respond to abuse while minimizing the impact on legitimate users.

## Describe the solution you'd like

Create a modular abuse prevention framework that can evolve over time. Some ideas include (but are not limited to):

* Risk scoring for newly created links.
* Domain and destination reputation checks.
* Detection of suspicious URL creation patterns.
* Click velocity and traffic anomaly detection.
* Detection of mass-created links from the same actor.
* Automated quarantine or review for high-risk links.
* Community reporting tools.
* Better internal moderation tooling.
* Configurable blocklists and allowlists.
* AI-assisted abuse classification where appropriate.
* Human review for uncertain cases.
* Telemetry and analytics to continuously improve detection accuracy.

The framework should be modular so that individual detection strategies can be added, removed, or improved independently without requiring major architectural changes.

## Describe alternatives you've considered

Current mitigation largely relies on manual reports and individual link removals. While effective for isolated cases, this does not scale as the platform grows and new abuse patterns emerge.

A strict allowlist or aggressive blocking strategy was also considered, but it would introduce unnecessary friction for legitimate users and conflict with spoo.me's goal of remaining simple and accessible.

Instead, a layered detection system combining multiple independent signals would provide a more scalable and balanced solution.

---

## Additional context

This issue is intended to serve as the primary design and discussion thread for abuse prevention within spoo.me.

The exact implementation is intentionally left open. Contributors are encouraged to propose ideas, research existing approaches used by other URL shorteners, and discuss potential architectures before implementation begins.

The goal is to build a system that significantly reduces abuse while maintaining a smooth experience for legitimate users.

Zingzy:
### Proposed architecture: layered detection

[...]

PR #297, opened 2026-08-17
feat: parse and index destination URL parts

Zingzy:
Part of #233.

Every stored link now carries the parsed parts of its destination (scheme, host, subdomain, registrable domain via the public suffix list) in a nested `dest` subdoc, stamped on all creation paths (v2, legacy `POST /`, `POST /emoji`) and re-stamped when `long_url` is edited. A sparse index on `dest.registrable_domain` covers `urlsV2`, `urls` and `emojis`.

Why: abuse takedowns, sweeps and blocklist application currently regex-scan `long_url` across the full collections. With the decomposition stored and indexed, pivoting on a destination becomes an index lookup, and it is the key the rest of the abuse-prevention work (verdicts, feeds, enforcement) is built on.

Details worth review attention:
- `shared/url_utils.parse_destination` is the single parsing seam: IDN is normalised to punycode, userinfo is discarded so `https://real.com@evil.com` keys on the real host, IP literals and suffix-less hosts key on themselves. The two existing inline tldextract usages were refactored onto it.
- Parsing failures never block a write: unparseable destinations simply omit `dest` (the index is sparse for the same reason).
- `scripts/backfill_url_dest.py` backfills all three collections (batched, `--dry-run`, prints a remaining-count that should reach zero). It must run in production before any feature that queries `dest` is enabled.

Verified with the full test suite and a local end-to-end run of all three creation paths plus the backfill against a live database.

## Summary by CodeRabbit

* **New Features**
  * Added destination metadata to shortened URLs, including scheme, host, subdomain, and registrable domain.
  * Destination metadata is refreshed when a URL is updated.
  * Added destination-based lookup indexes across current and legacy URL records.
  * Added migration support for backfilling destination metadata.

* **Bug Fixes**
  * Improved URL parsing and normalization for ports, IDNs, userinfo, invalid URLs, and public suffixes.
  * Standardized domain detection for referrers and custom domains.

* **Tests**
  * Expanded coverage for destination parsing, URL creation and updates, legacy routes, migrations, and indexing.


shared/url_utils.py
```diff
@@ -17,6 +20,72 @@
...
+    if not host:
+        return None
+    host = host.rstrip(".").lower()
+    if not host:
+        return None
+    if any(ord(ch) > 127 for ch in host):
+        # On IDNA failure keep the lowercased unicode form rather than
+        # dropping the host entirely.
+        with contextlib.suppress(UnicodeError):
+            host = host.encode("idna").decode("ascii")
```
Zingzy, line 74: Nit: `host.encode("idna")` is Python's stdlib IDNA-2003 codec, not UTS-46. It maps `ß`→`ss` and rejects several valid IDN forms (emoji labels, some ZWJ sequences) that a modern browser resolves differently. Verified against the branch venv: `"faß.de".encode("idna")` → `fass.de`, whereas the `idna` package (already a dependency) gives `xn--fa-hia.de`, which is where a browser actually navigates. Consequence is narrow — for that class of IDN the stored `registrable_domain` key won't match the true destination, so a takedown pivot could miss it. Write and backfill agree (both use this codec), so it's internally consistent; it's the divergence from real resolution that's the gap. If IDN destinations matter for abuse pivoting, `idna.encode(host, uts46=True)` is the faithful version.

scripts/backfill_url_dest.py
```diff
@@ -0,0 +1,129 @@
...
+from pymongo import MongoClient, UpdateOne
+
+_FILTER = {"dest": {"$exists": False}}
+_BATCH = 1_000
+_COLLECTIONS = (("urlsV2", "long_url"), ("urls", "url"), ("emojis", "url"))
+_tld = tldextract.TLDExtract(cache_dir=None)
+
+
+def parse_destination(url: object) -> dict | None:
+    """Mirror of shared/url_utils.parse_destination — keep in sync."""
```
Zingzy, line 46: The `# keep in sync` copy of `parse_destination` is the standard cost of a zero-context standalone migration, so I'm not asking you to import the real one. But "keep in sync" is a promise nothing enforces, and this copy will silently rot the day `shared/url_utils.parse_destination` changes (e.g. if you take the UTS-46 nit above). Since this is a one-shot that runs once and is then dead, consider deleting it from the tree after the prod backfill lands rather than leaving a second parser to drift.

Zingzy (commented):
**Mergeable, nothing blocking — this is the strongest PR in the stack and the right foundation for everything above it.** Verified end to end: ran the url-utils, model, service, index and legacy-route suites against the branch (304 passed), and fuzzed `parse_destination` directly against the branch venv with ~18 hostile inputs (userinfo spoof, IDN, IP literal, `[::1` ValueError path, trailing dot, port, 300-char host, `javascript:`). It never raised — the "stamping must never block a save" contract holds.

Architecture read: the single `parse_destination` seam in `shared/` (pure, no I/O — correct layer), the sparse index across all three collections, the absent-not-null discipline folded into the existing claim-field pop, and re-stamping on edit are all exactly right. The two inline `tldextract` usages collapsing onto `registrable_domain`/`is_registrable_apex` is a real dedup, not churn. `dest: UrlDestination | None` is additive and old cache entries/pre-backfill docs decode cleanly (you tested that).

Two nits, both non-blocking. Merge ships nothing (deploys fire on release), and the PR body already names the operational gate: the backfill must run in prod before anything queries `dest`. The one thing I'd add to that runbook — the sparse index build on `urlsV2` at prod scale should be `background`/rolling, not an implicit foreground build during `ensure_indexes` on boot.

scripts/backfill_url_dest.py
```diff
@@ -0,0 +1,129 @@
...
+        if not batch:
+            break
+        ops = [
+            UpdateOne(
+                {"_id": d["_id"]},
+                {"$set": {"dest": parse_destination(d.get(url_field))}},
+            )
+            for d in batch
+        ]
+        coll.bulk_write(ops, ordered=False)
```
Zingzy, line 99: **Should-fix: `bulk_write` has no error handling, and a known class of doc will always fail here.**

`BulkWriteError` is raised even with `ordered=False`. The repo already documents that v1 docs hit the 16MB ceiling from unbounded IP arrays: `repositories/legacy/legacy_url_repository.py:24` exists solely to catch error 10334 for that reason. `$set: {dest: ...}` on such a doc errors, kills the script, and on rerun the doc still matches `{"dest": {"$exists": False}}` so the run dies at the same document forever. `urls` never reaches zero remaining.

Catch `BulkWriteError`, log the offending `_id`s, and continue, so one 16MB doc costs one skipped row rather than the whole migration.

Zingzy (commented):
**Re-review at the same head. I am correcting my earlier verdict: I called this "the strongest PR in the stack, nothing blocking, two nits." The foundation is still right, but I missed a gap in what gets stamped and three defects in the backfill, and one of my supporting claims was wrong.**

Still true, and worth restating: the single `parse_destination` seam in `shared/` is the correct layer, the sparse index across all three collections is right, and re-stamping on edit is right. `dest` is additive and old docs decode cleanly. I re-verified the fuzz result (300k random hosts, zero exceptions from the string parsing).

**Should-decide, and it undercuts the premise of everything above this PR: `geo_rules` destinations are never stamped.** `_handle_geo_rules` (`services/url_service.py:308-330`) ends at `ops["geo_rules"] = ...` with no `dest` touch, and `services/url_service.py:906` stamps `request.long_url` only. But `routes/redirect_routes.py:207-214` redirects to `url_data.geo_rules[resolved_country]`. So a link with `long_url=https://google.com` and `geo_rules={"US": "https://evil.tld/login"}` stamps `dest.registrable_domain = "google.com"`, and every takedown query this PR exists to enable returns zero rows while the link keeps serving to US visitors. Geo cloaking is the standard evasion for this exact threat. Either stamp geo destinations behind a second sparse index, or say in the PR body that geo links are out of scope for v1 enforcement so it is a decision rather than a hole.

**Should-fix, the backfill will not complete on `urls`.** `scripts/backfill_url_dest.py:88-99` has no try/except around `bulk_write`, which raises `BulkWriteError` even with `ordered=False`. This repo already documents that oversized v1 docs exist: `repositories/legacy/legacy_url_repository.py:24` defines `_DOCUMENT_TOO_LARGE_CODE = 10334` for exactly the unbounded IP arrays on hot v1 shorts. Adding ~120 bytes of `dest` to a doc at the 16MB ceiling errors, the script dies, and on rerun the same doc still matches the filter so it dies again at the same place. `urls` never converges and the operator sees an intermittent-looking Mongo error rather than "one doc is too big."

**Should-fix, the backfill is quadratic.** The filter is `{"dest": {"$exists": False}}` (`:39`), which no index on this branch can serve, least of all the sparse `dest_registrable` that excludes those docs by construction. Each of the 20k iterations rescans from the start of the collection, skipping everything already stamped, against the primary with no `maxTimeMS` and no read preference. `count_documents(_FILTER)` runs twice more per collection. Carry a `last_id` and query `{"_id": {"$gt": last_id}, **_FILTER}` sorted by `_id` to make it one pass.

**Should-fix, the backfill does not call `parse_destination`.** `scripts/backfill_url_dest.py:45` is a hand-copy labelled "keep in sync", and the bodies already differ. It also pins `tldextract>=5.1` in its PEP-723 header against `tldextract==5.3.1` in `requirements.txt`, so `uv run --script` can resolve a different bundled suffix list. If a suffix moves between the ICANN and private sections, backfilled docs key on `blogspot.xyz` while app-stamped docs key on `evil.blogspot.xyz`, and an enforcement query silently returns a partial result set. Import the real function.

**Should-fix, `dest` leaks into the public `link.updated` webhook.** `_handle_long_url` puts `dest` into `ops` (`services/url_service.py:180`), and `ops` feeds `event_changes` (`services/webhooks/payloads.py:111`), whose own docstring says "Keys speak the snapshot's public names, never Mongo field names." `dest` is in neither `_CHANGE_KEY_PUBLIC` nor `link_snapshot`, so every destination edit now ships subscribers a `changes.dest` entry where `old` is a model instance and `new` is a dict. It does not crash (extra="allow"), but the Slack and Discord cards render a raw dict row, and it consumes one of the 12 capped change rows so it can push a real field off the card.

**Nit, but the comment is actively misleading.** `shared/url_utils.py:23-25` says `cache_dir=None` gives "the bundled PSL snapshot: deterministic, offline, no first-call network fetch." I tested this against the repo venv: `cache_dir=None` disables only the disk cache, and tldextract 5.3.1 still attempts `https://publicsuffix.org/list/public_suffix_list.dat` with `cache_fetch_timeout=None`, falling back to the snapshot only after that fails. To be fair to this PR, the pattern is pre-existing on `main` (`services/click/handlers.py:34`, `services/custom_domain_service.py:49`) and the fetch is memoized to one attempt per process, so this is an inherited hazard rather than a new one. But this PR is what consolidates it into one seam, which makes it the one place to fix it: pass `suffix_list_urls=()` to genuinely go offline, and correct the comment either way.

**Two smaller ones.** `registrable_domain` can be a leading-dot non-domain: when the host is itself a public suffix, `ext.domain` is empty and `shared/url_utils.py:77` produces `.co.uk`. Fall back to `host` when `ext.domain` is empty. And IDNA: `host.encode("idna")` (`:73`) is IDNA2003, while browsers use UTS-46, so `https://faß.de/` stamps `fass.de` while the browser resolves `xn--fa-hia.de`. That is a real, unrelated registrable domain, so a block on the stamped value sweeps innocent links and a block on the real one misses. The `idna` package is already a transitive dependency; `idna.encode(host, uts46=True)` matches browsers.

**Stack P0, raised here because this is where the index block lives: `dest.host` is never indexed.** This PR indexes `dest.registrable_domain` only (`repositories/indexes.py:85-88`), but every enforcement query above this PR keys on `dest.host`. I confirmed the index is still absent at the top of the stack. Detail is in my #304 review; adding it to the loop here is one line and fixes it for the whole stack.

**Merge safety.** Merging ships nothing. The operational gate stands and grows: the backfill must run before anything queries `dest`, and it cannot currently finish. Also note the sparse index builds are awaited inside the lifespan before routes serve (`app.py:145`), and pre-backfill zero docs have the path, so all three builds scan end to end on first boot after this deploys, per replica. Build them out of band or accept a slow first start.

scripts/backfill_url_dest.py
```diff
@@ -0,0 +1,129 @@
...
+    if dry_run:
+        sample = coll.find_one(_FILTER, {url_field: 1})
+        print(
+            f"[{coll.name}] sample parse: "
+            f"{parse_destination((sample or {}).get(url_field))}"
+        )
+        return
+    done = 0
+    while True:
+        batch = list(coll.find(_FILTER, {url_field: 1}).limit(_BATCH))
```
Zingzy, line 89: **Should-fix: this is O(N^2).** `{"dest": {"$exists": False}}` cannot be served by any index here (the new sparse `dest_registrable` excludes exactly these docs), so every iteration is a fresh scan from the start of the collection past everything already stamped. On a 20M-doc `urls` that is on the order of 10^11 document reads against the primary, with no `maxTimeMS`, repeatedly evicting the WiredTiger cache that the live redirect path depends on.

Carry the cursor: `{"_id": {"$gt": last_id}, **_FILTER}` sorted by `_id`, and it becomes a single pass.

shared/url_utils.py
Zingzy: Fixed in #318: parse_destination now encodes via idna.encode(host, uts46=True) (the idna package was already a dependency), falling back to the lowercased unicode form on IDNAError. The backfill copy got the same change so write and backfill still agree, now on the form browsers actually resolve.

scripts/backfill_url_dest.py
Zingzy: Agreed on both counts. The copy stays for the zero-context reason you gave, and it got the UTS-46 change in #318 so the two parsers still agree. Will delete the script from the tree once the prod backfill has run; adding that to the rollout checklist.

scripts/backfill_url_dest.py
Zingzy: Fixed in #318: bulk_write is wrapped in a BulkWriteError handler that logs the offending ops (first 10 with code and errmsg) and continues, and because the loop now paginates by _id the cursor moves past a poisoned doc instead of re-matching it forever. One 16MB doc costs one skipped row, and the final summary reports failed count and expected remainder.

scripts/backfill_url_dest.py
Zingzy: Fixed in #318: the loop carries an _id cursor (sort by _id, filter _id > last) so the whole migration is a single pass. Same shape you suggested.

PR #298, opened 2026-08-17
feat: scheduled task runner on a Mongo claim lease

Zingzy:
Infrastructure for recurring background jobs: feed syncs and sweeps for the abuse-prevention work (#233), and future crons like the custom-domain reverification pass and clicks archival.

Design, briefly:
- One `scheduled_tasks` collection holds durable state per task (`_id` is the task name): schedule, `next_run_at`, lease, enabled switch, last run outcome. The runner polls and claims due tasks with the same atomic `find_one_and_update` lease the webhook delivery executor already uses, so concurrent runners (worker plus app during a deploy overlap) can never double-run a task.
- Poll equals clock equals claim: there is no separate beat process. A task that was due during downtime runs once on the first loop after boot, and the next occurrence is always computed from now, so missed windows never replay.
- Schedules live in the code registry (validated at boot, `cronsim` for parsing); runtime state (enabled, next run, last run) lives in Mongo where operators can pause a task or arm an immediate run by setting `next_run_at`.
- `SCHEDULER_RUNTIME=auto|worker|embedded|off` mirrors `WEBHOOKS_RUNTIME`: the click worker hosts the loop when the deploy has one, otherwise the app process runs it. Needs only Mongo, so self-hosted deployments get it with zero extra services.
- A built-in hourly `scheduler-heartbeat` task writes a run record on every cycle; an external monitor alerting on its silence covers every failure mode of the machine, including the ones nobody anticipated.

Execution is at-least-once (a crashed run re-claims after the lease expires), so task handlers must be idempotent; that contract is documented on the registry.

Verified with the full suite plus live runs: registry sync, run-now, catch-up after restart, and a two-process race that confirmed the lease yields exactly one execution.

## Summary by CodeRabbit

* **New Features**
  * Added configurable cron-based background task scheduling.
  * Added persistent task tracking, leasing, run results, and on-demand execution.
  * Added a built-in hourly scheduler heartbeat.
  * Scheduler supports automatic or worker-based operation with configurable polling and lease intervals.
  * Added safeguards against concurrent execution and stale task claims.

* **Documentation**
  * Added example environment variables for scheduler configuration.


config.py
```diff
@@ -445,6 +445,41 @@ def delivery_log_ttl_seconds(self) -> int:
...
+      auto     — worker when a queue Redis is configured (a worker exists
+                 in that deploy), else embedded in the app lifespan
+      worker   — only the click worker runs the loop
+      embedded — the app process runs it as a lifespan task
+      off      — nothing polls; tasks stay armed in Mongo
+
+    Enabled by default: with only built-in tasks registered it is a
+    once-an-hour heartbeat, and feature tasks gate themselves on their own
+    settings. The lease makes overlapping runners (deploy overlap, worker
+    plus app) safe — a task can never double-run.
```
Zingzy, line 463: This says "a task can never double-run" but `lease_seconds` at line 480 documents the opposite: the lease only protects a run that finishes *within* it. `runner._execute` awaits `task.fn()` unbounded with no mid-run renewal, and `finish_run` releases on `_id` alone with no owner/lease fence — so a task that runs longer than `lease_seconds` (600s default, 10× the webhook executor's 60s) becomes claimable by a second runner *while still executing*, and a slow runner's late `finish_run` can clobber a newer claim's `next_run_at`/`last_run`. The honest guarantee is at-least-once with idempotent handlers (which the registry docstring already states correctly). The heartbeat and the fishfish sync (~3s) are both far under the lease, so there's no live double-run today — but reconcile this line with reality before a genuinely long task (a sweep, clicks archival) is added, and consider whether 600s is right for one VPS.

dependencies/wiring.py
```diff
@@ -282,6 +285,24 @@ def wire_services(app: FastAPI, settings: AppSettings, redis_client) -> None:
...
+    # Redis presence is the worker proxy. app.py starts/cancels the task.
+    sch_settings = settings.scheduler
+    task_registry = build_task_registry()
+    app.state.task_scheduler = TaskScheduler(
+        ScheduledTaskRepository(db["scheduled_tasks"]),
+        task_registry,
+        poll_interval=sch_settings.poll_seconds,
+        lease_seconds=sch_settings.lease_seconds,
+    )
+    app.state.task_scheduler_embedded = sch_settings.enabled and (
```
Zingzy, line 301: `auto` mirrors the webhook *expression* but not its semantics. For webhooks, queue-Redis-present implies the worker boots and hosts the executor, so `auto`→worker is self-consistent. Here the worker (`create_worker_app`) refuses to boot for the scheduler alone and explicitly "never justifies booting a worker by itself." So with `CLICK_EVENTS_QUEUE_REDIS_URI` set but clicks inline, meta off, webhooks off, this resolves to `worker` — and no worker exists → **nothing runs the scheduler, with no warning**, on the default `enabled=True`. Every sibling misconfig in `create_worker_app` logs a warning; this one is silent. Either resolve `auto` off the same predicate the worker actually uses to decide it's running, or emit a startup warning when the resolved host process won't exist. (Cosmetic tell of the copy: the scheduler reads a var literally named `queue_redis_for_webhooks`.)

services/scheduler/runner.py
```diff
@@ -0,0 +1,151 @@
...
+        self._repo = repo
+        self._registry = registry
+        self._poll_interval = poll_interval
+        self._lease = lease_seconds
+
+    # ── Loop ─────────────────────────────────────────────────────────────
+
+    async def run(self) -> None:
+        """Long-lived task; cancellation is the shutdown path."""
+        await self.sync_registry()
```
Zingzy: `sync_registry()` is awaited outside the loop's `try`. A transient Mongo error here (or the `E11000` two booting processes can hit racing the `ensure_task` upsert on the unique `_id`) escapes `run()`, and nothing installs a done-callback or restarts it, so the scheduler dies silently for the process lifetime. Worse at shutdown: `app.py` does `await asyncio.wait_for(task_scheduler_task, ...)` under `suppress(CancelledError, TimeoutError)` — a stored non-cancel exception is re-raised there, turning a dead scheduler into an unhandled error during lifespan shutdown. Wrap the initial sync (retry, or fold it into the first loop iteration inside the `try`). The webhook executor precedent has no unguarded pre-loop await.

Zingzy (commented):
**Mergeable — the claim query is genuinely correct and the design is sound — but two claims in the description are overstated, and there's one silent-no-op config path I'd fix before relying on this in prod.** Ran the scheduler/repo/settings/index/smoke suites against the branch (80 passed). I traced the lease against the webhook-executor precedent and against the `cronsim` 2.7 source in the branch venv.

What's solid: `claim_due`'s `find_one_and_update` filter is atomic and correctly shaped — `enabled` + `next_run_at $lte now` + the `claimed_until` `$or` are all in one filter, single-doc, sorted, `ReturnDocument.AFTER`. `{"claimed_until": None}` matches both null and missing in MQL, and `ensure_task` writes it null on insert, so the unset case is safe. `$setOnInsert`-only means restarts self-heal and an overdue task runs once on first poll; `finish_run` re-arms from post-run `now`, so missed windows never replay (confirmed against cronsim: `__next__` yields strictly `> now`, no hot-loop on a boundary). Failure isolation in the loop is right, and the heartbeat genuinely exercises the full claim path.

Three things to weigh — details inline. The headline: **"a task can never double-run" is false as written** (no lease renewal + unfenced release), and **`SCHEDULER_RUNTIME` does not actually mirror `WEBHOOKS_RUNTIME`'s semantics** — one config resolves to nobody running the loop, silently, on the default `enabled=True`. Neither blocks merge (nothing runs until a task is registered and the process hosts the loop), but the auto-resolution gap is worth closing before this carries the feed syncs in #300.

services/scheduler/runner.py
Zingzy: **Should-fix: a boot-time failure here kills the scheduler silently, for good.**

This await runs before the try/except loop. Reproduced: with `ensure_task` raising once at boot, `run()` dies with zero polls ever issued, and nothing logs at failure time since the exception sits unretrieved on the asyncio task. Two consequences:

- the process never schedules anything until its next restart, and the heartbeat monitor that would catch the silence does not exist yet
- at app shutdown, `wait_for` on the already-dead task re-raises the original exception through the lifespan, because the `suppress` only covers `CancelledError` and `TimeoutError`

The webhook executor cannot hit this: it has no await before its loop. Cheapest fix that keeps the shape is to sync inside the loop until it succeeds:

```python
async def run(self) -> None:
    synced = False
    while True:
        try:
            if not synced:
                await self.sync_registry()
                synced = True
            row = await self._repo.claim_due(lease_seconds=self._lease)
            ...
```

That also fixes the shutdown re-raise, since the task can no longer die.

services/scheduler/runner.py
```diff
@@ -0,0 +1,151 @@
...
+        for orphan in await self._repo.list_names() - registered:
+            log.warning("task_doc_orphaned", task=orphan)
+
+    # ── One run ──────────────────────────────────────────────────────────
+
+    async def _execute(self, row: ScheduledTaskDoc) -> None:
+        name = row.id or ""
+        task = self._registry.get(name)
+        if task is None:
+            # Claimed a doc this process has no handler for (orphan or
```
Zingzy: **Should-decide: this path consumes the occurrence instead of yielding it.**

During a deploy overlap, a process without the handler can win the claim, record an error run, and re-arm `next_run_at` to the *next* occurrence. The process that does know the task never gets that run, so a task's first-ever occurrence can be swallowed if it lands in the overlap window. The comment says the doc "stays claimable by a process that does know it", but the occurrence itself is spent.

The alternative: release the lease (`claimed_until: None`) without touching `next_run_at`, and let the knowing runner claim it within one poll. The two runners would trade claims for the few seconds of overlap, bounded and harmless.

In prod only the worker hosts the runner, so the window is seconds and the next occurrence runs fine. Leaving it as-is is defensible if a skipped occurrence during a deploy is acceptable; flagging so it is a conscious call.

repositories/scheduled_task_repository.py
```diff
@@ -0,0 +1,119 @@
...
+                "$set": {
+                    "last_run": result.model_dump(),
+                    "next_run_at": next_run_at,
+                    "claimed_until": None,
+                    "updated_at": datetime.now(timezone.utc),
+                }
+            },
+        )
+
+    async def invoke_now(self, name: str) -> bool:
```
Zingzy, line 146: **Nit: `invoke_now` and `find_by_name` have no caller outside tests.**

The ops path today is a direct Mongo write, so nothing reaches these yet. Keeping `invoke_now` as the codified arming contract for a future endpoint is reasonable, but two behaviors to know about if it stays: it returns True for a disabled task that will never actually run, and an arm that lands while the task is mid-run is silently overwritten by `finish_run`. `find_by_name` I would drop until something reads it.

Zingzy (commented):
Mergeable, nothing blocking. One should-fix in the runner's boot path, one design call on rollout overlap, and a couple of nits. CI is green, and I verified the claim semantics against a live Mongo instead of taking the PR body's word for it: a null `next_run_at` never matches the claim query, 20 concurrent `claim_due` calls yield exactly one winner, a held lease blocks re-claim, an expired lease reopens it, and a disabled task never claims.

**Architecture read.** The load-bearing choices are right. Poll = clock = claim removes the beat process entirely: comparing the durable `next_run_at` against now *is* the scheduler, so restarts self-heal and missed windows never replay. Code as source of truth for what/when plus Mongo for runtime state is the correct split, and `ensure_task` writing only `$setOnInsert` means an operator pause survives every deploy. The runtime resolution in wiring.py is copied verbatim from the webhook executor's line, so `auto` semantics cannot drift between the two systems.

**Should-fix** (inline at `services/scheduler/runner.py:47`): `sync_registry()` runs before the protected loop; a Mongo blip at boot kills the scheduler silently for the process lifetime, and the app's shutdown block then re-raises the stale exception. Reproduced, details inline.

**Should-decide** (inline at `services/scheduler/runner.py:106`): the unknown-task path consumes the occurrence during a deploy overlap instead of yielding it to the process that has the handler.

**Nits.**
- `invoke_now` and `find_by_name` have no caller outside tests (inline at the repository).
- A few comment blocks run three to five lines restating the `SchedulerSettings` docstring (the wiring.py scheduler block, the click_worker `run_scheduler` block). One tight line each; the design story already lives in the class docstring and the PR body.

**Merge safety.** Merging ships nothing (deploys fire on release publish). Worth noting this system is not flag-gated and defaults on, so the next release starts the poll loop and the hourly heartbeat in every deploy. That is deliberate and cheap: one indexed no-op query per 5s plus a no-op task per hour. The real reliance gate comes later. Before the first real task rides this: (1) create the external monitor that alerts on heartbeat silence, because the should-fix above is exactly the failure mode it exists to catch, and (2) remember the 600s lease must exceed the slowest task or a second runner will double-run it concurrently; clicks archival is the one to watch.

**Product.** This is the missing piece for three already-planned consumers: the abuse-prevention feed syncs and sweeps (#233), the custom-domain reverification pass (`reverify_active` and its config tunables have been on main with zero callers since the custom-domains work), and clicks archival. Mongo-only was the right call for self-host: no new services, and pause/run-now being plain Mongo writes means ops tooling needs no new API before it is useful.

**Notably good.** The lease survived a real 20-way concurrent claim race on the first try. Registry validation turns duplicate names and bad cron into boot errors, and cronsim rejects impossible dates like Feb 30 at construction, so there is no runtime cron edge left. All four collection sync points (wiring, worker, indexes plus its test, both smoke dicts) were updated.

Zingzy:
Pushed 2460d62 addressing the review findings:

- Boot-time registry sync now retries inside the loop (logged, poll-interval backoff, cancellation still propagates), and the concurrent `ensure_task` upsert race counts as success. A transient Mongo error at boot can no longer kill the scheduler silently.
- `claim_due` takes the registry's task names and adds `_id: {$in: names}` to the claim filter, so a process only claims tasks it can execute. The deploy-overlap case can't swallow an occurrence anymore; the runner's unknown-task branch is now a defensive log that no longer consumes the run.
- Claims stamp a per-claim token and `finish_run` is fenced on it: a run that outlives its lease can't clear the re-claimer's lease or overwrite its run state. Superseded finishers no-op with an info log.

`invoke_now` is left as is: the remaining quirk there (an immediate-run armed mid-execution is consumed by that run's own finish) is a semantics question rather than a fencing bug, and it has no caller outside tests yet. Can pick it up separately if wanted.

Suite went 2221 to 2227, all green.

config.py
Zingzy: Fixed in #318: the docstring now states the honest guarantee, at-least-once with idempotent handlers, dedupe only while a run finishes within the lease, and says a longer task needs the lease raised first since there is no mid-run renewal. A token-fenced finish_run felt like more machinery than the two current tasks (heartbeat, 3s fishfish sync) justify; happy to add it when a genuinely long task lands.

dependencies/wiring.py
Zingzy: Fixed in #318: wire_services now logs a task_scheduler_delegated_to_worker warning at boot whenever the scheduler is enabled, resolves to worker/auto, and this process will not embed it, spelling out that nothing polls unless a worker is actually deployed. Kept the resolution rule itself, since the app process cannot know the deploy topology, but the silence is gone.

services/scheduler/runner.py
Zingzy: Fixed in #318 with the shape from your later comment: the initial sync happens inside the loop's try (synced flag), so a transient Mongo error or the ensure_task E11000 race retries on the next iteration instead of killing the task and re-raising through lifespan shutdown.

Zingzy: Fixed in #318 exactly as sketched: the sync moved inside the loop behind a synced flag, so a boot-time failure retries every poll interval and the task can no longer die with an unretrieved exception.

services/scheduler/runner.py
Zingzy: Took the alternative in #318: the runner now calls a new release_claim (clears claimed_until, touches nothing else) and sleeps one poll interval, so the occurrence is yielded to the process that knows the task instead of being recorded as an error and re-armed past it. The overlap-window claim trading is bounded by the poll interval as you said.

repositories/scheduled_task_repository.py
Zingzy: Dropped find_by_name in #318. Kept invoke_now as the codified arming contract with its two caveats intact; the ops MCP run-now verb will be its first real caller.

Zingzy:
All commits here (through 2460d62, the review fixes) reached main via the safety-stack merge (#318), so this PR is done in substance. Closing; branch left in place.

PR #299, opened 2026-08-17
feat: report-triggered safety analysis, verdicts and enforcement

Zingzy:
Part of #233. Stacked on #297 (destination decomposition); review against that base.

The first vertical slice of the abuse-prevention framework: a user report now flows report, analysis, verdict, enforcement automatically, instead of waiting for manual triage.

How it works:
- `safety_verdicts` stores one judgment per destination host (tiered: toxic, gray, benign, uncertain, with a reason string for the audit trail). Host granularity avoids collectively punishing shared platforms; domain-wide action stays a deliberate operator move.
- `SafetyAnalyzer` runs a provider chain over a reported destination. The initial providers are the operator sources (the regex blocklist); the deep tier (page rendering plus model-based classification) plugs in later as another provider. No provider judging means an uncertain verdict and a review notification, never an inferred clean.
- Toxic verdicts trigger `SafetyEnforcer`: every active v2 link to that host is flipped to BLOCKED (served as 451 by the existing machinery), Redis and edge caches are evicted, `link.blocked` webhook events fire for owned links, and the operator channel gets an action-taken notification. Legacy v1 matches are counted and surfaced, not auto-deleted.
- Transport mirrors the click pipeline: accepted reports enqueue one analysis request per distinct destination host onto an `events:safety` stream consumed by the click worker (reader plus claimer plus dead-letter guard); without queue Redis, analysis runs inline in the app; with `SAFETY_ENABLED=false` (the default) nothing analyzes.
- Repeat reports of an already-judged host short-circuit: fresh verdicts skip re-analysis, and an existing toxic verdict re-runs enforcement idempotently so links created after the original block still die.

Everything is fail-open by contract: a broken provider, notifier or queue never blocks report storage or link serving.

Verified with the full suite and a live loop: report accepted, verdict written, links 451 within about a second, uncertain destinations routed to review while their links keep redirecting.


repositories/url_repository.py
```diff
@@ -292,6 +292,53 @@ async def check_alias_exists(self, alias: str, domain: str) -> bool:
...
+            {
+                "dest.host": host,
+                "status": UrlStatus.ACTIVE.value,
+                "owner_id": {"$ne": ANONYMOUS_OWNER_ID},
+            }
+        ).limit(limit)
+        docs = await cursor.to_list(length=limit)
+        return [UrlV2Doc.from_mongo(d) for d in docs]
+
+    async def block_active_by_dest_host(self, host: str) -> int:
```
Zingzy, line 327: Enable-blocker: this is a one-way door. It records no reason/timestamp/prior-status on the flipped docs, there is no `unblock_host` counterpart anywhere in the stack, and `url_service` refuses owner edit/delete on BLOCKED links — so a false-positive verdict mass-blocks a host AND locks every owner out, reversible only by hand-editing Mongo. Given the FP vectors in the summary, ship an `unblock_active_by_dest_host` (scoped to links this system blocked, via a `blocked_by="safety"` marker so you don't un-block manual bans) plus `blocked_reason`/`blocked_at` on the doc, before enabling. Also: the flip is unbounded while the invalidation/owned collections cap at 50k/1k silently — for a host over those caps, links flip but excess cache entries keep serving stale ACTIVE and >1k owned links get no `link.blocked`, with the result reporting the truncated count as if complete. Log loudly when a cap truncates.

services/safety/providers.py
```diff
@@ -0,0 +1,99 @@
...
+    name = "blocked_domain"
+
+    def __init__(self, repo: BlockedDomainRepository) -> None:
+        self._repo = repo
+
+    async def analyze(
+        self, url: str, host: str, registrable_domain: str
+    ) -> ProviderVerdict | None:
+        try:
+            if await self._repo.is_blocked(host):
```
Zingzy, line 56: `blocked_domains` is the custom-domain vanity/brand-reservation denylist (its repo docstring: Tranco top-N + brand lookalikes we refuse as *tenant* domains) — by design full of benign popular hosts. Treating a hit as a TOXIC *destination* verdict means one anonymous report of a link to `google.com`/`paypal.com` mass-blocks every link to it. I confirmed #300 deletes this provider, so the merged stack is safe — flagging because it's live and dangerous in this isolated commit: do not enable safety at this SHA, only at #300+.

services/safety/sinks.py
```diff
@@ -0,0 +1,75 @@
...
+    """Runs the analyzer in-process. The full analysis (Mongo lookups +
+    Discord embed) rides the request that triggered it, which is the
+    accepted cost of the no-worker rung."""
+
+    def __init__(self, analyzer: SafetyAnalyzer) -> None:
+        self._analyzer = analyzer
+
+    async def emit(self, event: SafetyAnalyzeEvent) -> None:
+        try:
+            await self._analyzer.analyze(event)
```
Zingzy, line 44: Enable-blocker on the no-queue-Redis rung: this awaits the *full* analysis inside the report POST — `find_by_host` + an uncached fetch of the entire `blocked-urls` collection + a synchronous regex scan over every pattern *on the event loop* + verdict upsert + a blocking Discord webhook POST — per distinct host, up to the batch size. A multi-host report will time out at the gateway (after storage already succeeded), and the surrounding code already pushes these same regex scans to `asyncio.to_thread`. Make the inline sink fire-and-forget (`create_task` + shield) or move analysis fully off the request. Also wrap `_enqueue_safety_analysis` itself — it runs post-storage relying entirely on the sink's internal catch.

services/safety/analyzer.py
```diff
@@ -0,0 +1,122 @@
...
+        self._notifier = notifier
+        self._reverdict_ttl = timedelta(hours=reverdict_ttl_hours)
+
+    async def analyze(self, event: SafetyAnalyzeEvent) -> None:
+        existing = await self._verdict_repo.find_by_host(event.host)
+        if existing is not None:
+            if existing.tier == VerdictTier.TOXIC:
+                # Already judged bad: enforcement is idempotent and cheap,
+                # and covers links created after the original block.
+                await self._enforcer.block_host(
```
Zingzy, line 55: The re-enforce-existing-toxic branch returns without notifying. Combined with enforce-order (upsert → enforce → notify), any exception in `block_host` after the flip leaves links BLOCKED + verdict TOXIC but `safety_action` never fired, and the retry lands here and silently re-enforces — so a mass block can happen with the operator never told. Notify on the re-enforcement path too (even a terse "re-enforced N new links for already-toxic host"), and consider making the enqueue path read PR #242's stored report velocity so one report ≠ fifty.

services/safety/events.py
```diff
@@ -0,0 +1,76 @@
...
+_EVENT_TYPE = "safety.analyze"
+
+
+class SafetyAnalyzeEvent(BaseModel):
+    """Request to judge one destination. Immutable fact-shaped payload."""
+
+    model_config = ConfigDict(frozen=True)
+
+    url: str
+    host: str
```
Zingzy, line 39: `host` has no `min_length`. Report intake can't produce an empty host today (`parse_destination` returns None instead), so this isn't reachable now — but this event is the worker's trust boundary, and a `host=""` event would upsert a `""` row into the unique-host verdict index and run `{"dest.host": ""}` enforcement across `urlsV2`. One-line guard (`min_length=1`, or reject empty host in `analyze`) closes it.

Zingzy (commented):
**Merge-safe (default-deny: `SAFETY_ENABLED=false`, and nothing runs until the flag flips *and* the #297 backfill lands), but NOT enable-safe yet.** This is the sharp end of the whole stack — it flips real users' links to 451 automatically — so I ran it hard: full safety/repo/worker/report/webhook suites against the branch (296 passed), three parallel adversarial passes against the branch source, and direct verification of every claim below. The architecture is right (per-host verdict, compute-once/read-at-edges, fail-open sinks, drop-don't-poison decode, enforce-order collects-before-flip). What follows is the hardening tail that must close **before `SAFETY_ENABLED=true` in prod**, because for spoo the false-positive blast radius is a reputation event, not a bug.

**The one that matters most — there is no reversal path.** `block_active_by_dest_host` flips links to BLOCKED with no audit trail (no `blocked_reason`/`blocked_at`/`prior_status`), there is no `unblock_host` anywhere in the stack (I grepped), and a BLOCKED link can't be edited or deleted by its owner (`Cannot modify/delete a blocked URL`). So a single wrong verdict mass-blocks a host, locks every owner out of their own links, and the only fix is hand-editing Mongo. A one-command reverse with an audit field is the price of automatic enforcement.

**Mass-block false-positive vectors** (verified against the branch): (1) `BlockedDomainProvider` treats any `blocked_domains` hit as TOXIC — but that collection is the custom-domain *vanity/brand reservation* denylist (Tranco top-N, brand lookalikes), i.e. deliberately full of *benign popular* hosts. One anonymous report of a link to such a host → every link to it 451s. **This provider is removed in #300**, which neutralizes it for the merged stack — but it's live in this commit, so if #299 could ever enable independently of #300, it's an outage button on an unauthenticated 5/min endpoint. (2) `validate_blocked_url` is `regex.search` unanchored over the *full* URL; a pattern matching a substring in the path/query of a *shared* host (e.g. `paypal` in a `docs.google.com/...` path) produces a host-keyed TOXIC verdict → mass-block of that shared host. Those patterns were authored for create-time *reject-one-write*; this PR silently retargets them to *retroactively kill every link to the host*. Match on host, or gate host-wide enforcement behind an enforcement-grade pattern set.

**Other enable-blockers** (inline + here): the inline sink runs the *full* analysis (uncached blocklist fetch + regex scan + blocking Discord POST) synchronously inside the report POST — up to N hosts per submission on the event loop, so a report can time out the gateway after storage; make it fire-and-forget. The verdict upsert can `DuplicateKeyError` on concurrent same-host analysis and, on the UNCERTAIN branch, that throws *before* the review notify → operator never told. A human/BENIGN verdict is not sticky (re-analyzed after the TTL) — **#301 fixes this** with the `decided_by != system` check, so it resolves in the stack. `link.blocked` is registered in `EVENT_REGISTRY` but missing from the copy/discord renderer tables, so the single most important webhook renders as a raw-JSON gray card. And `events.SafetyAnalyzeEvent.host` has no guard, so a `host=""` event off the stream would poison the unique verdict index.

Payload leak check: `LinkBlockedPayload` reuses `link_snapshot` — no password hash, no IP; clean. Report-weaponization: PR #242's dedupe+velocity is intact, but the analyzer never *reads* the velocity, so one anonymous report has the same enforcement power as fifty.

services/safety/enforcer.py
```diff
@@ -0,0 +1,120 @@
...
+
+        # 2. Flip.
+        blocked = await self._url_repo.block_active_by_dest_host(host)
+
+        # 3. Evict Redis per domain namespace.
+        by_domain: dict[str, list[str]] = defaultdict(list)
+        for alias, domain in pairs:
+            by_domain[domain or self._system_domain].append(alias)
+        for domain, aliases in by_domain.items():
+            await self._url_cache.invalidate_many(aliases, domain)
```
Zingzy, line 77: **Should-fix before enabling: a retry after this point cannot repair the cache.**

The collect-before-flip ordering above is correct. The problem is recovery. The invalidation set comes from `status == ACTIVE`, so once `block_active_by_dest_host` has run, a re-delivery recomputes `pairs = []` and returns without evicting anything. And re-delivery is the normal path: the consumer re-raises, the message stays pending, the claimer re-runs it into a no-op.

Any failure at the eviction or edge-purge step therefore leaves origin Redis serving the blocked destination for up to `redis_ttl_seconds` (3600), with the operator already told the links are blocked.

Deriving the invalidation set from `dest.host` regardless of status, or writing a short negative marker instead of a bare DELETE, makes the retry idempotent.

Zingzy (commented):
**Re-review at the same head. The enable-blockers I raised still stand, one of my supporting claims was wrong, and there are six new findings. Verdict is unchanged: merge-safe behind `SAFETY_ENABLED=false`, not enable-safe.**

**Correcting myself first.** I said the verdict upsert can `DuplicateKeyError` on concurrent same-host analysis. That is wrong. `repositories/verdict_repository.py:32` queries `{"host": host}` and `repositories/indexes.py:71` makes `host` the unique index, so the upsert predicate is the unique key. That is the supported pattern, not the race. The ordering half of that finding does stand and is broader than I wrote: `upsert_verdict` calls `self._col.update_one` directly, bypassing `BaseRepository`'s error handling, and the analyzer wraps none of it, so any Mongo write failure kills the frame before the notify. On the toxic branch that means enforcement can be partially applied and completely unannounced.

Still confirmed, unchanged: no reversal path (and blocked links cannot be edited or deleted by their owner); `BlockedDomainProvider` judging destinations against the custom-domain brand denylist; unanchored create-time regexes retargeted into host-wide retroactive enforcement; the inline sink running full analysis inside the report POST; human verdicts not sticky; `link.blocked` missing from the renderer tables.

Two amplifiers on the inline sink I did not name last time. The regex sweep runs on the event loop, while the meta-tags path deliberately offloads identical work with `asyncio.to_thread` and the comment "keep the regex scans off the hot loop" (`services/url_service.py:759`). And the Discord POST is inside the critical path at a 5s timeout, so 25 distinct hosts in one submission is a 125-second report POST during a Discord outage.

On `link.blocked`: `expand()` (`services/webhooks/registry.py:275`) matches `link.*` against the live registry, so every endpoint already subscribed to `link.*` starts receiving it on deploy, with no opt-in, rendered as a raw JSON card.

**New, and this is the one that should gate the flag flip: no `dest.host` index.** Every query this PR adds filters on `dest.host` (`repositories/url_repository.py:306`, `:315`, `:332`, plus the two legacy counters), but #297 indexes `dest.registrable_domain` only. The comment at `url_repository.py:295` reads "All three ride the sparse dest_registrable/dest.host paths," which suggests the belief that it is covered. It is not, and I confirmed it is still missing at the top of the stack. One toxic verdict is a full scan of `urlsV2`, `urls` and `emojis`, including a scanning `update_many` that holds write intents, on the box serving the redirect path. From an anonymous 5/min endpoint.

**New: a crash between the flip and the eviction strands the cache permanently, and the retry cannot repair it.** The enforcer computes its invalidation set from `status == ACTIVE` before flipping (`services/safety/enforcer.py:66-86`), which is correct ordering. But once the flip has run, a retry recomputes `pairs = []` and short-circuits, so the eviction never happens. The retry is the normal path: the consumer re-raises, the message stays pending, and the claimer re-runs it into a no-op. Origin Redis then serves the blocked destination for up to `redis_ttl_seconds = 3600`, and the edge for its own TTL, while the operator embed already reported the links as blocked.

There is a non-crash version of the same defect. `resolve` reads Mongo on a cache miss and writes the cache afterwards, so a redirect that read the doc as ACTIVE before the flip but writes the cache after the eviction re-populates an ACTIVE entry with a fresh hour TTL. `invalidate_many` is a bare DELETE with no tombstone or version, so this needs no crash at all, just an interleaving.

**New: silent truncation against an unbounded flip.** `list_active_alias_domain_by_dest_host` caps at 50,000 and `list_active_owned_by_dest_host` at 1,000 (`repositories/url_repository.py:300`, `:312`), while `block_active_by_dest_host` is unbounded. Neither call site compares the result length to the limit, and `EnforcementResult.cache_invalidated` reports the truncated number as fact. On a shared host with 200k links, all 200k flip, 50k get evicted, 150k keep redirecting for an hour, and 1,000 of the affected owners get a webhook. This repo already has the right convention two hundred lines up: `url_repository.py:134` logs a warning when a cursor comes back full, with the rationale written out.

**New: the edge purge result is discarded.** `bulk_delete` returns a bool that is false for an unconfigured client, an HTTP error after retries, a non-2xx, and a 200 with `unsuccessful_keys` (its own comment says "the return value must tell the truth"). Every other consumer honours it: `services/edge_cache/promotion.py:129` and `services/url_service.py:1245` both branch on it. The safety enforcer alone drops it (`enforcer.py:83`), then reports success. With a rotated CF token the Discord card says auto-blocked while the edge keeps serving for the full TTL.

**New: one anonymous submission can emit 25 review embeds, on the same webhook as the report summaries.** Dedupe is per host, so N distinct hosts is N embeds, with no batching or throttle. At 25 items and 5/min that is ~130 embeds/minute from one IP against Discord's ~30/min per-webhook limit, and `_deliver` does not honour `Retry-After`, it just logs and drops. The dropped messages include the legitimate report summaries interleaved with them. The abuse channel becomes the way to bury abuse reports.

**New: the attacker can suppress the action-taken embed.** `safety_action` and `safety_review` interpolate `sample_url` and the `context` dict into embed fields with no length clamp (`infrastructure/ops_notify.py:236`, `:266-271`), while `report_summary` deliberately clamps at 80 chars. Discord rejects a field value over 1024 with a 400. Register a link with a destination longer than 1024 chars, report it, and enforcement executes for real while the notification is dropped, by the same actor who triggered it. The `_deliver` return is discarded at both call sites.

**New, smaller: the re-enforcement path notifies nobody.** `analyzer.py:52` returns after `block_host` with no notifier call, so when a previously-toxic host keeps attracting new links, each subsequent report silently blocks them. The recurrence, which is the signal that a campaign is still live, is invisible outside a log line.

**New, smaller: `www.` is a free evasion.** Verdicts key on the raw host and `registrable_domain` is stored but never read. `parse_destination` preserves the `www.` label, so `www.evil.com` and `evil.com` are two verdicts and two enforcement runs. The rest of the codebase folds `www.` explicitly, including the edge contract. Host granularity is the right call for `*.workers.dev`; it is not right for `www.`.

**Merge safety.** Unchanged: default-deny, and nothing runs until the flag flips and the #297 backfill lands. The list above is the hardening tail before `SAFETY_ENABLED=true`, and the ordering I would use is the index first, then the reversal path, then the two mass-block vectors.

repositories/url_repository.py
Zingzy: Closed across #304, #316 and #318. The doc stamps landed in #304 (blocked_at/blocked_reason on all three collections). #318 adds the reversal: unblock_by_dest_host on the v2 repo (scoped to docs carrying blocked_reason, so manual bans stay untouched) plus SafetyEnforcer.unblock_host reversing all three collections and evicting cache. The 50k listing caps now log dest_host_listing_truncated instead of truncating silently. The ops MCP verb that calls unblock_host is the remaining piece and lives in the other repo.

services/safety/providers.py
Zingzy: Confirmed, and this stayed dead: #300 removed BlockedDomainProvider from the safety chains and nothing reintroduced it. Merge order protects the isolated SHA as you noted.

services/safety/sinks.py
Zingzy: Fixed in #318: InlineSafetySink grew a background mode used by the app wiring, emit detaches the analysis onto its own task (strong refs held, errors logged), so the report POST returns as soon as storage is done. The worker's own sweeps keep the synchronous form so a feed delta stays one analysis at a time.

services/safety/analyzer.py
Zingzy: Fixed in #318: re-enforcement now notifies whenever it actually blocked something (reason suffixed 're-enforced existing verdict'), and stays quiet on the routine zero-block case so known-bad hosts do not spam the channel. The retry-after-partial-failure case therefore pings. Report velocity from #242 as enqueue context is a good follow-up; noted.

services/safety/events.py
Zingzy: Fixed in #318: host is Field(min_length=1) on SafetyAnalyzeEvent with a comment saying why the trust boundary needs it.

services/safety/enforcer.py
Zingzy: Fixed in #318: the eviction sets are now status-blind (list_by_dest_host_with_urls and list_by_dest_host ignore status/blocked), so a re-delivered block recomputes the full key set and repairs the cache instead of no-opping. Flip queries still filter, so counts stay honest.

PR #300, opened 2026-08-17
feat: threat feeds and the create-time policy gate

Zingzy:
Part of #233. Top of the stack: #297 then #299, and this branch also merges #298 (the scheduler hosts the feed syncs). Review the commits after the merge commit; the earlier ones belong to the parent PRs.

Two halves:

**Threat feeds.** Feed sets live in Mongo (`safety_feed_domains`, one doc per feed and domain, primary-key membership reads), refreshed by full swap with an empty-download guard so a bad fetch never wipes the last known-good set. `FEED_REGISTRY` declares each feed as data: which chains it joins, its config switch, its sync task, its optional first-boot seed, its user-facing message. The gate, analyzer, wiring, worker and seeder all derive from the registry, so adding a feed is one entry. Shipping now:
- fishfish.gg, synced hourly as the scheduler's first feature task (about 43k domains, roughly 3 seconds per sync)
- Google Web Risk through the Lookup API (analysis tier only, never the request path), enabled by setting `SAFETY_WEB_RISK_API_KEY`
- `manual`: operator-curated destination domains, enforced at create and on report
- `shorteners`: redirect-chain refusal (create-time only), seeded once from `data/shortener_domains.txt`, DB-owned afterwards

**The create-time gate.** `UrlPolicyService` replaces the inline destination checks and is injected into every creation path, including the legacy routes that previously reimplemented validation (and one of which dropped the configured regex timeout). It reuses the analyzer's provider instances, so a signal is implemented once and fires at two moments: refuse the write at create, judge and enforce at analysis. A destination any tier has judged toxic also refuses new links via the verdict store, closing the loop without copying between lists. Rejection reasons are logged precisely but kept coarse on the wire, except the shortener refusal, which is published policy and says what it is. Pattern reads are now cached briefly in the shared provider instead of re-reading the collection on every request.

One correction included: the custom-domain vanity denylist is no longer consulted for destination judgments, since it intentionally lists popular brand domains and would have produced false positives.

Verified with the full suite and live runs of every path: fishfish-listed and shortener destinations refused at create on v2 and legacy, manual entries enforced within a request, verdict-only refusal confirmed with the originating pattern deleted.


services/url_service.py
```diff
@@ -808,33 +809,22 @@ async def create(
...
         # 2b. Geo rules — every destination gets the same two-stage validation
-        # as long_url (patterns already fetched above). Off the event loop:
-        # up to max_countries * len(patterns) synchronous regex scans, and
-        # the blocklist only grows.
+        # as long_url. Off the event loop: up to max_countries *
+        # len(patterns) synchronous regex scans, and the blocklist only
+        # grows. (Sync helper, so it takes raw patterns rather than the
+        # async gate — unify when the gate grows a sync façade.)
         if request.geo_rules:
+            blocked_patterns = await self._blocked_url_repo.get_patterns()
```
Zingzy, line 827: Enable-blocker: `geo_rules` destination URLs still go through the inline `validate_url` + pattern-only path (`_validate_geo_rules`), bypassing the feeds, the shortener refusal, and the toxic-verdict store that every other create path now enforces via `url_policy.check`. A link with a clean `long_url` and a fishfish-listed `geo_rules` target ships that destination to the targeted region, and #299 enforcement misses it too (`dest` is stamped from `long_url` only). Your deferral comment is honest, but this needs the sync façade (or per-target `await url_policy.check`) before enabling, or the gate's "sole authority" invariant is false in exactly the region-targeted case an abuser would use.

Zingzy (commented):
**Merge-safe (still behind `SAFETY_ENABLED=false`), and it fixes the worst #299 vector — but it opens one new create-path bypass and one secret-in-logs issue that are enable-blockers.** Ran the safety/feed-repo/index/legacy/url-service/bulk/smoke suites against the branch (390 passed). FEED_REGISTRY-as-catalog is a genuinely good abstraction (mirrors the webhook EVENT_REGISTRY discipline), the full-swap-with-sanity-floor refresh is correct (insert-new-then-purge-old, no empty window; the `<100` floor blocks a bad download), and I read `data/shortener_domains.txt` line by line — it's clean, no false-positive landmines (no `linktr.ee`/docs/drive, corporate hosts correctly excluded, spoo's own domains correctly absent since self-links are `validate_url`'s job). Removing `BlockedDomainProvider` from the destination chain (claim 9) is the right call and neutralizes the #299 brand-list oracle. The regex-timeout fix on the emoji route is real.

**The bypass (enable-blocker):** the gate is now the sole destination authority for `long_url` on every create path — *except* `geo_rules` destinations, which still run the old inline `validate_url` + pattern-only pair and never see the feeds, the shortener refusal, or the verdict store. So `POST /api/v2/urls` with a benign `long_url` and `geo_rules={"US": "https://<fishfish-listed>/"}` creates a link that serves a blocked destination to US traffic. Your own comment at the site flags it as deferred — but it makes the gate's stated invariant ("sole authority on whether a destination may be written") false, and it also escapes #299 enforcement since `dest` is stamped from `long_url` only.

**The leak (enable-blocker, only when Web Risk is configured):** `WebRiskProvider` passes the GCP key as a `?key=` query param, and `httpx` logs `HTTP Request: GET <full-url>` at INFO — but `httpx` is *not* in the silenced-logger list in `infrastructure/logging.py` (urllib3/werkzeug/botocore/boto3/pymongo are), and prod `LOG_LEVEL=INFO`. So every Web Risk lookup prints the API key (and the user destination under analysis) to stdout/Axiom. The docstring literally claims "the API key never appears in logs." Move it to the `X-Goog-Api-Key` header and/or silence the `httpx` logger.

**Two more to decide (inline):** there's no kill switch for the shortener refusal — it's `enabled=lambda s: True`, live on anonymous `POST /` even when the master safety switch is off, reversible only by deleting Mongo rows; for a change to a public bot-facing API that's the wrong shape. And the legacy routes discard `rejection.public_message` and always return "Blocked URL ⛔", so the *published* shortener policy reads to integrators as an opaque malice flag — the exact opposite of the coarse-vs-published split the gate designs for. Wiring completeness is otherwise fully correct (I checked `get_url_policy` + alias + `__all__` both blocks, the new collection in both smoke fakes, the index + its assertion, and the permissive test gate in conftest).

services/safety/providers.py
```diff
@@ -39,33 +42,126 @@ async def analyze(
...
+    async def analyze(
+        self, url: str, host: str, registrable_domain: str
+    ) -> ProviderVerdict | None:
+        try:
+            response = await self._http.get(
+                f"{self._api_base}/v1/uris:search",
+                params={
+                    "uri": url,
+                    "threatTypes": list(self._THREAT_TYPES),
+                    "key": self._api_key,
```
Zingzy, line 114: The GCP key rides as a `?key=` query param, and `httpx` logs the full request URL at INFO. `httpx` is not silenced in `infrastructure/logging.py` and prod is `LOG_LEVEL=INFO`, so the docstring's "the API key never appears in logs" is false — the key and the destination URI under analysis both land in stdout/Axiom on every lookup. Move the key to the `X-Goog-Api-Key` header (and silence/raise the `httpx` logger to WARNING) before setting `SAFETY_WEB_RISK_API_KEY` anywhere real.

services/safety/feeds.py
```diff
@@ -0,0 +1,223 @@
...
+        enabled=lambda s: True,
+    ),
+    FeedSpec(
+        name=SHORTENER_FEED,
+        reason_label="a link shortener (redirect chains are refused)",
+        gate=True,
+        # Gate-only: existing links to shorteners are the deep tier's
+        # chain-resolution problem, never a mass-block.
+        analyzer=False,
+        enabled=lambda s: True,
```
Zingzy, line 119: Should-decide: the shortener refusal is `enabled=lambda s: True`, so it's live on anonymous `POST /` even with the master `SAFETY_ENABLED=false`, and the only rollback for a bad entry is deleting Mongo rows by hand. This changes the shipped behavior of a public, bot-facing API with no env kill switch and no dark-launch. Given a shortener destination is a legitimate (if discouraged) use for some integrations, consider a `SAFETY_SHORTENERS_ENABLED` gate so this can be rolled out and rolled back like everything else, rather than being on the moment the code merges.

routes/legacy/url_shortener.py
```diff
@@ -111,22 +109,21 @@ def _get(key: str) -> str | None:
...
+            return JSONResponse(
+                {
+                    "UrlError": (
+                        "Invalid URL, URL must have a valid protocol and must follow"
+                        " rfc_1034 & rfc_2728 patterns"
+                    )
+                },
+                status_code=400,
+            )
         return JSONResponse({"BlockedUrlError": "Blocked URL ⛔"}, status_code=403)
```
Zingzy, line 127: The legacy routes discard `rejection.public_message` and hardcode "Blocked URL ⛔" for every non-`invalid_url` code. So a shortener refusal — a *published* policy the gate deliberately gives a descriptive message — reaches integrators as an opaque malice flag on exactly the highest-volume path, which is the opposite of the coarse-for-security / clear-for-policy split `UrlPolicyService` is built around, and makes `public_messages` dead weight here. Surface `rejection.public_message` on these routes (keeping the frozen key names `UrlError`/`BlockedUrlError`). No test covers a shortener rejection through a legacy route.

Zingzy (commented):
**Re-review at the same head. All four of my earlier findings stand, with one scope correction, and there are six new ones. Merge-safe, still not enable-safe, and one item is live even with the master switch off.**

Confirmed unchanged: the `geo_rules` bypass of the gate; the Web Risk key riding in a query string with `httpx` still absent from the silenced-logger list; no kill switch for the shortener refusal; the legacy routes discarding `rejection.public_message`.

**Scope correction on the Web Risk leak.** I implied it fires per create. It does not: `WebRiskProvider` is appended to `analyzer_providers` only (`dependencies/wiring.py:337-347`), never to the gate, so it leaks once per report-triggered analysis, not once per create. Still a leak, and the docstring at `services/safety/providers.py:87` still claims "The API key never appears in logs," which is false as wired. Worth noting the same unsilenced `httpx` INFO logger also prints the destination under analysis and the fishfish sync URL.

Still right, and I want to keep saying so: `FEED_REGISTRY` as catalog-as-code is a genuinely good abstraction, the insert-new-then-purge-old refresh with a sanity floor is the correct shape, and removing `BlockedDomainProvider` from the destination chain is the right call.

**New: nothing bounds what a feed can contain.** The only content validation is a floor of 100 domains and an `isinstance(str)` check. There is no allowlist of never-blockable domains, no absolute ceiling, no maximum-delta guard. Transport is equally open: the shared client has `follow_redirects=True`, and `services/safety/feeds.py:194-197` does `raise_for_status()` then `response.json()` with no content-type check and no body size cap. If fishfish is compromised or its DNS is hijacked and it serves 43k entries including `google.com` and `github.com`, that passes the floor, full-swaps in within the hour, refuses every create to those domains, and (because fishfish is `analyzer=True`) produces toxic verdicts that the enforcer acts on. Recovery is an out-of-band Mongo purge and the next hourly sync re-poisons. The same unbounded read is an OOM path for the click worker on a multi-GB body. A ceiling, a delta cap, and a small hardcoded never-block list would each be a few lines.

**New: the gate adds 5 to 7 sequential Mongo round-trips to every create, including when safety is off.** `services/safety/policy.py:77` iterates providers with no `gather`. With the default `SAFETY_ENABLED=false` chain the live providers are still `blocked_pattern`, `toxic_verdict`, `feed_manual` and `feed_shorteners`, and each `FeedDomainProvider` does up to two point reads (host, then registrable domain). The reads are indexed and individually cheap; the issue is the un-parallelised count on the anonymous `POST /` path, and that it runs with no way to turn it off. Related correctness gap: `ToxicVerdictProvider` checks only `host` (`providers.py:157`), unlike `FeedDomainProvider`, so a toxic verdict on `evil.com` does not refuse a create for `www.evil.com`.

**New: the pattern cache stampedes, and it abstains while holding a good cache.** `providers.py:198-207` updates `_fetched_at` only after the await resolves, with no lock and no in-flight future, so every concurrent create crossing the 30s boundary fires its own full `blocked-urls` read. The TTL is also effectively hardcoded: `dependencies/wiring.py:326` never passes `patterns_ttl_seconds` despite the docstring advertising it. Second half: when the refresh raises, the exception escapes to the outer handler which abstains, even though `self._cached` still holds a valid list. A transient Mongo blip disables regex blocking for the incident instead of falling back to the cache in hand.

**New: deleting the shortener rows as a kill switch is undone on the next boot.** `ensure_feed_seeds` runs unguarded in the lifespan on every app process start and re-seeds any feed whose count is zero. Since the shortener refusal has no config switch, the only full rollback is deleting all 45 rows, which sets the count to zero, which the next deploy or replica scale-up re-seeds. The docstring's claim that "an operator removing an entry stays removed" holds only while at least one entry remains. Separately, `load_shortener_seed` reads the file with no try/except and nothing above it catches, so a missing `data/shortener_domains.txt` is a hard boot failure for the whole app rather than a degraded feed. Packaging itself is fine, `data/` does ship in the image.

**New: toggling safety off later leaves a task doc that errors hourly forever.** `build_feed_tasks` filters on `spec.enabled(settings)`, so with `SAFETY_ENABLED=false` the fishfish task is not registered. But `sync_registry` never deletes docs, `ensure_task` uses `$setOnInsert` so `enabled` and `next_run_at` persist, and `claim_due` matches on those fields with no notion of registration. The orphan gets claimed hourly, finds no handler, logs `task_unknown`, and re-arms itself from the stored schedule. Inert on day one since the task was never registered, so this is a rollback hazard rather than a launch defect.

**New: the gate fails open on partial Mongo degradation.** Every provider swallows and abstains, which is deliberate and documented. A full outage is self-limiting because the insert fails too. The exposed case is `safety_feed_domains` and `safety_verdicts` reads timing out under load while `urlsV2` inserts still succeed: the gate silently degrades to format and self-link only, and every listed shortener, manual entry, fishfish domain and previously-verdicted destination becomes creatable at full anonymous volume. Nothing distinguishes "abstained because clean" from "abstained because the backend broke."

**One thing to check against prod before merge.** `BlockedDomainProvider` is gone and its replacement, the `manual` feed, starts empty with no sync task and no migration script. If any prod `blocked_domains` rows are genuinely destination-side rather than custom-domain vanity entries, they stop being enforced on deploy with no warning. Your comment says the collection was always the vanity denylist; worth a one-line query to confirm before this lands.

**Wiring completeness is otherwise correct**, including both smoke fakes, the index and its assertion, and the permissive test gate in conftest. I also re-checked the bulk endpoints and the claim flow: no bulk create exists and claim does not write a destination, so `geo_rules` remains the only ungated destination write.

services/safety/providers.py
Zingzy, line 114: **Enable-blocker (only once Web Risk is configured): this logs the GCP key.**

`httpx` logs `HTTP Request: GET <full url>` at INFO, and it is not in the silenced-logger list in `infrastructure/logging.py` (urllib3, werkzeug, botocore, boto3 and pymongo are). Prod runs `LOG_LEVEL=INFO`. The structlog redactor is field-name based, so a secret inside a log *message* is never redacted. Every lookup therefore prints the key and the destination under analysis to stdout and on to the log sink.

The docstring above claims the opposite. Move it to the `X-Goog-Api-Key` header and add `httpx` to the silenced list, since that logger also prints the fishfish sync URL and every other outbound call.

services/safety/feeds.py
```diff
@@ -0,0 +1,223 @@
...
+    )
+
+
+FEED_REGISTRY: tuple[FeedSpec, ...] = (
+    FeedSpec(
+        name=MANUAL_FEED,
+        reason_label="the operator blocklist",
+        gate=True,
+        analyzer=True,
+        enabled=lambda s: True,
```
Zingzy, line 110: **Should-decide: `enabled=lambda s: True` means these two feeds have no off switch.**

The manual and shortener feeds are built into `gate_providers` unconditionally in `wire_services`, outside the `SAFETY_ENABLED` branch (only the sink is gated). So with the master switch off, the shortener refusal is still live on anonymous `POST /`, which is a public bot-facing API.

The only rollback is deleting rows from `safety_feed_domains`, and that is self-undoing: `ensure_feed_seeds` re-seeds any feed whose count is zero on every app boot, so a full delete comes back on the next deploy.

For a change to published behavior on a public endpoint, this wants a real setting.

services/url_service.py
Zingzy: Fixed in #318: geo rule destinations now run the full url_policy.check per target on both the create path and _handle_geo_rules, same gate as long_url (feeds, shortener refusal, toxic verdicts). The old sync helper shrank to shape checks (entry cap, ISO codes). Tests updated, including a blocked geo destination rejected through the real gate.

services/safety/providers.py
Zingzy: Fixed in #318: key moved to the X-Goog-Api-Key header and httpx joined the silenced-logger list at WARNING, which also stops the fishfish sync URL and every other outbound URL from landing in the sink.

services/safety/feeds.py
Zingzy: Fixed in #318: SAFETY_SHORTENERS_ENABLED, default false, so the refusal is dark until flipped and rolls back by env. The manual feed got SAFETY_MANUAL_FEED_ENABLED (default true) so every published-policy gate has a switch.

routes/legacy/url_shortener.py
Zingzy: Fixed in #318: both legacy routes now surface rejection.public_message under the frozen key names, so the shortener refusal reaches integrators as the documented policy line while security blocks keep the coarse default. Added the missing test: a shortener rejection through POST / asserts the descriptive message.

services/safety/providers.py
Zingzy: Fixed in #318, see the thread above: header key, httpx silenced.

services/safety/feeds.py
Zingzy: Fixed in #318, see the thread above: per-feed switches, shorteners default off. On the self-undoing rollback: with the feed disabled its provider is never built, so re-seeding on boot no longer matters for the refusal itself.

PR #301, opened 2026-08-17
feat: creation-pattern anomaly scoring

Zingzy:
Part of #233. Top of the stack, on #300.

Bulk campaigns are bimodal: roughly half of malicious links never get a single click, and unique-link-per-victim phishing keeps per-link traffic near zero, so traffic-based and report-based triggers structurally miss them. Creation patterns are the one signal external feeds cannot see, and this adds it.

How it works:
- Fixed-window counters on the queue Redis (the hotness detector's math pointed at creation): creates per destination registrable domain and creates per creator IP hash, each at a burst window and a daily window. Counters fire once per window on exact threshold equality, so a sustained campaign produces one event per window, not a storm.
- Counting happens where creation happens: `UrlPolicyService.record_create` is called after every successful insert on all three creation paths, so the anonymous legacy majority is covered. Best effort by contract; it never raises and never blocks a write.
- A domain crossing enqueues the destination into the same analyzer the reports feed (`trigger="pattern"`), reusing the verdict store, enforcement and review notifications unchanged. An IP crossing has no single destination to judge, so it surfaces the creator to the operator channel; automated creator action is deliberately left to a later ban-ladder slice.
- Noise control: verdicts decided by a human never go stale, so marking a popular domain benign once permanently silences its recurring bursts. That gives an allowlist through the verdict store with no extra machinery. System-written verdicts keep the existing re-analysis TTL.
- Thresholds ship as deliberately conservative defaults and are tuned only through environment variables, calibrated by replaying past campaigns.

Degradation: no queue Redis, or safety disabled, means scoring is a no-op with a single boot log. Nothing new is required for self-hosted deployments.

Verified with the full suite and a live two-process run: third create to a fresh domain tripped the burst, the event rode the stream to the worker, an uncertain verdict with the burst context was written and a review notification sent; after marking the host benign as a human decision, an injected repeat event was skipped with a logged reason.


routes/legacy/url_shortener.py
```diff
@@ -224,6 +224,7 @@ def _get(key: str) -> str | None:
 
     legacy_repo = LegacyUrlRepository(db["urls"])
     await legacy_repo.insert(short_code, data)
+    await url_policy.record_create(url, request.client.host if request.client else None)
```
Zingzy, line 227: `request.client.host` here (and at the emoji route) is the raw socket peer, not the client. Behind spoo's CF tunnel + Caddy with no uvicorn forwarded-IP trust, that's the proxy address — the same value for every anonymous create — so the IP-burst counter buckets all legacy creates together, trips instantly, and spams `safety_review` with a meaningless hash while giving no real per-creator signal. The v2 path already does this right with `get_client_ip(request)` (reads `CF-Connecting-IP`/`X-Forwarded-For`); use the same helper on both legacy routes. This is the anonymous majority the PR says L1 is built to cover, so the IP dimension needs to actually work here.

services/safety/analyzer.py
```diff
@@ -56,6 +56,17 @@ async def analyze(self, event: SafetyAnalyzeEvent) -> None:
                     event.host, reason=existing.reason or "previous toxic verdict"
                 )
                 return
+            if existing.decided_by != "system":
```
Zingzy, line 59: Good — this `decided_by != "system"` short-circuit is the sticky-human-verdict / allowlist mechanism #299 was missing (a human BENIGN call permanently silences recurring bursts with no extra storage). Leaving a note so it isn't "simplified" away later: this line is load-bearing for the false-positive-recovery story across the whole stack — a system UNCERTAIN still re-analyzes on TTL, only human calls are sticky, which is the intended asymmetry.

Zingzy (commented):
**Merge-safe, and it quietly fixes a real #299 gap — but the IP-burst half of the detector is wrong on exactly the path the PR says it's for.** Ran the safety + settings + legacy-route + url-service suites against the branch (281 passed). The design reasoning is strong: fixed-window Redis counters on the *durable* queue Redis (correct — the cache Redis's `allkeys-lru` would evict them), exact-equality firing so a sustained campaign yields one event per window not a storm, reusing the reports analyzer/verdict/enforcement unchanged, and best-effort `record_create` that never blocks a write. And the `decided_by != "system"` short-circuit added to the analyzer here is the sticky human-verdict/allowlist mechanism that #299's review flagged as missing — nice that it lands in the same stack, and doubling the human-benign call as a permanent allowlist with no new machinery is elegant.

**The bug (should-fix before enabling L1):** the v2 path feeds `record_create` the real client IP via `get_client_ip(request)` (which reads `CF-Connecting-IP`/`X-Forwarded-For`), but the two legacy routes pass `request.client.host` — the raw socket peer. spoo runs behind the Cloudflare tunnel + Caddy and uvicorn has no forwarded-IP trust configured, so on the legacy paths `request.client.host` is the proxy's address, identical for every anonymous request. The IP-burst counter therefore hashes all legacy creates into one bucket → it trips almost immediately and stays tripped, producing a stream of meaningless `safety_review` notifications and zero real per-creator signal — on the anonymous legacy majority the PR body explicitly names as the reason L1 exists. Use `get_client_ip(request)` on both legacy routes to match v2. (The domain-burst half keys on the destination and is unaffected — it's only the IP dimension that's broken on legacy.)

Everything else checks out: raw IPs never enter Redis keys (sha256-truncated, tested), the `2x window` TTL self-decays, thresholds are conservative env-only tuning, and no-queue-Redis / safety-off both degrade to a logged no-op. Merge ships nothing; fold the IP fix in before `SAFETY_L1` counts anything in prod.

routes/legacy/url_shortener.py
Zingzy, line 227: **Should-fix: this is the proxy's address, not the creator's.**

`request.client.host` behind Caddy and Cloudflare is one constant value for every anonymous legacy request. Uvicorn is started with no `--proxy-headers` / `--forwarded-allow-ips`, and its default only trusts `127.0.0.1`, which a sibling container is not, so `request.client` is never rewritten. The v2 path a few files over uses `get_client_ip(request)` (CF-Connecting-IP, then X-Forwarded-For), and `middleware/rate_limiter.py:204` uses it too.

Consequence with the current thresholds: all legacy creates share one hash bucket, so site-wide legacy volume above roughly 4/min keeps it at or past the burst threshold in nearly every 600s window. That is about 144 review embeds per day, all naming the same hash, and zero per-creator signal on the anonymous legacy majority this feature exists for.

`get_client_ip(request)` here and at the emoji route (line 378). Note the neighbouring `creation-ip-address` field has the same flaw, which is probably where this came from.

services/safety/scoring.py
```diff
@@ -0,0 +1,159 @@
...
+            counts = await self._bump(registrable_domain, client_ip)
+        except Exception as exc:
+            log.warning(
+                "l1_scoring_failed", error=str(exc), error_type=type(exc).__name__
+            )
+            return
+        for (family, window), count in counts.items():
+            # Exact equality: each window fires exactly once (hotness
+            # precedent); a sustained campaign re-fires each new window.
+            if count != self._thresholds[(family, window)]:
```
Zingzy, line 86: **Should-fix: exact equality does not survive a redis-py retry.**

Concurrency is fine here, `INCR` is atomic and you use the returned value, so no value is skipped between racing callers.

Retries are the problem. The client is constructed with `socket_timeout=5` and no `retry=` argument, so redis-py's default `Retry(ExponentialWithJitterBackoff(), retries=3)` applies and `Pipeline.execute` wraps the whole stack in it. If the server applies the increments and the reply is lost (a brief blip, an AOF-rewrite stall), all four `INCR`s are re-sent and the counter jumps 48 to 52 in one logical create. `52 != 50` so nothing fires, and every subsequent create in that window is above the threshold and skipped by this same check.

Net effect: one transient timeout silently disarms that counter for the rest of its window, which is a full day for the daily one. `>=` combined with a `SET NX` marker on a `l1:fired:<key>` key keeps the fire-once property and survives both retries and double counting.

Zingzy (commented):
**Re-review at the same head. The legacy IP bug stands and I can now quantify it, and there are seven new findings, two of which defeat the detector outright. Merge-safe, not enable-safe.**

Confirmed, with numbers: the two legacy routes pass `request.client.host` while v2 passes `get_client_ip(request)`. Uvicorn is started with no `--proxy-headers` and no `--forwarded-allow-ips`, and its default `forwarded_allow_ips="127.0.0.1"` does not match a Caddy sibling container, so `request.client` is never rewritten. Every anonymous legacy create hashes into one bucket. Legacy limits are per real IP at 5/min, so site-wide legacy volume above roughly 4/min keeps that single counter above the burst threshold in essentially every 600s window: **about 144 operator review embeds per day, forever, each naming the same fixed hash.** `middleware/rate_limiter.py:204` already uses `get_client_ip` for exactly this reason. The domain half is unaffected.

The design reasoning I praised last time still holds: counters on the durable queue Redis rather than the LRU cache, exact-equality firing, best-effort `record_create`, and the sticky human-verdict rule are all right. Two of those turn out to be more fragile than they look.

**New: `PATCH` retarget bypasses L1 completely.** `record_create` is wired to the three creation paths only. `_handle_long_url` re-points an existing link at a new destination and runs the L0 gate, but never calls `record_create`. So: create 5,000 links at `example.com` over a week, then retarget all of them to `evil-campaign.tld` inside one minute at the authed limit of 60/min. The domain counter for the destination that now receives all the traffic stays at zero, in every window. L0 still checks each URL, so this defeats the pattern layer specifically, which is the entire PR.

**New: the exact-equality firing can miss, permanently, from a single Redis blip.** Under pure concurrency the check is safe, `INCR` is atomic and the pipeline uses the returned value, so no value is skipped. The problem is retries. The client is built with `socket_timeout=5` and no `retry=` argument, so redis-py's default `Retry(ExponentialWithJitterBackoff, retries=3)` applies, and `Pipeline.execute` wraps the whole command stack in it. A lost reply after the server applied the increments re-sends all four `INCR`s, so a counter can step 48 to 52 in one logical create. `52 != 50`, nothing fires, and every later create in that window is `count > threshold` and skipped by the same equality check. One transient timeout disarms that counter for the rest of its window, which for the daily counter is the rest of the day. A `>=` plus a `SET NX` fired-marker survives retries; exact equality does not.

**New: a queue-Redis wedge adds up to ~27s to every create on all three paths.** `record_create` is a bare await on the request path, and the scorer's try/except only catches after the client has finished timing out and retrying. Against a wedged-but-open Redis that is 4 attempts at 5s plus roughly 7s of backoff before the "best-effort, never blocks" except is reached. This is a new hard dependency on the queue Redis in the create path; before this PR the create path touched Mongo and the cache Redis only. With 2 workers in prod that converts the shortener into a request-holding queue. The link is already inserted at that point, so a client timeout leads to a duplicate create on retry. Give it its own short timeout budget or fire it as a background task.

**New: fixed windows are trivially straddled.** Buckets are wall-clock aligned via `now // window`. 49 creates at 09:59 and 49 at 10:00 is 98 creates in two minutes and the burst threshold of 50 is never reached in either bucket. Repeat every ten minutes and the burst tier never fires. The daily counter still catches it about 25 minutes in, so this costs the burst tier rather than everything. The TTL refresh on every INCR is fine, since the key name encodes the bucket, so there is no window drift, only slightly longer retention.

**New: counting and enforcement key on different things.** Counters key on `registrable_domain`, but the event carries whichever `host` happened to land on the threshold, and everything downstream (verdict lookup, upsert, `block_host`) keys on `host`. A campaign spread over `a1.evil.com` through `a50.evil.com` fires once, names one arbitrary subdomain, and if a feed judges it toxic the enforcer blocks that one host while 49 links keep serving. The counter has already fired for the window, so the next 50 produce nothing. Subdomain fan-out is the standard shape for these campaigns and it reduces L1's yield to one in N. The event already carries `registrable_domain`; nothing reads it.

**New: the human allowlist this PR advertises cannot be created.** `decided_by` defaults to `"system"` and `upsert_verdict` has exactly two production callers, both in the analyzer, neither of which passes it. There is no admin route, ops verb, CLI or migration that writes a human value. So the `decided_by != "system"` short-circuit is unreachable, and the mitigation for popular-domain false positives exists only as a manual `updateOne` against prod Mongo. Also worth knowing: `upsert_verdict` `$set`s `decided_by` unconditionally rather than `$setOnInsert`, so the stickiness is held up by two guard clauses returning early, not by the write being safe. A future third caller silently demotes human decisions.

**New, two smaller consequences of that rule.** A human `BENIGN` verdict silences the analyzer but not the L0 gate, since `ToxicVerdictProvider` only reacts to `TOXIC`, so marking a domain benign does not stop a feed from refusing creates to it. And the guard sits in `analyze()`, which is the funnel for every trigger, so a host marked benign once is never re-analyzed by any trigger at any age, including user reports, with only an info log to show for it. Reports still reach Discord through `report_summary`, so this is lost automation rather than lost visibility.

**Verified clean:** raw IPs never reach Redis keys, logs, the embed or the verdict store. One caveat worth stating: the hash is unsalted sha256 truncated to 16 chars, and the IPv4 space is small enough to invert by brute force, so it is pseudonymisation for key hygiene and not a confidentiality control. `record_create` is genuinely post-insert on all three paths and cannot raise into them.

**Merge safety.** Merging ships nothing. Before L1 counts anything in prod I would fix the legacy IP first (it is one call), then the retry-safe firing, then decide on the retarget path.

routes/legacy/url_shortener.py
Zingzy: Fixed in #318: both legacy routes now pass get_client_ip(request) to record_create, same helper as v2 and the rate limiter. The stored creation-ip-address field has the same flaw but is frozen v1 analytics shape, left alone.

services/safety/analyzer.py
Zingzy: Kept, and #316/#318 preserved the asymmetry: only the human check is sticky, system verdicts still re-analyze on TTL (now with trigger authority so a report outranks a sweep's freshness). Added your regression test for that path.

routes/legacy/url_shortener.py
Zingzy: Fixed in #318, see above.

services/safety/scoring.py
Zingzy: Fixed in #318 with the shape you suggested: at-or-past threshold plus a SET NX l1:fired marker (window-scoped key, 2x TTL), so a retried pipeline that jumps the counter past the threshold still fires exactly once. Two tests pin it: marker-lost means silent, counter-jump still fires.

PR #303, opened 2026-08-17
feat: retroactive sweeps for feed drift and screening coverage

Zingzy:
Part of #233. Top of the stack, on #301.

Everything built so far judges a destination when something happens: a link is created, a link is reported, a domain gets bursty. This adds the missing case, which is the common one for phishing kits: a destination that was clean at creation and turns bad later, or gets listed by a feed days after the links pointing at it were made.

Two sweeps, neither of which ever iterates the link corpus:

- **Feed delta.** Each feed sync now reports which domains are new in that cycle, and the sync immediately looks up the active destination hosts under those domains through the indexed destination fields, enqueueing one analysis per host. The cost is a handful of index lookups per sync regardless of corpus size, and the result is that a domain the world lists at noon takes down every existing link pointing at it minutes later.
- **Recent screening.** An hourly task collapses the recent creation window to its distinct destination hosts, drops the ones that already have a verdict, and screens the rest, capped per run. That guarantees every destination on the platform ends up screened rather than only the ones that were reported or noticed, which also gives an honest answer to "how quickly do you look at new destinations".

Supporting decisions in here:

- Sweep-triggered screenings that end unresolved record their verdict **silently**. Reports and bursts still ask for human eyes; coverage screening must not page an operator for every innocent new destination.
- Sweeps are registered like feeds: a catalog function builds them, so a future sweep rule is one factory plus one list entry and no changes to the scheduler, wiring or worker. The feed-delta sweep has no schedule of its own by design, since the sync already holds the delta.
- The window, the per-run cap and the on/off switch are settings; the feed-delta sweep needs none.

One fix included, caught by running it rather than by tests: the async driver returns a coroutine from `aggregate`, so calling `to_list` directly returned nothing and the delta sweep quietly swept zero hosts while reporting new domains. Both sweep queries now go through the repository's aggregate helper, a failed lookup logs at error level, and tests pin the awaited call.

Verified with the full suite and a live two-process run: a link created against a clean destination began serving 451 within seconds of that domain appearing in the feed, with a toxic verdict recorded and attributed to the sweep; the screening sweep judged four novel hosts and sent no notifications.


Zingzy (commented):
**Merge-safe, but it introduces the one bug in this whole stack that I'd call a silent security regression: the screening sweep suppresses user reports.** Ran the safety/repo/settings suites against the branch (246 passed). Note CI only ran `check_formatting` here — the test matrix doesn't run on a non-`main` base, so all three of these stacked PRs have had zero CI test coverage; I ran them locally instead.

The design is right and the "never iterate the corpus, iterate the small set and use the dest index" discipline holds: the delta sweep rides `dest.registrable_domain` (indexed), the recent sweep rides `_id` (ObjectId time prefix, no new index). Catalog-as-code matches the feeds discipline. The `aggregate`-not-awaited fix you found by running it is exactly the class of bug tests don't catch. And narrowing the shortener seed — dropping `t.co`, `lnkd.in`, `buff.ly` because the platform generates them on share rather than the user choosing them — is a genuinely better rule than what I reviewed in #300, and it retires most of that false-positive surface.

**The blocker.** Screening writes an UNCERTAIN verdict for *every* destination host created in the last 48h. The analyzer's freshness short-circuit then skips any host with a system verdict newer than `reverdict_ttl_hours` (24h). So: sweep screens `newphish.com` at T, a victim reports it at T+2h, and the analyzer returns at the freshness check — no providers, no notification, nothing. Before this PR only reported/bursty hosts had verdicts, so the first report always ran. Now the *first report of any fresh destination* is swallowed for 24 hours, which is precisely the fresh-campaign case reports exist to catch. I verified this is unchanged at the top of the stack, where it's worse: at #305 the report also never reaches the admission gate, so it never triggers deep investigation either — #305's "reports are always admitted, the P0 lane" is false for any host the sweep touched first. The verdict doc already stores `trigger`, so the fix is available: don't let a `sweep` verdict short-circuit a `report`/`edit`/`pattern` event.

**The escalation to weigh.** Before this PR a threat-feed entry only refused *new* creates. The delta sweep makes a feed entry retroactively block *existing* links, automatically, with no human. The empty-download guard protects against fishfish being down, not against fishfish being wrong — and one overbroad entry for a shared platform now fans out to up to 200 hosts of automatic enforcement. Given that #304 adds reversibility but nothing yet calls it, I'd want the unblock path wired before this runs against a live feed.

Three smaller things inline. Also: the app and the worker both register `safety-recent-screen`, but with different sinks (stream vs inline) — whichever process wins the Mongo claim determines whether the sweep enqueues or executes. Harmless today since the runtimes are meant to be exclusive, but it's a surprising thing to inherit.

services/safety/analyzer.py
```diff
@@ -125,6 +125,13 @@ async def analyze(self, event: SafetyAnalyzeEvent) -> None:
             sample_url=event.url,
             context=event.context,
         )
+        if event.trigger == "sweep":
```
Zingzy, line 128: Enable-blocker, and it's an interaction rather than anything wrong on this line. Silencing sweep notifications is correct — but the UNCERTAIN verdict this path just wrote is what breaks reports. Sweep screens every destination created in the last 48h; the freshness check at the top of `analyze` then skips any system verdict younger than `reverdict_ttl_hours`. A user report arriving in that 24h window returns before the provider chain and before `safety_review`, so the report is silently dropped. At the top of the stack it also never reaches the admission gate, so #305's "reports always admitted" doesn't hold either.

The data to fix it is already on the doc: gate the short-circuit on trigger authority, e.g. skip only when the incoming trigger is no more authoritative than `existing.trigger` (`sweep` < `pattern` < `report`/`edit`). Worth a test that asserts a report re-analyzes a host the sweep screened an hour ago — that's the regression that would have caught this.

services/safety/sweeps.py
```diff
@@ -0,0 +1,151 @@
...
+            except Exception as exc:
+                log.error(
+                    "sweep_delta_lookup_failed",
+                    domain=domain,
+                    error=str(exc),
+                    error_type=type(exc).__name__,
+                )
+                continue
+            for host, sample_url in hosts:
+                await self._sink.emit(
```
Zingzy, line 65: Should-decide, and it's the biggest posture change in the PR. Until now a fishfish entry only refused new creates at the L0 gate; this makes it retroactively block every existing link to that domain's hosts, automatically. The `_FISHFISH_MIN_SANE` floor guards against the feed being *down*, not against it being *wrong* — and a community feed listing one shared platform (a `*.pages.dev`, a free host) fans out to up to 200 hosts of enforcement with no human in the loop, each one a `block_host` that also refuses all future links there.

I'm not saying don't do it — retroactive coverage is the whole point. But I'd want at least one of: a cap on hosts auto-enforced per sync with the remainder going to review, or feed-delta hits routed to review rather than straight to enforcement for domains whose host count is above some threshold (a shared platform looks exactly like "one domain, many hosts" in this query). And I'd wire the `unblock` path from #304 before this points at a live feed.

repositories/url_repository.py
```diff
@@ -324,6 +324,50 @@ async def list_active_owned_by_dest_host(
         docs = await cursor.to_list(length=limit)
         return [UrlV2Doc.from_mongo(d) for d in docs]
 
+    async def list_active_hosts_by_registrable(
```
Zingzy, line 327: Nit: `limit=200` truncates silently. The recent-screen sweep logs `sweep_recent_screen_capped` when it drops work; this one doesn't, so a newly-listed domain with more than 200 active destination hosts is partially swept and nothing says so. Same one-line `log.warning` treatment would make the two consistent.

repositories/feed_domain_repository.py
```diff
@@ -29,16 +29,27 @@ class FeedDomainRepository(BaseRepository[None]):
...
-        """Swap *feed*'s set to exactly *domains*. Returns (kept, purged)."""
+    async def replace_feed(
+        self, feed: str, domains: Iterable[str]
+    ) -> tuple[int, int, set[str]]:
+        """Swap *feed*'s set to exactly *domains*.
+
+        Returns (kept, purged, new_domains) — the third element is the
+        domains that were NOT in the feed before this sync, which is the
+        feed-delta sweep's input: existing links pointing at a domain the
+        world just listed."""
```
Zingzy, line 40: Nit: `distinct("domain", ...)` on every sync is fine for fishfish (~43k domains, well under a megabyte) but `distinct` returns a single BSON array capped at 16MB, so a larger feed would start throwing here rather than degrading. If you add a feed in the URLhaus size class (~1M domains), this needs to become a projected `find` cursor. Worth a note now so the next feed doesn't discover it at 3am.

Zingzy (commented):
**Re-review at the same head. My blocker stands on the mechanism, but I overstated its consequence and one of my smaller findings was simply wrong. Eight new findings, and two of them are enable-gates.**

**Correcting myself, twice.** First, I said a report on a swept host produces "no notification, nothing." That is too strong. `ReportIntakeService` stores the report and fires `report_summary` to Discord before it ever reaches the analyzer, so a human does see the report. What is lost for up to 24 hours is the provider re-analysis (including Web Risk, whose answer changes hours after a kit goes live), the per-host review embed with trigger context, and any automatic enforcement. Second, I claimed the app and worker both register `safety-recent-screen` with different sinks and whichever wins the claim decides behavior. That is wrong: the two runtimes are mutually exclusive by config, and the losing process never starts its scheduler at all, so it never writes the doc. Withdrawn.

The blocker mechanism itself is confirmed exactly as described. `analyzer.py` reads `event.trigger` for the first time at line 128, after all three early returns, so there is no trigger-based exemption from the freshness short-circuit. The sweep writes `UNCERTAIN` with `decided_by="system"` for every dest host created in the last 48 hours, and the freshness TTL is 24 hours, so a report landing in that window is short-circuited. The verdict doc already stores `trigger`; letting a `report` or `edit` event override a `sweep` verdict is the narrow fix.

The design is still right, and the discipline of iterating the small set against the dest index rather than the corpus holds throughout. The `aggregate`-not-awaited bug you caught by running it is exactly the class tests do not catch. And narrowing the shortener seed to drop the platform share wrappers is a better rule than what I reviewed in #300.

**New, enable-gate: the first fishfish sync sweeps the entire feed.** `FISHFISH_FEED` declares no seed loader, so `safety_feed_domains` is empty until the first sync. `replace_feed` computes `new_domains` as everything not already present, which on an empty collection is every domain in the download, and the floor comment puts that in the thousands. The delta sweeper then runs one aggregation per domain, sequentially, then `block_host` for every host it finds, and each of those is currently unindexed (see #304). This recurs any time the collection is restored from backup or wiped. If that run exceeds the 600s lease the row becomes re-claimable while the first run is still going, which voids the scheduler's own no-double-run guarantee. Skipping the sweep when the feed was previously empty is the obvious guard.

**New, enable-gate: the recent-screen sweep blows through the Web Risk free tier by about 7x.** The docstring says cheap providers only reach these hosts and the cost is "milliseconds," on the strength of a deep-tier admission gate that does not exist at this head. The sweep feeds the same analyzer, whose chain ends in `WebRiskProvider`, and by construction it is reached for every novel benign host because the pattern and feed providers abstain on exactly those. At 1,000 enqueues hourly that is roughly 720,000 lookups per month against the 100k/month tier that `providers.py:82` explicitly sizes itself against. Once quota is exhausted the provider returns non-200 and abstains, so the sweep silently converts the platform's only online reputation source into a permanent abstainer for every trigger, and each host it burned quota on now carries a fresh verdict that blocks re-analysis for 24 hours.

**New: the recent sweep truncates at 20,000 hosts, silently and stably.** `list_recent_destination_hosts` applies `$limit` after `$group`, so it bounds output rather than work, and `_aggregate` passes no `allowDiskUse`. Nothing logs the truncation, and the cap-exceeded warning measures a different number further down. Group output order is unspecified, so a host that lands past position 20,000 is never screened at all, with no signal, while `hosts_seen: 20000` reads as a healthy run.

**New: the sweep is structurally blind to v1 and emoji.** The window mechanism is the ObjectId time prefix, which only `urlsV2` has, and `SweepDeps` carries only `UrlRepository`. Legacy creation is still live and still stamps `dest`. So an abuser using `POST /` exclusively is never screened, and the "nothing sits permanently unjudged" property does not hold for the two generations that #304 exists to enforce against.

**New: the feed-delta sweep misses every subdomain-granularity feed entry.** The sweeper pivots on `dest.registrable_domain`, but the create gate checks the exact host first and the registrable domain second. When a feed lists `login.stealer.co.uk`, the gate refuses new creates while the sweep matches zero existing links, because their stamped registrable domain is `stealer.co.uk`. The two halves of the same feed disagree about what "listed" means. The correct query needs `dest.host`, which is the field with no index, so this gap and the #304 P0 share a root cause.

**New: the delta sweep can overwrite report evidence and suppress the review ping.** `recent_screen_task` filters out already-verdicted hosts; `FeedDeltaSweeper.sweep` does not. `upsert_verdict` `$set`s `trigger`, `context`, `sample_url` and `reason` unconditionally. So a reported host whose verdict has aged past the TTL can be re-upserted by an unattended job with `trigger="sweep"` and the sweep's own context, dropping the reporter's reason and reported codes from the operator's only triage surface, and the sweep path then returns before `safety_review`.

**New: the sweeper is `None` in two configurations and says nothing about it.** The worker builds `FeedDeltaSweeper` only when the safety sink exists, and skips `build_sweep_tasks` under the same condition, with no warning, unlike the sibling degradation in wiring which does warn. Worse combination: safety on, scheduler embedded, queue Redis configured, no worker deployed. The app's scheduler then enqueues 1,000 events an hour to a stream with no consumer, they rotate out at maxlen, no verdict is ever written, and every run reports `enqueued: 1000` and looks healthy.

**New, small: the shortener list change is a no-op on any existing deploy.** Dropping `t.co`, `lnkd.in` and `buff.ly` from `data/shortener_domains.txt` only takes effect on an empty feed, since `ensure_feed_seeds` skips a feed with rows and there is no sync task and no migration. On prod, users sharing a tweet keep getting refused after the PR that says they should not. The behavior differs only on a virgin database, which is to say in the test suite.

**New, small:** `skipped` is logged but deliberately left out of the returned detail dict, so the operator-facing `last_run.detail` never shows that the sweep capped. Given the truncation above it, that is the one number worth persisting.

**Checked and clean:** the ObjectId window math has no timezone bug (pymongo handles the aware datetime correctly), the `$in` batch size is fine, and `distinct` on the feed is well under the reply cap at current sizes.

**On merge and CI.** Still worth flagging that only `check_formatting` runs here, because the test matrix does not run on a non-`main` base, so these stacked PRs have had no CI test coverage. I ran the full suite at the top of the stack instead: 3308 passed.

services/safety/analyzer.py
Zingzy: Fixed in #318: the freshness short-circuit is gated on trigger authority (sweep < pattern < report/edit), so a report re-analyzes a host the sweep screened an hour ago. Your suggested regression test is in: report re-analyzes, and the inverse asserts a sweep never re-analyzes a fresh report verdict.

services/safety/sweeps.py
Zingzy: This resolved more cleanly than either option: #316 removed screening's power to originate host blocks entirely. A feed-delta hit now blocks only the judged link and escalates the host-wide decision to the investigation tier, which must justify scope (host_usage breadth) before code widens anything, so a community feed listing a shared platform fans out to per-link blocks plus one investigation, not 200 host blocks. The unblock path you wanted wired first also exists now (#318).

repositories/url_repository.py
Zingzy: Fixed in #318: logs feed_delta_hosts_truncated with the domain and limit, same loud-truncation contract as the recent-screen sweep.

repositories/feed_domain_repository.py
Zingzy: Noted in #318 as a comment on the distinct call, sized against fishfish and naming the projected-find-cursor replacement, so the next feed hits the note instead of the 16MB cap.

PR #304, opened 2026-08-17
feat(safety): block legacy v1/emoji links with a per-link audit trail

Zingzy:
## What

Safety enforcement now covers all three URL collections instead of only urlsV2.

- v1 and emoji documents gain a reversible `blocked` flag plus `blocked_at` / `blocked_reason` stamps. An absent flag means active, so existing documents need no backfill.
- Blocked legacy links serve the same 451 page as blocked v2 links, through the existing resolve path. No route changes.
- The same `blocked_at` / `blocked_reason` audit pair now lands on v2 status flips. `updated_at` is overwritten by any later edit; these stamps survive, and they become the only reason of record once per-link blocking exists.
- The enforcer flips and cache-evicts v1/emoji links in the same collect, flip, evict order as v2, instead of counting them for manual cleanup. Emoji cache eviction canonicalizes the VS16 form.
- `unblock` reverses a block completely; the caller owns cache eviction, same contract as bulk-delete.

## Why

Legacy links are the majority of the corpus and until now the only enforcement available for them was manual deletion: irreversible, evidence-destroying, and a bare 404 for anyone who clicks. Blocking preserves the document, tells the visitor why the link died, and can be undone if a verdict turns out wrong.

## Verified

- 3231 unit and integration tests green.
- Live loop end to end: legacy + emoji links created clean, destination pattern added, report submitted, analyzer verdict, enforcer flip, both links serving 451 within seconds, audit stamps on both docs, unblock + cache evict restoring the 302.


repositories/legacy/legacy_url_repository.py
```diff
@@ -102,11 +103,55 @@ async def check_exists(self, short_code: str) -> bool:
...
+        self, host: str, *, limit: int = 50_000
+    ) -> list[str]:
+        """Short codes of not-yet-blocked links pointing at *host*."""
+        cursor = self._col.find(
+            {"dest.host": host, "blocked": {"$ne": True}}, {"_id": 1}
+        ).limit(limit)
+        docs = await cursor.to_list(length=limit)
+        return [d["_id"] for d in docs]
+
+    async def block_by_dest_host(self, host: str, *, reason: str) -> int:
```
Zingzy, line 125: P0 before enabling: this filter (and the `list_unblocked_ids_by_dest_host` above it, and the four equivalents on the emoji and v2 repos) keys on `dest.host`, which has no index anywhere in the stack. Only `dest.registrable_domain` is indexed, from #297. So each `block_host` does six unindexed scans across `urlsV2` + `urls` + `emojis` — and `urls` is the biggest collection you have — on a box that's concurrently serving the redirect path. #303's feed-delta sweep can trigger this up to 200 times per feed sync.

One line in the `indexes.py` loop that already creates `dest_registrable` fixes it; a compound `[("dest.host", 1)]` (sparse, same rationale as the existing one) is enough since every one of these is an equality match. Worth also correcting the comment in `url_repository.py` that reads as if `dest.host` were already covered.

repositories/legacy/legacy_url_repository.py
```diff
@@ -102,11 +103,55 @@ async def check_exists(self, short_code: str) -> bool:
...
+                "$set": {
+                    "blocked": True,
+                    "blocked_at": datetime.now(timezone.utc),
+                    "blocked_reason": reason,
+                }
+            },
+        )
+        return int(result.modified_count)
+
+    async def unblock(self, short_code: str) -> bool:
```
Zingzy, line 142: This closes the reversal gap I raised on #299 at the data layer, which is the important half — thank you. Two things keep it from being operationally real:

1. `UrlRepository` has no `unblock` counterpart, so v2 links still can't be reversed at all. A false-positive host block flips v2 + v1 + emoji, and only two of the three can be undone.
2. Nothing calls any of these — no service, no route, no ops tool. `git grep` finds only tests.

So an operator hitting a bad verdict still hand-edits Mongo. Given `spoo-ops-mcp` already has the takedown/triage surface, an `unblock_host(host)` there that reverses all three collections and evicts cache would finish this. Until something calls it, I'd keep treating "reversal exists" as not yet true for the pre-enable checklist.

repositories/legacy/legacy_url_repository.py
```diff
@@ -102,11 +103,55 @@ async def check_exists(self, short_code: str) -> bool:
...
+        """Reverse a safety block — the thing deletion could never offer.
+        The audit stamps are removed with the flag: an unblocked link is
+        indistinguishable from a never-blocked one on the wire.
+
+        The caller owns cache eviction (same contract as bulk-delete):
+        without it the cached BLOCKED entry keeps serving 451s until TTL."""
+        result = await self._col.update_one(
```
Zingzy, line 149: Should-decide: `$unset`-ing `blocked_at`/`blocked_reason` means an unblocked link keeps no record that it was ever blocked or why. The docstring six lines up calls those fields "the per-link audit trail" — an audit trail that a reversal erases is the one case where you most want it, since a wrong block followed by a quiet unblock leaves nothing to learn from (or to show someone who asks why their link died for a day).

"Indistinguishable from a never-blocked one on the wire" is the right goal, but the wire shape is a DTO concern, not a storage one. Consider keeping the stamps and clearing only `blocked` (add `unblocked_at`), so the response looks identical while the document remembers.

Zingzy (commented):
**This is the PR I asked for on #299, and the design is better than what I suggested — but it lands one P0 performance gap and stops one step short of actually giving an operator a reversal.** Ran the full suite against the branch: 3159 passed. (CI only ran `check_formatting` — the test matrix doesn't run on a non-`main` base.)

What's right, specifically: `blocked` absent-means-active with a `$ne: True` filter means ten million legacy docs need no backfill, and the same `$ne` makes a re-block leave the *original* `blocked_at`/`blocked_reason` intact, which is the property an audit trail actually needs. The collect-then-flip ordering is preserved across all three collections. Emoji cache keys canonicalize VS16 via `v2_lookup_code` before eviction — that's the kind of detail that would otherwise show up months later as "why is this emoji link still serving." And restructuring the resolve path so BLOCKED applies on every schema while EXPIRED/INACTIVE stay v2-only is the correct shape, with the cache-hit case covered too. Replacing irreversible manual deletion with a reversible flag that serves a 451 instead of a bare 404 is straightforwardly better on every axis: evidence, UX, and recoverability.

**The P0: `dest.host` is not indexed anywhere in the stack.** #297 indexed `dest.registrable_domain` (sparse, all three collections), but every enforcement query keys on `dest.host` — there are 25 such query sites across the three repositories now, and this PR adds six of them. So a single `block_host` runs an unindexed scan across `urlsV2`, `urls` (which this PR's own description calls the majority of the corpus) and `emojis`, twice each: once to collect the invalidation set, once to flip. On one CCX13 that also serves ~400k redirects/day. It gets worse upstack — #303's feed-delta sweep can fire this up to 200 times in a single sync task. The code comment in `url_repository.py` reads "All three ride the sparse dest_registrable/dest.host paths," which suggests the belief that `dest.host` is covered; it isn't. The fix is one line in the existing index loop in `indexes.py`.

**The reversal is half-wired.** `unblock` exists on the legacy and emoji repos, but (a) there is no `unblock` on `UrlRepository` at all, so v2 — the collection with the status machine — still has no reverse, and (b) nothing anywhere calls any of them: no service, no route, no ops tool. So the capability I flagged as a hard pre-enable gate on #299 now exists as an unwired repository method. An operator facing a false positive still can't undo it without writing a script. `spoo-ops-mcp` is the natural home for the tool.

One judgement call inline about `$unset`-ing the audit stamps on unblock. Silent 50k caps on the two new `list_unblocked_ids_by_dest_host` calls carry over the same truncation concern I raised on #299's v2 equivalents — flip is unbounded, collection caps at 50k, and nothing logs when they disagree.

services/safety/enforcer.py
```diff
@@ -61,20 +64,34 @@ def __init__(
...
+        legacy = await self._legacy_repo.block_by_dest_host(host, reason=reason)
+        legacy += await self._emoji_repo.block_by_dest_host(host, reason=reason)
+
+        # 3. Evict Redis per domain namespace. v1/emoji only ever live on
+        #    the system domain; emoji cache keys use the canonical VS16
+        #    form, not the stored ``_id`` variant.
         by_domain: dict[str, list[str]] = defaultdict(list)
         for alias, domain in pairs:
             by_domain[domain or self._system_domain].append(alias)
+        by_domain[self._system_domain].extend(legacy_ids)
```
Zingzy, line 88: **Should-fix: this can hand one unbounded Redis DELETE up to 150,000 keys.**

All three collections now funnel into the single system-domain bucket, and the three collect calls cap at 50,000 each, so `invalidate_many` can be handed 150,000 keys in one blocking command against the same Redis the redirect hot path reads from. Before this PR the ceiling was 50,000.

`infrastructure/cache/url_cache.py` swallows any exception into a log line, so if the command is rejected for size, every one of those entries stays live for the full `redis_ttl_seconds` (3600) and the operator is told the links are blocked.

The edge path already solves this: `CloudflareKVClient.bulk_delete` chunks at `_BULK_MAX_ITEMS = 10_000`. Same treatment here.

Separately, the emoji leg is currently unreachable: emoji links are never cached (`url_service.py:1542-1551` excludes them), so the VS16 canonicalization below and the test asserting it cannot fire in production. The canonicalization is correct, it just has nothing to evict.

repositories/legacy/legacy_url_repository.py
```diff
@@ -102,11 +103,55 @@ async def check_exists(self, short_code: str) -> bool:
...
+            {"_id": short_code, "blocked": True},
+            {"$unset": {"blocked": "", "blocked_at": "", "blocked_reason": ""}},
```
Zingzy, line 151: **Should-decide: this erases the audit trail on the way out, and nothing reads it anyway.**

Two connected facts. `blocked_reason` is never read anywhere at this head: no query filters on it, no DTO projects it, and `link_snapshot` uses an explicit allowlist that omits it. And the only reversal path `$unset`s it along with `blocked_at`.

So the field is write-only and self-deleting, while the v2 twin's docstring argues the opposite case for keeping them (`schemas/models/url.py:206`: "`updated_at` is lossy (any later touch overwrites); these survive").

It matters more than it looks because of the re-block loop: the host verdict stays TOXIC after an unblock, and `analyzer.py:52` re-runs `block_host` on the next event for that host, whose filter `{"blocked": {"$ne": True}}` matches the just-unblocked link. The re-block then writes a fresh `blocked_at`, so the original enforcement time is unrecoverable.

Either keep the stamps and add an `unblocked_at`, or drop the pretence that this is an audit trail.

Zingzy (commented):
**Re-review at the same head. My P0 stands (with corrected numbers) and the unblock findings stand, but I missed the most important one in this PR: blocking a v1 or emoji link does not actually hide its destination. Two public unauthenticated endpoints still serve it and still report the link as active.**

**The blocker I missed.** `ResolvedPublicLink.effective_status()` returns `"expired"` or `"active"` for v1 and emoji and never consults the new `blocked` flag. Its docstring says "v1/emoji docs have no status field and can never be inactive or blocked," which was true before this PR and is exactly what this PR changes. Both public surfaces gate destination disclosure on that string: `services/public_preview_service.py:111-117` reveals the destination when `status == "active"`, and `services/public_stats_service.py:224` serves `long_url` under the same condition. I ran it against the branch with a blocked v1 doc:

```
doc blocked flag      : True
effective_status()    : active
preview reveals dest  : True -> https://credential-harvest.example/login
stats long_url served : https://credential-harvest.example/login
```

So after a CERT report is actioned, `GET /{code}` correctly returns 451 while `GET /api/v1/public/preview/{code}` returns the phishing URL as JSON at 30/min to anyone, and the stats page does the same. Both modules were written around the opposite invariant and say so in their own docstrings: "a public surface must never reveal a destination the redirect would refuse." Fixing `effective_status` to consult `blocked` closes both callers at once, and it is the choke point. This is the one I would fix before merge rather than before the flag flip, because the takedown story is the point of the PR.

**Correcting my own numbers.** I said "25 query sites across the three repositories, 6 added here." The real count is 10 `dest.host` predicates in non-test code, 4 added by this PR. The finding is unchanged and if anything sharper per call: at this head one `block_host()` is **7 unindexed passes**, `urlsV2` three times (two collects plus the flip) and `urls` and `emojis` twice each. It was 5 before this PR. The only index on `urls` and `emojis` is the sparse `dest_registrable` from #297, and I confirmed no `dest.host` index exists anywhere up to the top of the stack. The new `{"blocked": {"$ne": True}}` predicate is additionally non-indexable by construction, since `$ne` against an absent field cannot use a sparse or partial index. The comment at `url_repository.py:295` still reads as though `dest.host` is covered. One line in the existing index loop in `indexes.py` fixes it for the whole stack.

Everything I praised last time still holds and I want it on the record: absent-means-active with `$ne: True` needs no backfill across ten million legacy docs, the same `$ne` makes a re-block preserve the original `blocked_at`, collect-then-flip ordering is preserved across all three collections, the emoji eviction canonicalizes VS16, and a reversible 451 beats an irreversible delete on every axis.

**The reversal is still half-wired, and now it is also unstable.** `unblock` exists on the legacy and emoji repos with zero callers anywhere (no service, no route, no ops verb), and `UrlRepository` has none at all, so v2 still has no reverse and a blocked v2 link cannot be edited or deleted by its owner either. New on top of that: there is no exemption after an unblock. The host's verdict is still `TOXIC`, and `analyzer.py:52` re-runs `block_host` on *every* subsequent event for a toxic host, while `block_by_dest_host` filters `{"blocked": {"$ne": True}}` which matches a just-unblocked link. So a manual reversal is silently undone by the next report, L1 burst or feed-delta sweep touching that host, and because unblock `$unset`s the stamps, the re-block writes a fresh `blocked_at` and the original enforcement time is gone. Reversal needs to write to `safety_verdicts` too, or carry a per-link allowlist.

On the `$unset` judgement call I raised last time, one more fact that settles it: `blocked_reason` is never read anywhere at this head. No query filters on it, no DTO projects it, `link_snapshot` uses an allowlist and omits it. It is write-only, and the only reversal path deletes it. Either read it somewhere or stop pretending it is an audit trail.

**New: `invalidate_many` issues one unbounded Redis DELETE, and this PR triples the key count.** The enforcer now funnels all three collections into the single system-domain bucket, so the ceiling per call is 50,000 plus 50,000 plus 50,000 in one blocking command, against the same Redis the redirect path reads from. It was 50,000 before this PR. The edge path got this right: `CloudflareKVClient.bulk_delete` chunks at 10,000. Any Redis exception here is swallowed into a log line, so a rejected oversized command leaves every entry live for the full hour TTL. Chunking is a few lines and mirrors the existing precedent.

**New: enforcement eviction silently no-ops when the host process has no cache Redis.** The worker treats cache Redis as optional and says so in a comment that predates enforcement moving into the worker. Every `UrlCache` method returns early on `None` with no warning. So a worker with queue Redis but no `REDIS_URI`, which is a supported combination, flips Mongo, logs `safety_host_blocked` with a nonzero `cache_invalidated` computed from the collected set, and evicts nothing, while the app keeps serving from its own Redis for an hour. This PR extends that from v2-only to all three collections.

**The truncation caps carry over and now apply to two more collections.** Four collect methods cap (50k, 50k, 50k and 1k) while all the flips are unbounded, and none of them compares the result length to the limit. The codebase has the right convention two hundred lines up in the same file, with the rationale spelled out: a full cursor means something bypassed the ceiling and the numbers are silently short. A 60k-link campaign host flips 60,000 and evicts 50,000, and the operator sees a `cache_invalidated` number they have no way to know is wrong.

**Small: the emoji eviction branch is unreachable.** Emoji links are never cached (`url_service.py:1542-1551` excludes them explicitly), so the VS16 canonicalization at `enforcer.py:88-91` and the test pinning `evicted == ["⭐"]` assert something production cannot exercise. The canonicalization itself is correct and consistent with `resolve()`, so this is dead code rather than a bug. Also `count_by_dest_host` is now callerless on both legacy repos; the docstrings were updated instead of the methods removed.

**Checked and clean:** HEAD requests go through the same resolve path and 451 correctly. `prior_status` genuinely does not matter for v1 and emoji, since they have no stored status and expiry is derived on read, and for v2 it is moot because the flip filters on `ACTIVE`. Concurrent block and unblock have no lost-update window, both put the target state in the filter. `blocked_reason` does not leak to users; the destination exposure above is via `url`/`long_url`, not the new fields.

repositories/legacy/legacy_url_repository.py
Zingzy: Fixed in #318: sparse dest_host index on all three collections in the same indexes.py loop, and the misleading comment in url_repository.py now says which index serves which pivot.

repositories/legacy/legacy_url_repository.py
Zingzy: Both halves closed in #318: UrlRepository.unblock_by_dest_host (scoped to docs carrying blocked_reason so manual bans stay untouched) and SafetyEnforcer.unblock_host reversing v2 + v1 + emoji with cache eviction after the flip. The ops MCP verb that calls it is the remaining piece and lives in the other repo; agreed the pre-enable checklist keeps 'reversal exists' unticked until that lands.

repositories/legacy/legacy_url_repository.py
Zingzy: Fixed in #318: unblock now unsets only the flag, keeps blocked_at/blocked_reason, and stamps unblocked_at, on the single-code path and the new host-wide reversals, v2 included. The wire stays clean because resolution only reads the flag/status, exactly the DTO-vs-storage split you described.

services/safety/enforcer.py
Zingzy: Fixed in #318: invalidate_many chunks at 10,000 keys per DELETE, same treatment as the edge KV client. On the unreachable emoji leg: kept, since eviction is now status-blind and the canonicalization is correct; it starts mattering the day emoji entries are cached, and it costs one list call until then.

repositories/legacy/legacy_url_repository.py
Zingzy: Fixed in #318, see the thread above: stamps survive, unblocked_at added. Your re-block loop point is also why the stamps matter less as a loss now, but they no longer get erased regardless.

Zingzy (commented):
**Re-checked at the fixed stack tip (`3dd506d7`). The blocker from my last review is not fixed, and the surface is wider than I reported. Everything else here has moved in the right direction.**

`ResolvedPublicLink.effective_status()` is byte-identical to this PR's head. It still never reads the `blocked` flag and its docstring still asserts v1/emoji "can never be inactive or blocked", which this PR is precisely what makes false. I re-ran it against the tip:

```
doc blocked flag      : True
effective_status()    : active
preview reveals dest  : True -> https://credential-harvest.example/login
stats long_url served : https://credential-harvest.example/login
```

`raw_v1` is the full unprojected document, so `blocked` is sitting in memory and simply not consulted.

I said two surfaces last time. The complete list is seven, and four of them never touch `effective_status`, so fixing that one function is necessary but not sufficient:

1. `GET /api/v1/public/preview/{code}`, v1 branch
2. Public stats page and API, via `long_url`
3. Legacy HTML preview `GET /{code}+`, which deliberately bypasses `resolve()` status checks so that blocked and expired links can still render preview information, and reads `doc.url` with no `blocked` check
4. Legacy stats `GET/POST /stats/{code}`
5. Legacy export `GET/POST /export/{code}/{fmt}`, in CSV, XLSX, JSON and XML
6. `get_stats_pipeline` in `routes/legacy/helpers.py` projects `url` but not `blocked`, `blocked_at` or `blocked_reason`, so items 4 and 5 could not gate on it even if they tried
7. For contrast, the v2 preview branch is correct, because `v2_effective_status` does read BLOCKED

There is no test anywhere combining a blocked v1 link with a public surface, which is why the suite stays green.

**What did get fixed, credited properly.** `dest.host` is now indexed on all three collections, and the `{"blocked": {"$ne": True}}` residual is servable behind the `dest.host` equality. The collect set is now status-blind, which fixes the strand-on-retry problem I raised on #299. `unblock_by_dest_host` and `enforcer.unblock_host` now exist, audit stamps survive reversal via `unblocked_at` instead of being unset, and cache eviction is chunked at 10,000. Three of the four collectors now warn on truncation.

**Still open beyond the blocker.** `unblock_host` has zero callers outside tests, so the reversal capability is still unreachable for an operator. The legacy and emoji unblock filters are `{"dest.host": host, "blocked": True}`, which clears manual operator bans that predate safety, while the v2 path correctly requires `blocked_reason` to exist; the docstring claims the guard covers all three. `unblock_host` also does not touch the verdict, so the host stays TOXIC and the next report, burst or sweep re-blocks with a fresh `blocked_at`. The edge `bulk_delete` return is still discarded, `invalidate_many` still swallows per-chunk failures while the count is reported as evicted, and `list_active_owned_by_dest_host` and `list_active_owned_by_aliases` still truncate at 1,000 with no warning.

Full detail and the new findings are on #318, where the fixes live.

PR #306, opened 2026-08-19
feat: self-serve account deletion with scheduled erasure

Zingzy:
Users can now delete their account themselves, which the terms of service have promised for a while. Prompted by an erasure request that had to be handled by hand: there was no deletion path anywhere in the product, and support-side removal meant a manual pass over every collection.

Design, briefly:

- Two phases. `DELETE /api/v1/me` re-authenticates and flips the account to `PENDING_DELETION` with a `purge_after` deadline (grace window, default 7 days, `ACCOUNT_DELETION_GRACE_DAYS`). A recurring `account-erasure-sweep` task (every 10 minutes, on the scheduler from #298) hard-erases accounts past their deadline. The users collection is the queue: no job documents, and since every cascade step is a `delete_many` or `$pull` and the user doc goes last, at-least-once execution is safe. A crashed run simply re-queues the account.
- Re-auth: password accounts must present the password (a stolen session alone cannot delete an account), OAuth-only accounts type their exact email. Failures return a uniform 403.
- During grace, login is blocked on both the password and OAuth paths with `X-Error-Code: ACCOUNT_PENDING_DELETION`, checked only after credentials are proven so deletion state cannot be probed. `POST /auth/restore` cancels with valid credentials; unknown email, wrong password and not-pending all share one 403 so accounts cannot be enumerated.
- The cascade removes: links (with per-link cache and edge purge), custom domains, click analytics (timeseries, metaField-only predicate), API keys, verification tokens, page layouts, app grants, webhook endpoints/events/deliveries, report submissions, reporter references and flag allowlist entries, R2 objects under the owner prefix, and the PostHog person (env-gated, best-effort). The email address and the R2 prefix are captured before anything is deleted; the user document is deleted last.
- Two transactional emails (deletion scheduled, deletion completed) ride the existing ZeptoMail provider; unconfigured deployments fall back to no-ops.
- Refresh requires an ACTIVE account, so the status flip revokes rotation immediately; already-issued access tokens live out their short expiry.

Also corrects the `DELETE /urls/{url_id}` docs, which claimed click analytics are removed with the link. They never were; account deletion is now the path that erases them.

Stacked on #298 (the sweep is a scheduled task). Verified with the full suite plus a live end-to-end run: request, blocked login, restore, re-request with zero grace, sweep, and a per-collection audit confirming the erased account left nothing behind while a second account's data was untouched.

## Summary by CodeRabbit

* **New Features**
  * Added account deletion requests with password or email confirmation and rate limiting.
  * Added a grace period and one-time email links for restoring accounts.
  * Added scheduled permanent erasure of account data, links, analytics, and integrations.
  * Added deletion-request, cancellation, and completion email notifications.
  * Blocked password, OAuth, and API-key access while deletion is pending.
  * Added optional analytics erasure and improved object storage cleanup.
  * Preserved uploaded content paths when storage credentials change.

* **Documentation**
  * Clarified that deleting an individual URL does not remove its analytics data.


services/account_erasure_service.py
```diff
@@ -0,0 +1,279 @@
...
+    async def erase(self, user_id: ObjectId) -> dict[str, int]:
+        """Run the full cascade for one account. Returns per-step counts.
+
+        A missing user returns ``{}`` — already erased by a prior attempt.
+        Mongo/R2 failures propagate (the user doc is still there, so the
+        next sweep retries); PostHog and mail failures are swallowed —
+        neither may park an account in PENDING_DELETION forever.
+        """
+        user = await self._user_repo.find_by_id(user_id)
+        if user is None:
```
Zingzy, line 190: **Blocker: no status re-check here, so a restore that lands mid-sweep is erased anyway.**

`sweep()` selects a batch of 25 up front, then erases serially. Each cascade does an R2 list plus per-object deletes, a Cloudflare domain cascade, and a dozen `delete_many` calls, so the gap between selection and a late account's turn is minutes.

In that gap a user can call `POST /auth/restore`. `UserRepository.restore` matches (still PENDING_DELETION), flips to ACTIVE, unsets `purge_after`, and returns 200 "account restored". Then this code re-reads the doc, sees it is not None, and runs the full cascade on an ACTIVE account. `delete_hard` is `self._delete({"_id": user_id})` with no status predicate either.

Every other transition in `user_repository.py` is a guarded filter: `mark_pending_deletion` requires ACTIVE, `restore` requires PENDING_DELETION. The one irreversible operation is the only unguarded one.

Minimum fix here:

```python
if user is None or user.status is not UserStatus.PENDING_DELETION:
    return {}
```

Better, because it also closes the concurrent-runner case: have `find_purge_due` hand out claimed users via `find_one_and_update` into a terminal status, so selection and erasure are one atomic step. The lease is 600s with no renewal and both the app and worker register this task, so two runners on the same batch is reachable today.

services/account_deletion_service.py
```diff
@@ -0,0 +1,187 @@
...
+    def _verify_reauth(
+        user: UserDoc, password: str | None, confirm_email: str | None
+    ) -> None:
+        """Prove the request comes from the account owner, not a session.
+
+        Password accounts re-auth with the password ONLY — accepting the
+        typed-email fallback there would let a hijacked session skip the
+        password. OAuth-only accounts type their exact account email.
+        """
+        if user.password_set and user.password_hash:
```
Zingzy: **Blocker: for OAuth-only accounts the re-auth secret is inside the credential being re-authenticated.**

The else-branch accepts the account email as proof. That value is minted into the access token and returned by `GET /auth/me`, next to `password_set`, which tells an attacker which branch applies. So anyone holding a stolen token reads the email and satisfies this check.

That alone would be recoverable if the grace period worked for these accounts. It does not:

- `restore()` above requires `password_hash`, so OAuth-only accounts are rejected
- the OAuth sign-in path refuses to auto-restore
- `request_password_reset` returns early when `password_set` is false, so they cannot bootstrap a password
- `/auth/refresh` is dead once status leaves ACTIVE, so the only window is `set_password` with whatever remains of a 900s access token

The docstring and the shipped `openapi.json` both promise an "OAuth restore flow"; grepping the branch finds no such route. For every Google, GitHub and Discord signup the seven-day grace period is decorative and a leaked token is terminal.

Options: gate OAuth-only deletion behind a fresh provider re-auth rather than a typed email, or ship the restore path (a signed one-shot cancel link in the "deletion scheduled" mail would cover it without a login), or block OAuth-only deletion until one of those exists.

Separately, the guard reads `user.password_set and user.password_hash`, so a document with the flag set and the hash missing falls through to the typed-email branch. I found no writer that produces that state, but `if user.password_hash:` fails closed instead of open.

services/account_erasure_service.py
```diff
@@ -0,0 +1,279 @@
...
+        # 5. Pulls — shared docs stay, the user's identifiers go.
+        counts["reports_pulled"] = await self._report_repo.pull_reporter(user_id)
+        counts["feature_flags_pulled"] = await self._feature_flag_repo.pull_allowlisted(
+            user_id, email
+        )
+        # 6. R2 — profile pictures + og images under the owner prefix.
+        counts["r2_objects"] = await self._sweep_r2(r2_prefix)
+        # 7-8. External systems, then the doc itself — LAST.
+        await self._erase_posthog(user_id)
+        await self._send_confirmation(email)
```
Zingzy: **Should-fix: the "account deleted" mail goes out before the account is deleted.**

`_send_confirmation` swallows its errors, but `delete_hard` on the next line propagates. `sweep()` catches it, the account stays PENDING_DELETION past its deadline, and the ten-minute cron re-runs the whole cascade, sending the same "your account has been permanently deleted" mail again on every tick until the delete succeeds.

This is also the only non-idempotent step in a cascade whose at-least-once safety is otherwise real: `delete_many`, `$pull`, R2 delete with 404-as-success and the PostHog call are all genuinely re-runnable, and no counters are decremented anywhere.

Swapping these two lines fixes both problems, and the mail is already best-effort so a post-delete send loses nothing. The ordering rationale in the module docstring applies to R2 and PostHog, which are irreversible; the mail is not in that category.

Zingzy (commented):
**Merge-safe in the sense that nothing here is wired into an existing path, but not ship-safe: three things can destroy an account that should have survived, and one of them needs only a stolen session. The engineering underneath is careful, so the findings are about specific joints rather than the design. Suite is green (3519 passed).**

I want to lead with what is right, because it is a lot and it shapes how seriously to take the rest. The cascade order is genuinely load-bearing and correct: email and the R2 prefix are captured before anything is deleted, and the user document goes last, so a crash re-queues the account. The anonymous-owner sentinel guard is the single most important safety property in the PR and it works. I ran it:

```
urlsV2   ANONYMOUS_OWNER_ID   -> BLOCKED (owner_id must be a real account id)
urlsV2   None                 -> BLOCKED (owner_id must be a real account id)
urlsV2   real ObjectId        -> ALLOWED
clicks   ANONYMOUS_OWNER_ID   -> BLOCKED (owner_id must be a real account id)
```

Without that, erasing any account would have deleted every anonymous link and click on the platform. The clicks predicate is correctly metaField-only for a timeseries collection, `list_keys` is properly paginated and raises rather than returning a partial listing (with the right reasoning in its docstring), the new partial index exactly matches the sweep's query shape, `mark_pending_deletion` and `restore` are both guarded transitions, and `DELETE /api/v1/me` is `JwtUser` so API keys and delegated app tokens genuinely cannot reach it. Password accounts correctly get no typed-email fallback, and the docstring explains why.

## Blocker 1: a restore that lands during a sweep is erased anyway

`erase()` re-reads the user but only checks whether it exists:

```python
user = await self._user_repo.find_by_id(user_id)
if user is None:
    return {}
```

It never re-checks `status`, and `delete_hard` is unguarded (`{"_id": user_id}`). Meanwhile `sweep()` selects a batch of 25 up front and then erases them serially, each one doing an R2 list plus per-object deletes, a Cloudflare custom-domain cascade, and a dozen `delete_many` calls.

So a user who calls `POST /auth/restore` after the batch was selected but before their turn gets a successful guarded flip to ACTIVE, a 200 response saying the account is restored, and then total erasure minutes later. The window is the batch duration, not a Mongo operation.

The tell is that every other transition in `user_repository.py` is fenced on status. The one irreversible operation is not. One line at the top of `erase()` closes it, though a claim-style `find_one_and_update` into a terminal state would also fix the concurrent-runner case below.

## Blocker 2: OAuth-only accounts can be destroyed by a stolen token, with no way back

Two things combine badly.

The re-auth for an OAuth-only account is typing the account email. That value is in the access token itself and is returned by `GET /auth/me` alongside `password_set`, which tells an attacker which branch to take. So for these accounts the "second factor" is inside the credential being re-authenticated.

And there is no restore path for them. `restore()` requires `password_hash`, so it rejects them; the OAuth sign-in path explicitly refuses to auto-restore; and `request_password_reset` returns early when `password_set` is false, so they cannot bootstrap a password. The only opening is `set_password` with an access token that has at most 900 seconds left, since `/auth/refresh` is already dead once status leaves ACTIVE. The docstring and the shipped `openapi.json` both promise an "OAuth restore flow"; there is no such surface in the backend.

Net effect for every Google, GitHub and Discord signup: a leaked token means irreversible account destruction and the seven-day grace period is decorative. Either ship the OAuth restore path before this reaches users, or require something the token does not already contain.

## Blocker 3: the flip does not revoke API keys

`dependencies/auth.py` contains no reference to `UserStatus` at all, so the API-key branch never consults account status. Session refresh is correctly cut (`credentials.py` checks `status != ACTIVE`) and device sessions and app grants are cut too, which makes this look like an oversight rather than a decision.

An API key therefore keeps full write access for the entire grace window and is only destroyed at erasure. The most common reason someone deletes an account under duress is that a credential leaked, and this is exactly the credential that keeps working.

## Should-fix

**The confirmation mail is sent before the user document is deleted.** `_send_confirmation(email)` then `delete_hard(user_id)`. `delete_hard` propagates, `sweep()` catches and continues, and the account stays PENDING_DELETION past its deadline, so the ten-minute cron re-runs the cascade and sends "your account has been permanently deleted" again. It is also the only non-idempotent step in an otherwise genuinely re-runnable cascade, and the first send is a false attestation until the delete lands. Swapping the two lines costs nothing since the mail is already best-effort.

**R2 objects survive a `SECRET_KEY` rotation and the erasure reports success anyway.** The prefix is `HMAC(secret, owner_id)`, and `image_ingest.py` says outright that "old sweeps need the old secret", but nothing acts on it. I ran it:

```
old SECRET_KEY prefix : ab77e42b69cee5a8
new SECRET_KEY prefix : 8ff38e818e4ed2e0
sweep looks under     : 8ff38e818e4ed2e0
objects stored under  : ab77e42b69cee5a8
=> finds 0 objects, reports success
```

After any rotation, every erasure logs `r2_objects=0`, deletes the user document, and leaves that user's profile picture and OG images live at their public URLs with nothing left to map them back. Those objects are also written `immutable, max-age=31536000` and the cascade issues no CDN purge, so even a correctly-found object keeps serving from the edge after the origin copy is gone.

**Restore sends no notification and needs no session.** It takes email and password on a public endpoint. Someone who has the victim's password, which is the scenario that prompted the deletion, can silently cancel it, and the victim is never told. The same cycle resets `purge_after` each time, so the grace window is indefinitely extendable by whoever holds the password.

**The OAuth callback leaks deletion state before the email-verification gate.** `_ensure_not_pending_deletion` runs on a bare `find_by_email` one line before `_can_auto_link`, which is the only place that requires `provider_info.email_verified`. Three distinguishable outcomes let an unauthenticated caller learn whether an address is registered and whether it is scheduled for erasure, and because the callback is a browser GET the message renders into the error page body. Moving the check inside the verified branch fixes it. The password login path is ordered correctly, so this is OAuth-specific.

**Erasure bypasses the blocked-link guard.** Interactive delete refuses `BLOCKED` links; `delete_all_by_owner` deletes every document for the owner regardless of status. So an abuser whose campaign was swept and blocked can delete their account and, seven days later, the blocked documents, their `blocked_at` and `blocked_reason` audit trail, and the alias reservations are all gone, freeing aliases that are still circulating in phishing mail and on third-party blocklists. Given the takedown history this is the finding I would weigh most after the three blockers.

**Two cascade queries are full collection scans.** `report_submissions` is only indexed on `created_at`, so `delete_by_reporter`'s `$or` scans the whole append-only submission log, and `reports` has no index on `reporter_ids`, so `pull_reporter` scans too. That is two scans per erased account, up to fifty per batch, on the box serving the redirect path.

**The sweep can outlive its lease.** The lease is 600 seconds with no mid-run renewal, the cron is every ten minutes, and a batch is 25 full cascades where a single link-heavy account does one Redis invalidate and one edge purge per link plus one HTTP DELETE per R2 object. When the run overruns, the other process claims the row and `find_purge_due` hands it the same users, because nothing marks work in flight. Both the app and the worker register this task. Nothing derives the batch limit from the lease and nothing raises the lease alongside this PR.

**`delete_all_for_owner` is an unbounded `while True` that always re-reads page zero.** Termination depends entirely on `delete_by_id` actually removing each document, and its boolean return is discarded. If a delete ever no-ops, the loop spins forever inside the erasure, hammering Cloudflare and holding the scheduler slot.

**Pre-claim clicks are never erased.** `meta.owner_id` is stamped at click time and `claim` does not backfill it, so clicks on a link created anonymously and later claimed keep the anonymous sentinel. The cascade correctly refuses that sentinel, so those rows survive while their parent link is deleted, leaving dangling documents that still carry visitor IPs. Both `DELETE /api/v1/me` and the management docs promise analytics go with the account.

## Should-decide

**`ACCOUNT_DELETION_GRACE_DAYS=0` is a valid production value.** The comment scopes it to integration smoke tests but nothing enforces that, and this repo already has the pattern for exactly this case: `config.py` refuses to boot when `CUSTOM_DOMAINS_MOCK_DCV` is set in production. With zero grace the account is erased on the next ten-minute tick, and combined with the OAuth finding above that makes deletion immediate and unconditional for most users. A one-character typo is unrecoverable.

**Restore is not bounded by `purge_after`.** It filters on status only, so an account whose cascade aborted midway (links and domains already destroyed, user document still present) can be "restored" to an empty account with a success message.

## Nits

`# D6:` appears twice as an internal requirement label; the constraint after the dash is the part worth keeping. The seven numbered step comments in `erase()` mostly restate the call beneath them, though numbers 3, 5 and 7 carry real constraints worth attaching to their lines.

`AccountErasureService` is built two ways: a shared factory for the worker and an inline construction in `wire_services`, with fourteen positional repository arguments. The app path passes already-wired services and the worker builds its own, so two paths are defensible, but the recent "pass url_policy in worker-side erasure wiring" commit is evidence this already drifted once, and a mis-ordered positional repo in a destructive cascade fails silently. Keyword-only arguments would make the next drift a type error.

## The test I would add

Coverage is broad and the partial-failure tests are genuinely good, but the repository tests assert filter dictionaries rather than outcomes, which cannot catch a predicate that over-matches. There is no test anywhere that another account's data survives an erasure. One integration test with two populated accounts, erasing A and asserting B is untouched, is worth more than the rest of the mock suite for this feature, and it is the property everything else is trying to protect.

One thing outside this PR that it inherits: `rate_limit_key` hashes the raw `Authorization` header without verifying it, so a random bearer value per request yields a fresh bucket. That makes the 3/hour budget on the public `/auth/restore` endpoint, which performs a real argon2 verify and returns a distinguishable success, considerably softer than the comment above it assumes. Pre-existing, but this PR adds an attractive target to it.

infrastructure/storage/r2.py
```diff
@@ -117,6 +137,54 @@ async def put_object(self, key: str, data: bytes, *, content_type: str) -> str:
...
+                host=self._host,
+                path=path,
+                query=query,
+                payload_hash=hashlib.sha256(b"").hexdigest(),
+                access_key_id=self._access_key_id or "",
+                secret_access_key=self._secret_access_key or "",
+            )
+            try:
+                resp = await self._http.request(
+                    "GET", f"{self._endpoint}{path}?{query}", headers=signed
```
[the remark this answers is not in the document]
Zingzy: Fixed in e9f7258: R2StorageClient rejects non-https endpoints at construction. Loopback hosts (localhost, 127.0.0.1, ::1) stay allowed for local MinIO.

repositories/token_repository.py
```diff
@@ -139,6 +139,17 @@ async def delete_by_user(
...
+    async def delete_by_user_or_email(self, user_id: ObjectId, email: str) -> int:
+        """Delete every token tied to the user id OR the email address.
+
+        Account-erasure path: pre-signup flows key tokens by email alone,
+        so an id-only sweep would leave the address behind. Returns the
+        number of documents deleted.
+        """
+        return await self._delete_many(
+            {"$or": [{"user_id": user_id}, {"email": email}]}
+        )
```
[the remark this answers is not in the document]
Zingzy: Fixed in e9f7258: the email clause is dropped when the value is empty or missing; the user_id clause always applies.

routes/auth/routes.py
```diff
@@ -237,6 +240,41 @@ async def logout(
...
+
+    **Authentication**: Not required (public endpoint)
+
+    **Rate Limits**: 3/hour
+
+    **Security**: Returns a uniform 403 for wrong credentials, unknown
+    email, and accounts not pending deletion — prevents account
+    enumeration and state probing.
+    """
+    await deletion_service.restore(body.email, body.password)
```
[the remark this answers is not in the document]
Zingzy: Fixed in ae2a1cd: the deletion notice now carries a one-shot cancel link (256-bit secret, only the hash stored, single use, expires with the grace window) and POST /auth/restore accepts it as an alternative proof. OAuth-only accounts can cancel without credentials, and every restore sends a notification mail.

services/account_deletion_service.py
```diff
@@ -0,0 +1,187 @@
...
+                email, and accounts not pending deletion — the endpoint
+                must not become an account-state oracle.
+        """
+        svc_log = log.bind(op="account.restore")
+
+        user = await self._user_repo.find_by_email(email.strip().lower())
+        if user is None or not user.password_hash:
+            svc_log.info("account_restore_failed", reason="invalid_credentials")
+            raise ForbiddenError(_RESTORE_FAILED)
+        if not verify_password(password, user.password_hash):
```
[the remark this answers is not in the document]
Zingzy: Fixed in ae2a1cd: the early-failure paths verify against a fixed dummy hash before returning, so timing no longer distinguishes account types.

services/account_deletion_service.py
```diff
@@ -0,0 +1,187 @@
...
+        typed-email fallback there would let a hijacked session skip the
+        password. OAuth-only accounts type their exact account email.
+        """
+        if user.password_set and user.password_hash:
+            if password and verify_password(password, user.password_hash):
+                return
+        else:
+            supplied = (confirm_email or "").encode()
+            if secrets.compare_digest(supplied, user.email.encode()):
+                return
```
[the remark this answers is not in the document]
Zingzy: Partially addressed in ae2a1cd: the emailed one-shot cancel link makes the grace period effective for these accounts. A stolen token can start the countdown but cannot intercept the cancel path, and the refresh token dies on the status flip, so the attacker locks themselves out too. Requiring a fresh provider re-auth to initiate deletion is deliberately deferred to a follow-up.

services/account_erasure_service.py
```diff
@@ -0,0 +1,279 @@
...
+        next sweep retries); PostHog and mail failures are swallowed —
+        neither may park an account in PENDING_DELETION forever.
+        """
+        user = await self._user_repo.find_by_id(user_id)
+        if user is None:
+            return {}
+        # Captured BEFORE any deletion: the confirmation mail and the
+        # email-keyed predicates need these after the docs are gone.
+        email = user.email
+        r2_prefix = owner_key_prefix(user_id, self._key_secret)
```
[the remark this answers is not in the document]
Zingzy: Fixed forward in a2120bf: the owner prefix is pinned on the user document at first upload and erasure prefers the stored value, sweeping both prefixes when stored and computed differ. Objects uploaded before the field existed remain a documented caveat, since past rotations are not recoverable.

services/account_erasure_service.py
```diff
@@ -0,0 +1,279 @@
...
+        counts["reports_pulled"] = await self._report_repo.pull_reporter(user_id)
+        counts["feature_flags_pulled"] = await self._feature_flag_repo.pull_allowlisted(
+            user_id, email
+        )
+        # 6. R2 — profile pictures + og images under the owner prefix.
+        counts["r2_objects"] = await self._sweep_r2(r2_prefix)
+        # 7-8. External systems, then the doc itself — LAST.
+        await self._erase_posthog(user_id)
+        await self._send_confirmation(email)
+        await self._user_repo.delete_hard(user_id)
```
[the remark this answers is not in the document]
Zingzy: Fixed in 39cae33: the confirmation mail goes out only after delete_hard succeeds.

services/account_erasure_service.py
```diff
@@ -0,0 +1,279 @@
...
+        user stays PENDING_DELETION and the next sweep retries. The batch
+        limit bounds one run; the recurring schedule drains any backlog.
+        """
+        due = await self._user_repo.find_purge_due(
+            now=datetime.now(timezone.utc), limit=self._batch_limit
+        )
+        erased = failed = 0
+        for user in due:
+            try:
+                await self.erase(user.id)
```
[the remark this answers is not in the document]
Zingzy: Fixed in 39cae33: erasure claims the account into a new ERASING status via a guarded update requiring a due deadline before any destructive step. Restore matches PENDING_DELETION only, delete_hard requires ERASING, and crashed cascades re-claim their ERASING doc on the next sweep.

openapi.json
```diff
@@ -339,6 +339,99 @@
         }
       }
     },
+    "/auth/restore": {
+      "post": {
+        "tags": [
+          "Authentication"
+        ],
+        "summary": "Restore Account",
+        "description": "Cancel a pending account deletion during the grace period.\n\nValidates email + password against the account and flips it back to\nACTIVE, clearing the purge deadline. Only works while the account is\nPENDING_DELETION (i.e. before `purge_after` from ``DELETE /api/v1/me``\npasses); OAuth-only accounts restore via the frontend's OAuth restore\nflow instead.\n\n**Authentication**: Not required (public endpoint)\n\n**Rate Limits**: 3/hour\n\n**Security**: Returns a uniform 403 for wrong credentials, unknown\nemail, and accounts not pending deletion \u2014 prevents account\nenumeration and state probing.",
```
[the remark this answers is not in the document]
Zingzy: Deliberately not enforced, with reasoning: after 39cae33 the sweep's atomic claim is the point of no return, so a restore landing in the post-deadline pre-claim window still wins, which is the user-friendly boundary. A half-erased account can no longer be restored either, since restore excludes ERASING.

openapi.json
```diff
@@ -339,6 +339,99 @@
...
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ErrorResponse"
+                }
+              }
+            }
+          },
+          "403": {
+            "description": "Forbidden \u2014 insufficient permissions or scope",
+            "content": {
```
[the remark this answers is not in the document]
Zingzy: Fixed in e9f7258 and the spec regenerated.

services/account_erasure_service.py
Zingzy: Fixed in 39cae33 with the claim shape you suggested: a new ERASING status, claimed by a guarded update requiring PENDING_DELETION or ERASING plus a due purge_after, before any destructive step. delete_hard is fenced on ERASING, find_purge_due includes ERASING so a crashed cascade retries, restore stays PENDING_DELETION-only, and both login paths treat ERASING like PENDING_DELETION. The concurrent-runner waste is also bounded now: sweep stops starting new cascades after a configurable time budget (default 480s, under the 600s lease).

services/account_deletion_service.py
Zingzy: Fixed in ae2a1cd via your second option: the deletion notice carries a signed one-shot cancel link (hash at rest, single use, valid for the whole grace window) and POST /auth/restore accepts it as an alternative proof, no login needed. Restores on either path send a cancellation notice, so a silent cancel is visible too. The password_set-without-hash fallthrough now fails closed as you suggested. Fresh provider re-auth for initiating deletion is deferred; with the cancel path live, a stolen token starts a countdown it cannot win, since refresh dies on the flip and the link lands in the owner's inbox.

services/account_erasure_service.py
Zingzy: Fixed in 39cae33, the two lines are swapped: delete_hard first, mail only on success. Your framing was right that the mail was the one non-idempotent step; the repeat-send-per-tick failure mode is covered by a test now.

Zingzy:
All review findings addressed across two pushes (through e9f7258), replies on each inline thread. The ones raised in the review body rather than inline:

**Blocker 3, API keys live through grace:** fixed in ae2a1cd. The key path already loads the owning user, so the gate costs no extra read: a valid key on a PENDING_DELETION or ERASING account now gets the same 403 and X-Spoo error code as login. INACTIVE is deliberately not gated, keeping parity with the login paths.

**Blocked links erased with their audit trail:** fixed in e4be83d. delete_by_owner now excludes BLOCKED docs; the retained docs get their creator PII scrubbed in place (creation_ip, meta updated_ip, password) while status, alias, blocked_at and blocked_reason stay, under the Art. 17(3) abuse-prevention carve-out. Counts surface as urls_blocked_retained. Their Redis cache entries are still invalidated, which turned out to be load-bearing: the cache carries the password hash the scrub removes.

**Pre-claim clicks:** fixed in e4be83d. Erasure now also deletes clicks by meta.url_id in chunks for every enumerated link (retained BLOCKED ones included), so clicks stamped with the anonymous sentinel before a claim no longer survive their owner.

**Restore silent and unbounded:** restores now notify the account email on both paths (ae2a1cd). On the deadline question I went the other way, see the inline reply: the ERASING claim is the point of no return, so a pre-claim restore wins by design.

**OAuth callback probing deletion state:** fixed in 7685dd4, the check moved inside the verified auto-link branch.

**Cascade full scans:** fixed in 21d88c2, reporter_ids on reports and reporter_id plus reporter_email on report_submissions are indexed now.

**Sweep vs lease:** the time budget in 39cae33 stops new cascades at 480s by default; with the claim fence a rare overrun is wasteful rather than harmful.

**Unbounded domain loop:** fixed in 21d88c2, a no-op delete raises instead of spinning and a pass cap bounds the loop.

**R2 prefix vs SECRET_KEY rotation:** fixed forward in a2120bf, the prefix is pinned on the user doc at first upload and erasure sweeps stored plus computed when they differ. Pre-existing rotations stay unrecoverable and the docstring says so, as does the note about edge-cached objects outliving origin deletion until TTL.

**Grace zero in prod:** boot refusal in ae2a1cd, same shape as the mock-DCV guard.

**Nits:** taken in e9f7258, labels dropped, restating comments deleted, the erasure constructor is keyword-only.

**The two-account survival test:** genuine gap, not implemented. The repo has no Mongo-backed test anywhere: integration and smoke suites mock app.state.db, mongomock is not a dependency, and CI provisions no Mongo service. Inventing that infrastructure inside this PR felt wrong. Proposal: add a Mongo service container to the tests workflow in a follow-up and land the A-erased-B-untouched test there; the property is currently covered by predicate-shape assertions plus the live prod-mirror run, where a 6.4M-click account was erased with the bystander account byte-identical after.

The rate_limit_key observation about unverified Authorization headers is acknowledged as pre-existing and worth its own fix outside this PR.

infrastructure/storage/r2.py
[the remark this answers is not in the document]
Zingzy: Declining the stricter version, deliberately. Loopback traffic never leaves the host, so there is no wire for the cleartext to cross; an observer who can read loopback already has local access to the same environment that holds the actual credentials, so the SigV4 signature discloses nothing they lack. The check matches the parsed hostname exactly against localhost, 127.0.0.1 and ::1, so no lookalike-host bypass exists and every remote endpoint must be https. Requiring TLS for local MinIO would add self-signed certificate setup to development for no risk reduction.

repositories/indexes.py
```diff
@@ -37,6 +37,25 @@ async def ensure_indexes(
...
+    }
+    try:
+        await users_col.create_index(
+            [("status", 1), ("purge_after", 1)], **_sweep_index_spec
+        )
+    except OperationFailure as e:
+        if getattr(e, "code", None) != 85:  # IndexOptionsConflict
+            raise
+        # Mongo rejects option edits on an existing name: drop-recreate.
+        await users_col.drop_index("pending_deletion_sweep")
```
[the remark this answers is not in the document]
Zingzy: Fixed in 87abc50: the drop is wrapped to tolerate IndexNotFound (code 27), mirroring the TTL-index guard in the same file. Other OperationFailures still fail startup.

repositories/user_repository.py
```diff
@@ -71,3 +75,109 @@ async def complete_onboarding(
...
+        re-claim still matches. Returns True when the claim holds.
+        """
+        return await self._update(
+            {
+                "_id": user_id,
+                "status": {
+                    "$in": [
+                        UserStatus.PENDING_DELETION.value,
+                        UserStatus.ERASING.value,
+                    ]
```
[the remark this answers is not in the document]
Zingzy: Fixed in 87abc50: claims now stamp erasure_claimed_at and carry an account-level lease (ACCOUNT_ERASURE_CLAIM_LEASE_SECONDS, default 900s). A live cascade's fresh claim can no longer be re-claimed; only expired claims are, and find_purge_due shares the same clause builder so the sweep's view and the claim guard cannot drift.

services/account_deletion_service.py
```diff
@@ -0,0 +1,310 @@
...
+            )
+            return restore_token
+        except Exception as exc:
+            log.error(
+                "account_deletion_restore_token_failed",
+                user_id=str(user.id),
+                error=str(exc),
+                error_type=type(exc).__name__,
+            )
+            return None
```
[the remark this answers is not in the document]
Zingzy: Fixed in 8ddeedd: for OAuth-only accounts the restore token is minted and persisted BEFORE the status flip, with the token expiry computed from the same timestamp the flip stores, so pending always implies a usable proof. Persistence failure aborts the request with the account still ACTIVE; a lost flip race cleans up the just-minted token by hash so a concurrent winner's proof survives. Password accounts keep post-flip best-effort minting since credential restore always exists for them.

config.py
```diff
@@ -694,6 +727,18 @@ def blocked_self_domains(self) -> tuple[str, ...]:
     max_active_api_keys: int = 20
     max_date_range_days: int = 90
     http_client_timeout: float = 5.0
+    # Account deletion (GDPR Art. 17): days between the deletion request
+    # and the erasure sweep purging the account (0 = purge on the next
+    # sweep — integration smoke only), and how many due accounts one
+    # sweep run erases (the */10 cron drains any backlog).
+    account_deletion_grace_days: int = 7
```
Zingzy, line 734: **Still open: zero grace is a legal production value with nothing stopping it.**

The comment above still scopes `0` to "integration smoke only" and the validator still allows it, with no boot guard and no startup warning.

The severity dropped now that the restore link exists for every account type, so this is no longer the "OAuth users have no recourse at all" case. But at zero grace `purge_after` is `now`, the next ten-minute sweep claims the account, and the emailed link has roughly that long to be noticed and clicked. A one-character typo in an env var is still unrecoverable.

The pattern is already in this validator a few lines down:

```python
if self.env == "production" and self.custom_domains.mock_dcv:
    raise ValueError("CUSTOM_DOMAINS_MOCK_DCV must not be enabled in production")
```

Refusing to boot on production plus zero grace, or a loud startup warning if you want to keep the escape hatch, would match it.

tests/unit/repositories/test_erasure_deletes.py
```diff
@@ -0,0 +1,36 @@
...
+        (WebhookDeliveryRepository, "delete_by_user", "user_id"),
+        (WebhookEventRepository, "delete_by_owner", "owner_id"),
+    ],
+)
+@pytest.mark.asyncio
+async def test_erasure_delete_filters_on_the_user(repo_cls, method, filter_key):
+    col = make_collection()
+    col.delete_many = AsyncMock(return_value=MagicMock(deleted_count=5))
+    count = await getattr(repo_cls(col), method)(USER_OID)
+    col.delete_many.assert_awaited_once_with({filter_key: USER_OID})
```
Zingzy, line 35: **Still open: this asserts the filter, not the outcome.**

`col.delete_many.assert_awaited_once_with({filter_key: USER_OID})` proves the code passed the dictionary the test author expected. It cannot catch a field name that both the code and the test agree on, and it cannot catch a predicate that over-matches, which is the one failure mode where this feature does damage that cannot be undone.

Everything else in this PR is now defended by a behavioural test, including the parts I would have expected to stay mocked: the claim race, the mail ordering, the budget exhaustion. This is the remaining gap, and it is the one guarding the worst outcome.

One integration test against a real Mongo would cover it: two accounts, each with links, clicks, api keys, webhook rows, report submissions and a flag allowlist entry; erase A; assert every one of B's documents is unchanged. The anonymous-sentinel guards are genuinely well tested, but they only cover the sentinel, not a plain wrong-owner match.

Worth noting the local docker Mongo already makes this cheap to write, and I used it to verify the claim race for this review.

Zingzy (commented):
**All three blockers are closed and nearly everything else with them. Several fixes are better than what I proposed. Two things are still open, both minor next to what was here before. Suite is 3605 green.**

I checked each fix at the choke point rather than reading the commits, and proved the important one by execution.

## The blockers

**The mid-sweep restore race is closed properly.** You took the stronger of the two options: an atomic `claim_for_erasure` that guard-flips to a new `ERASING` status, rather than the one-line status check I suggested. Proven three ways against a real Mongo:

```
A restore-then-claim : restore=True  claim=False   status=ACTIVE
B claim-then-restore : claim=True    restore=False status=ERASING
C second runner steal: False
```

A restore that lands first wins and the account survives. A claim that lands first makes erasure final. A second runner cannot steal a live claim. Preserving `purge_after` so a re-claim still matches, and letting an expired claim lease re-claim a crashed cascade, are both the right details. This also closes the separate finding about restoring into a half-erased empty account, since `restore` no longer matches an `ERASING` doc.

**API keys now go dark during the grace window.** The check covers `PENDING_DELETION` and `ERASING`, raises the same error surface as the login gate, and logs the blocked key. Worth noting it costs nothing: the user document was already being fetched on that path, so this is a status comparison rather than a new round trip. Documenting the one exception to `get_current_user`'s never-raises contract in its own docstring is the right call.

**The OAuth dead-end is solved, and the ordering detail is better than what I asked for.** A one-shot `deletion_restore` token, hash-only at rest, expiry riding `purge_after` so the link dies exactly when the grace does, consumed atomically via `find_one_and_update` on `used_at: None`, so a raced double-click has exactly one winner. What I did not think to ask for is the asymmetry: for OAuth-only accounts the token is minted *before* the status flip and a mint failure aborts with the account still ACTIVE, while password accounts keep best-effort minting after the flip because credentials always work. That is the correct place to put the care, since for one of those two the link is the only way back.

## The rest

Every other finding is closed, and a few of the fixes are worth calling out because they went past the report.

The confirmation mail now runs *after* `delete_hard` and only when it actually removed a document, with an `account_erase_doc_already_gone` warning otherwise. That is better than the swap I suggested, because it also covers a retry against an already-erased account.

Blocked links are retained and scrubbed rather than deleted, on explicit Art. 17(3) grounds: `owner_id`, status, alias and the `blocked_*` audit fields stay so the enforcement record survives and the alias can never be re-registered, while `creation_ip`, the `meta_tags.updated_ip` mirror and the link password are unset. That is a real legal position rather than a code patch, and I think it is the right one. Once the user document is gone the retained `owner_id` no longer resolves to a person, so it reads as a pseudonymous correlation key for abuse purposes. Worth making sure the privacy policy says blocked links outlive account deletion, since that is now a user-visible commitment living only in a docstring.

The R2 prefix is pinned at upload time and both prefixes are swept when the stored and computed values diverge, which handles the rotation case and the pre-pin case together. The reporter cascades get three indexes, and the comment shows the `$or` subtlety was understood: Mongo only skips the scan when each branch has its own index. The domain loop is bounded by a computed `max_passes` and logs `audit.domain.erase_delete_noop` at error when a delete no-ops, so the condition that would have caused the spin is now loud instead of infinite. Pre-claim clicks are handled by chunked `url_id` deletes plus an owner mop-up, both metaField-only and both scoped to the user's own links.

The lease overrun got a more complete answer than I expected: a 480-second sweep budget at 80% of the scheduler lease that stops *starting* new erasures, a 900-second claim lease sized to exceed the budget plus one cascade, and a measured figure for a heavy prod account behind it. `sweep()` now also returns `deferred` and `skipped`, which fixes the older complaint that a full batch and a huge backlog were indistinguishable in `last_run.detail`.

Two I did not flag in the posted review but which were in the underlying analysis are also fixed: the OAuth callback now checks deletion state inside the `_can_auto_link` verified branch, so an unverified callback can no longer probe it, and `restore` burns a dummy verify on the unknown-email path so timing cannot separate "no such account" from "wrong password". The constructor is keyword-only now, with the reasoning written where the next person will read it.

I also checked the new `ERASING` value against every status gate in the codebase. The paths that need a specific message list it explicitly, and the rest use `!= ACTIVE`, which excludes it by construction. Adding an enum member is exactly where this kind of thing goes wrong and it did not here.

## Still open

`ACCOUNT_DELETION_GRACE_DAYS=0` remains a legal production value with no boot guard and no startup warning, while the comment still scopes it to integration smoke. The severity is lower than it was, because the restore link now exists for everyone, but at zero grace that link lives about ten minutes. The `CUSTOM_DOMAINS_MOCK_DCV` precedent in the same validator is still the shape I would copy.

The cross-account isolation test still is not there. `test_erasure_deletes.py` continues to assert the filter dictionary rather than the outcome, which cannot catch a predicate that over-matches, and that is the one failure mode where this feature does unrecoverable damage. Everything else in this PR is now defended by a test; this is the gap I would close last but would still close.

Neither blocks the merge.

repositories/url_repository.py
```diff
@@ -218,6 +219,80 @@ async def delete_many_by_owner_and_domain(
...
+        identifies the creator as a person goes: ``creation_ip``, the
+        ``meta_tags.updated_ip`` audit mirror, and the link ``password``.
+        Returns the number of BLOCKED docs retained (matched, not
+        modified — an already-scrubbed doc still counts as retained).
+        """
+        if not owner_id or owner_id == ANONYMOUS_OWNER_ID:
+            raise ValueError("owner_id must be a real account id")
+        try:
+            result = await self._col.update_many(
+                {"owner_id": owner_id, "status": UrlStatus.BLOCKED.value},
```
Zingzy: **The retention this implements is undone two steps later on custom domains.**

This scrub is correct, and `delete_by_owner` correctly carries `{"status": {"$ne": BLOCKED}}`. But the cascade's next step is `domain_service.delete_all_for_owner`, which per domain calls `delete_many_by_owner_and_domain`, and that filter is:

```python
{"owner_id": owner_id, "domain": domain}
```

No status exclusion. So every BLOCKED link on one of the owner's own custom domains is hard-deleted immediately after being scrubbed and counted, which means `counts["urls_blocked_retained"]` over-reports and the enforcement record vanishes for precisely the links an abuser is most likely to have parked on a domain they control. The alias reservation goes with it.

The system-domain path holds; only the per-domain one leaks. Adding the same `$ne` guard to `delete_many_by_owner_and_domain` restores the invariant, though it is worth deciding explicitly whether a retained BLOCKED link should also keep its custom domain doc alive, since the alias reservation is meaningless if the domain is released.

Related: the retained docs' clicks are deleted, because `url_result.url_ids` includes the retained ids and feeds `delete_by_url_ids`. If the retention is meant as an audit trail, losing the click history may or may not be intended.

services/oauth_service.py
```diff
@@ -347,6 +375,9 @@ async def handle_callback(
         existing_email_user = await self._user_repo.find_by_email(provider_info.email)
         if existing_email_user:
             if self._can_auto_link(existing_email_user, provider_info, provider_key):
+                # Only AFTER _can_auto_link proved the provider verified this
+                # email — unverified callbacks must not probe deletion state.
+                self._ensure_not_pending_deletion(existing_email_user, provider_key)
```
Zingzy, line 383: **This gate is right here, and the `LINK` branch above it has none.**

Putting `_ensure_not_pending_deletion` inside the `_can_auto_link` verified branch is exactly the fix for the probing leak, and the comment says why.

The gap is the other branch. `action == OAuthAction.LINK` runs the whole link flow without this call and finishes at `_make_tokens`, so an account in PENDING_DELETION or ERASING can complete a link callback and receive a fresh access and refresh pair.

That reopens what the API-key change closed. Three token paths now refuse a pending-deletion account (password login, API key, OAuth login) and this fourth one issues new credentials, reachable with an access token still inside its fifteen minutes. Worth gating before `_link_provider` as well, since linking a provider onto an account mid-erasure leaves a dangling reference once the user doc goes.

infrastructure/storage/r2.py
```diff
@@ -60,7 +83,15 @@ def __init__(
             if endpoint_url
             else f"https://{account_id}.r2.cloudflarestorage.com"
         )
-        self._host = urlparse(self._endpoint).netloc
+        parsed = urlparse(self._endpoint)
+        # SigV4 signs but never encrypts — a plain-http endpoint would put
+        # object bytes and credentials-derived signatures on the wire.
+        if parsed.scheme != "https" and parsed.hostname not in _LOOPBACK_HOSTS:
```
Zingzy, line 89: **Right check, wrong altitude: this turns a self-host misconfiguration into a boot crash.**

The reasoning is sound and I would keep the refusal itself: SigV4 signs without encrypting, so a plain-http endpoint puts object bytes and signatures on the wire.

The problem is that it raises from `__init__`, and `wire_services` constructs the client unconditionally whenever R2 is enabled. `_LOOPBACK_HOSTS` covers only `localhost`, `127.0.0.1` and `::1`, so a compose deployment addressing MinIO by service name (`http://minio:9000`, the normal way to do it) now fails to boot with a message telling them to use localhost.

This repo's posture elsewhere is that optional infrastructure degrades rather than taking the app down. Two options that keep the protection: validate in the settings validator so it surfaces as a configuration error at the same place as the other env guards, or treat it as "R2 not enabled" and log loudly, which matches how a missing R2 credential already behaves.

Zingzy (commented):
**Correction to my previous review, plus four findings I missed on the first pass. One of them is a security hole in a path the deletion gate does not cover.**

## I was wrong about the grace guard

My last review said `ACCOUNT_DELETION_GRACE_DAYS=0` was still unguarded in production. That is false. The guard is at `config.py:915`, three lines below the `mock_dcv` precedent I told you to copy, and it does exactly that:

```python
if self.env == "production" and self.account_deletion_grace_days < 1:
    raise ValueError("ACCOUNT_DELETION_GRACE_DAYS must be >= 1 in production")
```

I looked at the field declaration and the `@field_validator` and stopped there, without checking the `model_validator` where the environment-aware guards live, which is the place I had myself named as the right home for it. Sorry for the noise. That item is closed, and so is my other open one only partially: see the test note at the end.

## New: the blocked-link retention is bypassed on custom domains

This one undercuts a fix I praised. Step one retains and scrubs BLOCKED links and reports the count. Step two then calls `delete_all_for_owner`, which per domain runs `delete_many_by_owner_and_domain`, whose filter is `{"owner_id": owner_id, "domain": domain}` with no status exclusion. Every BLOCKED link the account parked on its own custom domain is hard-deleted seconds after being counted as retained.

So `counts["urls_blocked_retained"]` over-reports, the enforcement record disappears for exactly the links an abuser is most likely to have put on a domain they control, and the alias reservation goes with it. The system-domain path is correct; the custom-domain path is not. The `$ne: BLOCKED` guard that `delete_by_owner` already carries needs to be in the per-domain delete too.

## New: the OAuth link flow mints fresh tokens for a pending-deletion account

The `action == OAuthAction.LINK` branch never calls `_ensure_not_pending_deletion`, and it ends at `_make_tokens`. So an account in PENDING_DELETION or ERASING can complete a link callback and receive a new access and refresh pair.

That reopens the hole the API-key fix closed. Password login refuses, the API-key branch refuses, and the OAuth login branch refuses, but this fourth token-minting path does not, and it is reachable with an access token that is still inside its fifteen minutes. The collision path got the gate; the link path was missed. It needs the same call before `_make_tokens`, and probably before `_link_provider` too, since linking a provider to an account being erased leaves an orphan.

## New: the R2 endpoint guard turns a self-host misconfiguration into a boot failure

The https-or-loopback check runs in `R2StorageClient.__init__`, which `wire_services` calls unconditionally when R2 is enabled. `_LOOPBACK_HOSTS` is `localhost`, `127.0.0.1`, `::1`, so a compose deployment pointing at `http://minio:9000` by service name now raises at startup rather than degrading.

The security reasoning is right, and I would keep the refusal. The issue is where it fires: this repo's stated posture is that optional infrastructure degrades rather than takes the app down, and a self-hoster who had a working MinIO setup gets a crash loop with a message about localhost. Refusing to enable the R2 feature, or validating in the settings validator so it surfaces as a config error, would keep the protection without the boot failure.

## Smaller, from the same pass

The R2 prefix pin only helps accounts that have uploaded since the pin landed, because `set_storage_prefix_if_absent` fires at first upload rather than at signup, and there is no backfill script. For a user whose objects predate the pin, `storage_prefix` is absent, so the sweep falls back to the single computed prefix, which is the original bug for that population. `_sweep_r2` also still returns zero without distinguishing "no objects" from "wrong prefix", so erasure reports success either way.

`TokenRepository.delete_by_user_or_email` is the same `$or` shape you just indexed for the reporter cascades, but `verification-tokens` has no index on `email`, so that step is still a collection scan. Worth the same treatment while the reasoning is fresh.

The restore token is consumed before `_user_repo.restore` runs, so a transient Mongo error on the flip leaves the token spent. For an OAuth-only account that is terminal: credential restore is impossible and a re-request returns 409. Consuming after a successful flip, or restoring the token on failure, closes it. Separately the token travels as a URL query parameter in the email link, which puts it in browser history and any Referer.

## On the test

My previous review said the cross-account isolation test was missing, and that stands. There is still no integration test anywhere for this feature, no real or in-memory Mongo in `tests/`, and the repository tests continue to assert filter dictionaries. Given that this pass found a predicate that deletes retained records, that gap is now demonstrated rather than hypothetical: a two-account fixture erasing A would have caught the custom-domain bypass immediately, because B is not even needed for it, only a BLOCKED link on the erased account's own domain.

Zingzy:
Third review round addressed (384a8a5 and f554de1):

- Blocked-link retention now holds through the custom-domain cascade: the erasure path scrubs and retains BLOCKED docs on the owner's own fqdn, and urls_blocked_retained counts what actually survives. The interactive bulk deletes (?domain= and domain revoke) still hard-delete owner BLOCKED links, unchanged deliberately since they are shipped behavior; flagged as a follow-up because single-link delete refuses BLOCKED.
- The OAuth LINK action is gated on PENDING_DELETION and ERASING, closing the fourth token-minting path.
- The R2 https guard stays strict in the client and degrades at wiring: an insecure endpoint disables R2 with a structured warning instead of crashing self-hosted boots.
- Restore tokens are validated, then flipped, then consumed, so a transient failure no longer burns an OAuth-only account's only proof. verification-tokens.email is indexed for the erasure $or. scripts/backfill_storage_prefix.py pins prefixes for uploads that predate the field; run it before any SECRET_KEY rotation.
- New tests/db suite runs real queries against a real MongoDB (throwaway database per test, real ensure_indexes, the worker's actual composition graph). Test A is the domain-cascade regression: BLOCKED links on the erased account's own custom domain survive scrubbed, counts reconcile against the database. Test B asserts a bystander account's documents are byte-identical after a neighbor's erasure. CI gained a mongo:8 service container; locally the suite skips with a hint when Mongo is down, and fails loudly in CI.

routes/auth/routes.py
```diff
@@ -237,6 +240,59 @@ async def logout(
...
+    """Cancel a pending account deletion during the grace period.
+
+    Two mutually exclusive proofs: ``email`` + ``password`` validates the
+    account credentials, ``restore_token`` consumes the one-shot link from
+    the deletion notice email (the only path for OAuth-only accounts).
+    Either flips the account back to ACTIVE, clears the purge deadline,
+    and sends a cancellation notice to the account address. Only works
+    while the account is PENDING_DELETION (i.e. before `purge_after` from
+    ``DELETE /api/v1/me`` passes); once erasure has started, restore is
+    refused for good.
```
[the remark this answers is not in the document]
Zingzy: Fixed in 0db2c7a: the docstring (and regenerated spec) now name the erasure sweep's claim as the hard boundary, with restores in the deadline-to-claim window succeeding by design.

services/custom_domain_service.py
```diff
@@ -386,6 +390,71 @@ async def delete(
...
+            for doc in page:
+                if self._url_service is not None:
+                    await self._url_service.delete_all_by_domain(owner_id, doc.fqdn)
+                if not await self._edge.announce_revoked(doc.fqdn):
+                    log.warning(
+                        "audit.domain.erase_eviction_failed",
+                        fqdn=doc.fqdn,
+                        domain_id=str(doc.id),
+                        owner_id=str(owner_id),
+                    )
```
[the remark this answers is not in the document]
Zingzy: Fixed in 0db2c7a: a failed announce_revoked now raises before the doc delete, mirroring the delete-noop path, so the account stays ERASING and the re-queued sweep retries the eviction. The doc survives precisely so there is something left to retry against.

config.py
Zingzy: Closed in an earlier round, location for the record: the boot guard lives in the model_validator at config.py:915 next to the CUSTOM_DOMAINS_MOCK_DCV precedent (field validators only carry the >= 0 shape check). ENV=production with grace < 1 refuses to boot; both directions are tested.

tests/unit/repositories/test_erasure_deletes.py
Zingzy: Addressed in f554de1: the new tests/db suite runs the cascade against a real MongoDB (throwaway database per test, real ensure_indexes, the worker's actual composition graph) and asserts outcomes, not filters — including the domain-cascade regression and a bystander account compared document-for-document. CI gained a mongo:8 service container; filter-shape tests remain as fast-path coverage.

repositories/url_repository.py
Zingzy: Fixed in e0ea26a: the erasure path passes retain_blocked=True through the domain cascade, which scrubs and retains BLOCKED docs on the fqdn before deleting the rest, and the retained count reconciles with what survives (pinned by the real-Mongo Test A in f554de1). The interactive bulk domain-deletes keep their shipped hard-delete behavior; flagged separately as a follow-up since single-link delete refuses BLOCKED.

services/oauth_service.py
Zingzy: Fixed in 9c59c6d: the LINK branch gates on PENDING_DELETION and ERASING right after the linked-to account loads, before any provider write or token mint, consistent with the other three paths.

infrastructure/storage/r2.py
Zingzy: Fixed in 60aa116: the client keeps the strict refusal, and the wiring (app and worker factory both, via a shared builder) catches it, logs r2_storage_disabled_insecure_endpoint, and wires storage as None — the same degraded state as unconfigured, so a compose self-host boots with R2 off instead of crashing.

scripts/backfill_storage_prefix.py
```diff
@@ -0,0 +1,118 @@
...
+        batch.append(
+            UpdateOne(
+                # Re-guarded per doc: a concurrent first upload pinning the
+                # prefix mid-run must win (same guard as the app's repo).
+                {"_id": user_id, "storage_prefix": None},
+                {"$set": {"storage_prefix": _owner_key_prefix(user_id, secret_key)}},
+            )
+        )
+        if len(batch) >= _BATCH_SIZE:
+            stamped += coll.bulk_write(batch).modified_count
```
Zingzy: **Nit: unguarded `bulk_write`, though rerun is safe here.**

A transient write error (primary stepdown, network blip) aborts the run with a traceback.

This is much less serious than the same gap in `backfill_url_dest.py` was, for two reasons. There was a known, guaranteed-recurring failure there (16MB v1 documents), so the script died at the same document on every rerun. Here the write is a 16-character string onto a user document, so there is no standing failure mode, and the `storage_prefix: None` filter means a rerun skips everything already stamped and resumes correctly.

So the cost is an operator seeing a traceback and not knowing how far it got, rather than a migration that can never finish. One `try`/`except BulkWriteError` that logs the failed ids and continues would close it, matching what `backfill_url_dest.py` now does.

The two things around it are right and worth keeping: the per-document re-guard on `storage_prefix: None` so a concurrent first upload wins, and the rotation caveat at the top of the docstring, which is the constraint that decides whether running this is useful at all.

Zingzy (commented):
**Everything is closed. I verified each fix at the executing code rather than the commit message, and ran the new database suite myself. One nit left, in the backfill script.**

The real-Mongo suite is the part I want to lead with, because it is the thing I asked for three times and it arrived better than I described. `test_erase_leaves_bystander_documents_byte_identical` sets up two accounts with links, clicks, layouts and tokens, snapshots B, erases A, and asserts `after == before` on the whole snapshot rather than field by field. It also gives B a claimed link carrying a pre-claim click still stamped with the anonymous sentinel, which is exactly the shape that a careless owner-scoped delete would sweep up. `test_erase_retains_blocked_links_through_domain_cascade` pins the bug I found last round.

I ran them against a real Mongo: 5 passed. And I checked the availability policy holds, because a test suite that skips silently is worth very little. With `CI=1` and an unreachable server it calls `pytest.fail` rather than skipping, so the coverage cannot quietly disappear when a service container goes missing. Full suite with the database tests included: 3624 passed.

The rest, each checked at the choke point:

The blocked-link bypass is fixed with an opt-in `retain_blocked` flag defaulting to False, so the interactive domain cascade still deletes everything a user asks it to while the erasure path opts in. Scoping it that way rather than changing the shared filter is the right call, since the two callers genuinely want different things.

The OAuth link branch now calls `_ensure_not_pending_deletion` before it proceeds, closing the fourth token-minting path.

The R2 endpoint guard took the option I would have picked: the constructor stays strict, and the composition root catches the `ValueError` and degrades to the same disabled state as unconfigured R2, logging `r2_storage_disabled_insecure_endpoint` with the endpoint and reason. A self-host compose pointing at `http://minio:9000` boots.

The restore token is now validated, then flipped, then consumed only on success, so a transient error on the flip no longer burns the token. The one-shot property still holds because a replay fails at the guarded flip. That was the case that was terminal for OAuth-only accounts.

`verification-tokens` gets its `email` index, so the last `$or` branch in the cascade is covered.

## One nit, in the backfill

`bulk_write` is unguarded, so a transient write error aborts the run. It matters much less here than it did in `backfill_url_dest.py`, because the filter is `storage_prefix: None` and a rerun therefore resumes correctly rather than dying at the same document forever. Still one `try`/`except BulkWriteError` away from being restartable without an operator wondering how far it got.

Two things about that script I want to note as correct rather than flag. The rotation caveat is stated at the top of the docstring, which is where an operator will actually read it, and it is the constraint that decides whether running the script is useful at all. And the per-document re-guard on `storage_prefix: None` means a concurrent first upload wins the race instead of being clobbered, which is the same guard the app's own repository uses.

The HMAC derivation is duplicated from `image_ingest.owner_key_prefix`, which is the pattern I flagged on the earlier backfill. Here I think it is justified: the script is deliberately standalone with no project context, and the duplicated logic is two lines of HMAC rather than a URL parser with its own edge cases. Worth knowing it exists if that derivation ever changes.

Nothing blocking. This is the most thorough response to a review I have seen in this repo, and the two tests in `tests/db/` are the reason I would now believe the cascade does what it says.

scripts/backfill_storage_prefix.py
Zingzy: Fixed in ec01982: each batch flush is guarded, a transient failure logs and continues, and the script exits nonzero with a rerun hint. The None-filter makes the rerun resume exactly where it left off, as you noted.

PR #308, opened 2026-08-19
fix(schemas): advertise epoch seconds in expire_after request schema

Zingzy:
The `expire_after` validator on `CreateUrlRequest`, `UpdateUrlRequest` and `BulkUpdateExpiryRequest` accepts Unix epoch seconds as well as ISO 8601 strings, and the field descriptions already say so. The generated OpenAPI schema only advertised `string date-time | null`, so SDKs that generate their request types from the spec had to widen the type by hand.

This sets `json_schema_input_type` on the three before-validators so the schema becomes `anyOf [string date-time, integer, null]`, and regenerates `openapi.json` via `make openapi`. Runtime parsing and validation are unchanged, the diff to `openapi.json` is exactly the three added `integer` variants.

The TypeScript, Python and Rust SDK repos vendor this spec and will refresh their copies once this merges.

## Summary by CodeRabbit

* **Enhancements**
  * URL expiration values can now be provided as ISO 8601 date-time strings, Unix timestamps, or `null` when creating, updating, or bulk-updating URLs.
  * API documentation now accurately reflects the supported expiration formats.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -9593,6 +9593,9 @@
                 "format": "date-time"
               },
               {
+                "type": "integer"
+              },
+              {
                 "type": "null"
               }
             ],
@@ -10142,6 +10145,9 @@
               {
                 "type": "string",
                 "format": "date-time"
+              },
+              {
+                "type": "integer"
               },
               {
                 "type": "null"
@@ -12825,6 +12831,9 @@
               {
                 "type": "string",
                 "format": "date-time"
+              },
+              {
+                "type": "integer"
               },
               {
                 "type": "null"
--- a/schemas/dto/requests/bulk.py
+++ b/schemas/dto/requests/bulk.py
@@ -90,7 +90,9 @@ class BulkUpdateExpiryRequest(BulkIdsRequest):
         examples=[1767225600],
     )
 
-    @field_validator("expire_after", mode="before")
+    @field_validator(
+        "expire_after", mode="before", json_schema_input_type=datetime | int | None
+    )
     @classmethod
     def _parse_expire_after(cls, v: str | int | None) -> datetime | None:
         # Same coercion as UpdateUrlRequest.expire_after — one parser for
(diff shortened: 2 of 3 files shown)


Zingzy (commented):
Correct, minimal, and it closes the audit item that prompted it. Merge it.

Verified rather than taken from the description:

- `json_schema_input_type` on the before-validator is the right Pydantic v2 hook for this — it describes the *input* side while the field annotation keeps describing the parsed type, so runtime coercion is genuinely untouched.
- All three `expire_after` before-validators are covered, and there are no others: `git grep` across `schemas/dto/requests/` finds exactly the three this PR touches.
- `datetime` is a real runtime import in both modules (`url.py:9`, `bulk.py:16`). Worth stating explicitly because it's the one way this could have gone wrong: `from __future__ import annotations` makes the *annotations* lazy, but `json_schema_input_type=datetime | int | None` is a keyword-argument expression evaluated eagerly at class-definition time. A `TYPE_CHECKING`-only import would have been an import-time `NameError` rather than a subtle schema bug.
- The regenerated schema is right on all three: `[{"type":"string","format":"date-time"}, {"type":"integer"}, {"type":"null"}]`, and the `openapi.json` diff really is exactly three added `{"type": "integer"}` variants and nothing else.

Blast radius is nil — schema-only, no request path changes behaviour.

## The tail worth doing in the same sitting

**All three SDKs now vendor this file and diff it against `main`.** TS and Go had spec-drift jobs already, and Python gained one in its doctrine backport (`ci.yml:56`). So merging this turns three CI jobs red at once — including on unrelated PRs in those repos — until the vendored copies are refreshed. Not a reason to hold the merge, just a reason not to leave a gap between it and the three refreshes.

**Two hand-widenings become dead weight and should go with the refresh:**

- TypeScript's `Omit<..., "expire_after">` override in `links.ts`
- Rust's manual widening, which carries a comment explaining that the committed spec is knowingly narrower than the validator

Both exist *only* because the schema was wrong. Left in place after this lands, they become unexplained overrides that the next person will be reluctant to touch — the comment in the Rust one will actively mislead, since it documents a discrepancy that no longer exists.

Good to see `tests.yaml:40-46` already regenerates and diffs `openapi.json` in CI, so this can't quietly drift back out of sync.

PR #309, opened 2026-08-19
feat(telemetry): register sdk-rust and sdk-go client slugs

Zingzy:
The Rust SDK ships with `X-Spoo-Client: sdk-rust/<version>` as its default tag. Adding `sdk-rust` to `FIRST_PARTY_CLIENTS` lets `created_via` persist it, so links created through the SDK attribute properly, the same way `sdk-ts` and `sdk-py` are registered.

`sdk-py` was already added in #307, so this only adds the Rust slug plus the matching test case.

## Summary by CodeRabbit

* **New Features**
  * Added recognition for the Rust SDK as a first-party client.

* **Bug Fixes**
  * Correctly identifies Rust SDK requests, including versioned client strings such as `sdk-rust/0.2.0`.

Amended: sdk-go was also missing from the allowlist. The Go SDK has shipped since v0.1.0 with a default sdk-go/<version> tag, so its traffic was logged but never attributed as first-party. Added with the same shape as the other SDK entries.

### DIFF
--- a/shared/client_tag.py
+++ b/shared/client_tag.py
@@ -29,6 +29,7 @@ FIRST_PARTY_CLIENTS = frozenset(
         "bot",
         "sdk-ts",
         "sdk-py",
+        "sdk-go",
         "sdk-rust",
     }
 )
--- a/shared/client_tag.py
+++ b/shared/client_tag.py
@@ -20,7 +20,17 @@ _CLIENT_TAG_RE = re.compile(r"^([a-z0-9_-]{1,32})(?:/([A-Za-z0-9._-]{1,16}))?$")
 # (created_via) accepts only members, so arbitrary header values can never
 # become permanent, unfilterable document history.
 FIRST_PARTY_CLIENTS = frozenset(
-    {"dashboard", "landing", "snap", "raycast", "cli", "bot", "sdk-ts", "sdk-py"}
+    {
+        "dashboard",
+        "landing",
+        "snap",
+        "raycast",
+        "cli",
+        "bot",
+        "sdk-ts",
+        "sdk-py",
+        "sdk-rust",
+    }
 )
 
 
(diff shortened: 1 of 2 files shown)


Zingzy (commented):
Correct and complete — merge it. Both slugs verified against what the SDKs actually send, rather than against the PR description:

- Go: `defaultClientTag()` returns `"sdk-go/" + Version`, falling back to bare `"sdk-go"` (`version.go:14-19`).
- Rust: `format!("sdk-rust/{}", env!("CARGO_PKG_VERSION"))` (`src/client.rs:269`).

Both parse to the registered slugs under `_CLIENT_TAG_RE`, and the test targets `first_party_client` — the function that actually gates persistence — not just `parse_client_tag`. The set now covers every client that sends the header; spooWASDK predates it entirely.

## The amendment is the more interesting half

`sdk-go` shipping since v0.1.0 without ever being attributed is worth a moment, because the reason it went unnoticed is structural rather than an oversight.

`parse_client_tag` is deliberately permissive — any shape-valid slug is logged, so "a new client shows up in queries without a deploy". `first_party_client` is an allowlist, so only members reach `created_via`. That split is the right design, and the module comment says so. But the consequence is that a genuinely first-party client can look completely healthy in Axiom while its `created_via` is silently dropped on every write, and **nothing anywhere reports the disagreement**.

That's the shape that will repeat on SDK five. A cheap guard: when a slug parses successfully, matches a first-party shape (`sdk-*` is the obvious one), and isn't in `FIRST_PARTY_CLIENTS`, log a warning once per process. It turns a silent drop into a line in Axiom, which is all it needed to be caught the first time.

## Not worth doing

**Backfill.** spoo-go v0.1.0 was tagged 2026-08-19, so the unattributed window is about a day with essentially no external users. Recording it here so nobody spends an afternoon on it later.

PR #318, opened 2026-08-27
fix(safety): close the stack review findings

Zingzy:
## Addendum: detection changes

- **The per-IP creation counters are removed.** The rate limiter already bounds per-IP volume, a 92-day replay showed every real campaign concentrating on destination domains, and an IP is a poor creator identity with no reviewable object behind its alert.
- **Hot-link screening**: the hotness detector (already running in the click worker) gains an abuse consumer — a hot link whose destination host has no verdict yet emits a screening event (trigger `hot`). Activation-time coverage for pre-seeded campaigns.
- **Redirect terminal resolution**: a new seeded resolve-class feed (`redirectors`: platform share wrappers like t.co and lnkd.in that the gate deliberately allows) marks creates for chain-walking. The terminal URL is resolved with the shared SSRF guard, the terminal host judged through the normal screening path, and links wrapping that exact redirector URL are killed when the terminal verdict covers them. The wrapper host itself is never judged and never gets a verdict. Both triggers ride the shared deep-tier budget and stay silent on unresolved screenings.


services/safety/analyzer.py
```diff
@@ -164,9 +173,7 @@ def matcher(u: str, _p: str = verdict.path_pattern) -> bool:
 
             scope_note = f"scoped to pattern {path_pattern}"
         else:
-            # URL-scoped and host-scoped signals both enforce on the judged
-            # URL only: a host-wide claim (a feed listing) is exactly what
-            # the deep tier exists to confirm before anyone acts on it.
+            # Host-scoped claims wait for the deep tier; enforce the judged URL only.
             scope, path_pattern = "links", None
```
Zingzy: **The scope fix over-corrected here: feed hits now enforce per-URL.**

Only `BlockedPatternProvider` sets `scope="path_pattern"`. `FeedDomainProvider` and `WebRiskProvider` set neither scope nor pattern, and although `ProviderVerdict.scope` defaults to `"host"`, this branch never reads it. So every feed and Web Risk hit lands here and becomes `scope="links"` with `matcher = (u == event.url)`.

Consequences: a fishfish-listed domain blocks exactly one link per event, `FeedDeltaSweeper` emits one event per host carrying one `sample_url` so the rest are only reached one at a time, and `ToxicVerdictProvider` then refuses only that exact string at create, so `?a=1` gets through.

My original complaint was the opposite failure, a narrow judgment enforced host-wide. A feed listing an entire domain is host-level evidence and should keep host scope; the unanchored-regex case is the one that needed narrowing. Reading `verdict.scope` here rather than treating everything non-pattern as `links` would restore that distinction.

services/safety/investigation.py
```diff
@@ -362,10 +355,7 @@ async def _enact(
             )
         elif decision.action == "block_aliases":
             if verdict.scope == Scope.PATH_PATTERN and verdict.path_pattern:
-                # Pattern scope kills every matching link (all three
-                # collections), not just the reported ones; the pattern
-                # itself still waits for a human before it can refuse
-                # future creates via the blocklist.
+                # The pattern still waits for a human before it can refuse creates.
                 pattern = verdict.path_pattern
```
Zingzy, line 373: **Blocker: the model's regex is compiled and executed with no validation.**

I need to correct my own earlier review. I praised this PR specifically because `path_pattern` was "never compiled or executed, only proposed to a human, which defuses ReDoS and over-broad-regex enforcement." The scope fix removed that property, and nothing replaced it.

`InvestigationVerdict.path_pattern` is a plain `str | None` with no validator, and `matching_blocked_pattern` catches only `TimeoutError`. Tested against the tip:

```
OK    '^https://evil\.tld/kit/.*'        -> matched
RAISE '^https://evil.tld/(kit/.*'         -> error: missing )
RAISE '^https://evil\.tld/kit/[a-z.*'    -> error: unterminated character set
RAISE '*bad'                              -> error: nothing to repeat
```

`upsert_verdict` runs before `_enact`, so the bad pattern is durably stored first. The `regex.error` then escapes `block_matching`, `investigate` and `DeepAnalysisConsumer.consume`, which has no handler, so the message goes pending and the claimer re-runs a full investigation up to `deep_max_deliveries`. Every later event on that host re-reads the stored pattern and hits the same error.

Two small fixes: a field validator that `regex.compile`s the pattern (rejecting the verdict, which already degrades to review), and catching `regex.error` next to `TimeoutError` in `matching_blocked_pattern`. Worth also passing a tighter timeout here, since the matcher runs in Python over every link on the host at 0.2s per call and fails open.

The comment above the field still says an operator reviews it before it reaches the blocklist. That is no longer true.

services/safety/investigation.py
```diff
@@ -329,20 +328,14 @@ async def investigate(self, event: SafetyAnalyzeEvent) -> None:
...
+    def _corroborated(self, event: SafetyAnalyzeEvent) -> bool:
+        """An INDEPENDENT hard signal agreed with a toxic call — computed only
+        from facts code observed (the screening context or the feed_lookup
+        contextvar), never from the model's prose or the trigger alone."""
         screening = ((event.context or {}).get("screening") or "").lower()
         if any(src in screening for src in ("feed", "web_risk", "blocked_pattern")):
             return True
-        joined = " ".join(verdict.evidence).lower()
-        return any(src in joined for src in ("feed:", "web_risk", "hard hit"))
+        return saw_hard_hit()
```
Zingzy, line 348: **The corroboration flag is lost whenever `feed_lookup` shares a turn with another tool.**

The redesign is right: computed from observed facts, no model prose, no report short-circuit, and confidence now gates. The delivery mechanism does not survive the task boundary.

`_hard_hit.set(True)` runs inside the tool coroutine; this reads it from the agent-run task. pydantic-ai 2.31.0 dispatches tool calls with `asyncio.create_task` (`_tool_execution.py:808` and `:1180`), and a `ContextVar` set inside a task never propagates back to its parent. Verified:

```
inline await        -> parent sees: True
gather(coroutines)  -> parent sees: False
gather(create_task) -> parent sees: False
```

So a real feed or Web Risk hit silently fails to corroborate on any multi-tool turn, and under the default `corroborated` policy that sends a confirmed hit to review instead of blocking. Fail-safe direction, but the mechanism is defeated in the common case.

`test_model_prose_never_corroborates` and `test_burst_scam_self_corroborated_by_feed_hit_blocks` both set the variable inside a mocked `runner.run`, i.e. the same task, so they pass regardless.

Carrying the flag on the deps object the tools already receive, or reading the hit back out of the tool result, both survive `create_task`.

services/safety/scoring.py
```diff
@@ -49,74 +43,66 @@ def __init__(
...
-            for window in (self._burst_window, _DAILY_WINDOW):
-                bucket = now // window
-                keys.append(("ip", window, f"l1:ip:{window}:{ip_hash}:{bucket}"))
+            keys.append((window, f"l1:dom:{window}:{registrable_domain}:{bucket}"))
         pipe = self._redis.pipeline(transaction=False)
-        for _, window, key in keys:
+        for _, key in keys:
             pipe.incr(key)
             # 2x window: inspectable after the window closes, self-decaying.
             pipe.expire(key, window * 2)
```
Zingzy, line 103: **Leaked loop variable: every counter gets the daily TTL.**

The first loop binds `window` over `(self._burst_window, _DAILY_WINDOW)` and leaves it bound to `_DAILY_WINDOW`. This loop unpacks as `for _, key in keys`, discarding the per-key window, but this line still reads `window`. So the 600-second burst key is expired at `86400 * 2`, not `600 * 2`.

Retention only, since `_mark_fired` computes its own bucket from the right window, so the fired-marker and the counts stay correct. But the comment directly above says "self-decaying" and burst keys now outlive their window by 144x on a `noeviction` Redis.

The pre-refactor loop was `for _, window, key in keys:` and was correct. `for window, key in keys:` restores it.

Zingzy (commented):
**Final pass over the whole stack at `3dd506d7`. Most of what I raised is genuinely fixed, several at exactly the right choke point. Three things are not, and two of them are new, introduced by the fixes themselves. Suite is green (3364 passed), ruff format and lint clean.**

First, the structural point, because it changes how this should merge. The fixes live here and in #316, at the top of a nine-deep stack. PRs 297 through 305 still carry every original defect at their own heads. So merging bottom-up puts known-defective code on main at each step until this lands. Deploys fire on release publish rather than merge, so that is survivable, but it means **do not cut a release between merging #297 and merging this**, or squash the stack.

Also worth flagging: this PR has no test matrix and no `check_formatting` in CI, because neither runs on a non-`main` base. The largest PR in the stack (2256 additions) has had no automated test run at all. I ran the suite and the formatter locally instead.

---

## Fixed, verified by reading the executing code

The scheduler boot death is properly closed. `sync_registry` now retries inside the loop, and I ran it: two induced Mongo failures at boot, the runner logs `task_registry_sync_failed` each time, recovers on the third attempt, reaches `claim_due`, and cancels cleanly. The unknown-task occurrence-swallow is fixed at the root by name-scoped claims, and `finish_run` is now fenced on `claim_token` so a superseded runner cannot clobber state.

`dest.host` is indexed on all three collections. That was the widest-reaching one-line fix in the stack.

Corroboration is rebuilt correctly in design: `_corroborated` no longer takes the model's verdict at all, the `trigger == "report"` short-circuit is gone, confidence now gates via `corroborated and high`, and the flag is set only inside `if hits:` in `feed_lookup`. See the caveat below, though.

Scope binding reaches all three readers: the create gate through `verdict_covers`, the analyzer short-circuit through `_reenforce`, and the investigation stores the enacted scope rather than the claimed one, downgrading unapproved TOXIC to UNCERTAIN.

SSRF is closed properly by reusing `safe_fetch` instead of recopying it: RDAP goes through `fetch_public`, TLS connects to a vetted IP with SNI carrying the name, `resolve_chain` pins the resolved IP so there is no re-resolve window, and the error strings no longer leak exception class names. The gzip-bomb GET fallback is now streamed and closed before any body bytes are read.

The report lane has its own daily budget with a global fixed-window key, so wildcard DNS buys hosts but not budget, and it fails closed on Redis. The Web Risk key moved to `X-Goog-Api-Key` and `httpx` is silenced. The freshness short-circuit is now trigger-aware, so a report outranks a sweep verdict. Per-IP L1 counters were deleted outright, which kills the legacy proxy-IP bug, the junk embeds and the hashing concern in one subtraction, and the threshold firing is now retry-safe via `>=` plus a `SET NX` marker. The inline sink runs in the background. `link.blocked` re-enforcement now notifies. The PSL fetch is genuinely offline via `suffix_list_urls=()`, IDNA is UTS-46, and the backfill has a `BulkWriteError` guard and an `_id` cursor.

That is a lot of correct work, and several of these are better than what I suggested.

---

## Still open: the blocker from #304 is unfixed, and larger than I described

`ResolvedPublicLink.effective_status()` still never reads the `blocked` flag, and the docstring still asserts v1/emoji "can never be inactive or blocked." I re-ran it at this tip:

```
doc blocked flag      : True
effective_status()    : active
preview reveals dest  : True -> https://credential-harvest.example/login
stats long_url served : https://credential-harvest.example/login
```

`raw_v1` is the full unprojected document, so `blocked` is in memory and simply not consulted. I said two surfaces before; the real list is seven, and four of them do not even go through `effective_status`:

1. `GET /api/v1/public/preview/{code}`, v1 branch
2. Public stats page and API (`long_url`)
3. Legacy HTML preview `GET /{code}+`, which deliberately bypasses `resolve()` status checks so blocked links can still render
4. Legacy stats `GET/POST /stats/{code}`
5. Legacy export `GET/POST /export/{code}/{fmt}` in all four formats
6. The `$project` allowlist in `routes/legacy/helpers.py` projects `url` but not `blocked`, so the legacy handlers could not gate on it even if they tried
7. The v2 preview branch is correct, for contrast

Fixing `effective_status` covers 1 and 2. Items 3 through 6 need the projection widened and their own checks. Until then a takedown returns 451 on the redirect while six other endpoints hand out the destination and report the link as active.

---

## New, introduced by the fix stack

**The scope fix over-corrected, and screening enforcement is now near-inert.** In `_handle_toxic`, any provider verdict that is not `path_pattern`-scoped falls to the else branch and becomes `scope="links"` with `matcher = (u == event.url)`. Only `BlockedPatternProvider` sets a pattern; `FeedDomainProvider` and `WebRiskProvider` set neither, and `ProviderVerdict.scope` defaults to `"host"` but is never consulted here. So a fishfish hit on a domain now blocks exactly one link, the one URL in the event, and the create gate refuses only that exact string. Adding `?a=1` creates fine. My original complaint was that a narrow judgment was enforced host-wide; the fix inverted it so a genuinely host-wide judgment is enforced per-URL. A feed listing a domain is host-level evidence and should keep host scope; the pattern case is the one that needs narrowing.

**Model-authored regex is now compiled and executed with no validation.** This one I want to own directly: I previously praised this PR because `path_pattern` was "never compiled or executed, only proposed to a human." The scope fix removed that property. `verdict.path_pattern` is a plain `str | None` with no validator, and it now flows into `matching_blocked_pattern`, which catches only `TimeoutError`. I tested patterns a model plausibly emits:

```
OK    '^https://evil\.tld/kit/.*'        -> matched
RAISE '^https://evil.tld/(kit/.*'         -> error: missing )
RAISE '^https://evil\.tld/kit/[a-z.*'    -> error: unterminated character set
RAISE '*bad'                              -> error: nothing to repeat
```

The verdict is stored before `_enact` runs, so the bad pattern persists, and the consumer has no handler, so the message goes pending and re-runs a full investigation up to three times. There is no `regex.error` handling anywhere in the safety layer. Two small fixes: a pydantic validator that `regex.compile`s `path_pattern`, and catching `regex.error` alongside `TimeoutError`. The comment on the field still claims an operator reviews it before it reaches the blocklist, which is no longer true.

**The corroboration flag does not survive parallel tool calls.** `_hard_hit.set(True)` happens inside the tool coroutine; `saw_hard_hit()` reads it from the agent-run task. pydantic-ai 2.31.0 dispatches tool calls with `asyncio.create_task` (`_tool_execution.py:808` and `:1180`), and a context variable set inside a task is invisible to its parent. I confirmed both halves:

```
inline await        -> parent sees: True
gather(coroutines)  -> parent sees: False
gather(create_task) -> parent sees: False
```

So whenever the model calls `feed_lookup` in the same turn as any other tool, a real feed or Web Risk hit fails to corroborate. The direction is fail-safe, it under-blocks rather than over-blocks, but the mechanism only works in the single-tool-call case and the existing test sets the variable in the same task, so it passes either way. Returning the hit from the tool call into the run result, or having the deps object carry a mutable flag, both survive the task boundary.

**A 50,000-clause `$or`.** `block_matching` builds `pairs` from `list_by_dest_host_with_urls`, default limit 50,000, and passes it unchunked into both `list_active_owned_by_aliases` and `block_active_by_aliases`, each rendering `{"$or": [{"alias":..., "domain":...}, ...]}`. The legacy path next to it does the right thing with `{"_id": {"$in": [...]}}`. This is the shared-platform case, which is exactly what the scope work exists to serve.

**Every L1 counter now gets a 48-hour TTL.** In `_bump` the second loop unpacks as `for _, key in keys` but line 103 still reads `window`, which is left bound to `_DAILY_WINDOW` from the first loop. The 600-second burst key gets `expire(key, 172800)`. Retention only, since `_mark_fired` uses the right window, but the "self-decaying" comment is now false and burst keys live 144 times longer than intended on a `noeviction` Redis.

---

## Carried over, unfixed

`unblock_host` now exists and is correct, but still has zero callers outside tests, so an operator facing a false positive still cannot reverse one without touching Mongo, and a blocked v2 link still cannot be edited or deleted by its owner. On the legacy and emoji side `unblock` clears `blocked` on every blocked link at the host, including manual operator bans that predate safety, while the v2 path correctly guards on `blocked_reason` existing; the docstring claims the guard applies to all three.

`link.blocked` is still missing from the three renderer tables, so the most alarming event the product emits still renders as a raw JSON card, and `expand()` still auto-delivers it to every `link.*` subscriber. The edge `bulk_delete` return is still discarded while the enforcer reports success. `PATCH` retarget still bypasses L1 entirely. Geo destinations are now correctly gated at create and edit, which closes the bypass, but they are still not stamped into `dest`, so retroactive enforcement and the sweeps remain blind to them. `www.evil.com` and `evil.com` are still two verdicts. The Discord embed fields are still unclamped, so a long model-authored reason still drops the notification while enforcement proceeds. The `decided_by != "system"` allowlist still has no write path, and the deep tier bypasses the guard entirely.

Two smaller new ones worth a look: `HotLinkScreen`'s legacy fallback is not scoped by domain, so a hot alias on a custom domain that misses v2 can resolve against an unrelated v1 link and screen the wrong destination; and prompt caching is enabled unconditionally for all Anthropic tasks on conversations that contain fenced attacker-controlled page text, with a shared agent cache and no per-investigation isolation.

---

## What I would do before this merges

Fix the v1/emoji destination leak, validate `path_pattern`, and restore host scope for feed-sourced verdicts. Those three are small and they are the ones that decide whether enforcement is trustworthy and whether a takedown is actually a takedown. The contextvar and the `$or` are the next two. Everything else can ride into the hardening tail before the flag flips.

routes/legacy/url_shortener.py
```diff
@@ -488,6 +503,7 @@ async def preview_url(
                         "meta_tags": v2.meta_tags.model_dump()
                         if v2.meta_tags
                         else None,
+                        "blocked": v2.effective_status() is UrlStatus.BLOCKED,
```
Zingzy: **Blocker: this calls a property as a method, so the preview page 500s for every v2 link.**

`UrlV2Doc.effective_status` is a `@property` (`schemas/models/url.py:224`), while the new `LegacyUrlDoc.effective_status` (`:280`) is a plain method. The two legacy call sites above are correct; these two v2 ones are not:

```
v2.effective_status (property) -> UrlStatus.ACTIVE
v2.effective_status()          -> TypeError: 'UrlStatus' object is not callable
```

`GET /{short_code}+` is live and public, and both v2 paths reach this: the six-character branch when the v1 lookup misses (here), and the non-six-character branch which tries v2 first (line 521, same expression).

`v2.effective_status is UrlStatus.BLOCKED` on both lines fixes it.

Nothing caught this because the only test touching this route asserts registration rather than behaviour, and the new leak test covers the model and resolver layers. One request per branch against this handler would have.

routes/legacy/url_shortener.py
```diff
@@ -526,6 +543,19 @@ async def preview_url(
             status_code=404,
```
Zingzy, line 544: **Blocker: one of the four branches never sets `blocked`, so this gate cannot fire for it.**

The gate is right, but the v1 fallback further up (non-six-character code, v2 lookup missed, legacy hit) builds `url_data` with only `_id`, `url` and `password`. No `blocked` key, so `url_data.get("blocked")` is falsy and a blocked v1 link on a custom-length alias renders its destination here.

The emoji branch, the six-character v1 branch and the two v2 branches all set it; that one producer was missed.

Adding `"blocked": doc.effective_status() is UrlStatus.BLOCKED` to that branch closes it. Given two of the four producers were wrong in this commit (see the property bug), this is the spot that wants a route-level test per branch rather than another read-through.

services/safety/scoring.py
Zingzy, line 103: **Still open from the last round: every counter gets the daily TTL.**

The first loop leaves `window` bound to `_DAILY_WINDOW`. This loop unpacks as `for _, key in keys`, discarding the per-key window, but this line still reads `window`. Captured the actual expire calls at this tip:

```
key window=   600  ttl=172800  expected=  1200  <-- WRONG
key window= 86400  ttl=172800  expected=172800  OK
```

So the 600-second burst key lives 144x longer than designed. Counts and firing stay correct because `_mark_fired` computes its own bucket, so this is retention only, but the comment directly above says "self-decaying" and on a `noeviction` Redis these accumulate.

`for window, key in keys:` restores it.

Zingzy (commented):
**Everything I called a blocker last round is genuinely fixed, and I verified each one by running it rather than reading it. But the destination-leak fix introduced a live 500 on a public route and left one of its own four branches still leaking. Those two are merge blockers. Suite is 3374 green, ruff format and lint clean.**

## Fixed, verified by execution

The blocked-destination leak is closed across all seven surfaces. `effective_status()` now returns `blocked` for v1 and emoji, the shared `$project` selects the flag with an `$ifNull` default, and legacy stats, export and preview all return the same 451 the redirect serves. Putting `effective_status` on the model with a docstring that says out loud "so a new surface cannot forget the check the way six of them already did" is the right structural fix, not a patch of six call sites.

The model regex is validated. I re-ran my crash cases and nothing raises now: `is_valid_pattern` gates before use, `matching_blocked_pattern` catches `regex.error`, and an unusable pattern degrades to `Scope.LINKS` with a loud `safety_model_pattern_invalid`. That is the conservative direction, and it is pinned by a test.

The scope over-correction is fixed properly, at the level I hoped for. Providers now declare how far their signal reaches, feeds say `host`, Web Risk says `links`, the pattern provider says `path_pattern`, and `_handle_toxic` branches three ways on that. The docstring states the principle exactly: enforce as far as the signal reaches, never wider and never narrower. Three new tests pin all three cases.

Corroboration now survives the task boundary. Swapping the boolean contextvar for a mutable flag object is exactly right, because the child task mutates the same object. Verified against the real `reset_hard_hit`/`saw_hard_hit`:

```
inline await                       parent sees hard hit: True
gather(coroutines)                 parent sees hard hit: True
gather(create_task)                parent sees hard hit: True
no-hit control                     parent sees hard hit: False
```

You also closed the `?a=1` bypass I described, which I did not raise as a separate item. `_without_query` compares scheme, host and path, so a links-scoped verdict covers the query-string variant while a different path still misses.

Geo is handled by subtraction rather than by building a second index, which is the right call for a feature that never launched. And the tripwire test is the best thing in this delta: it fails the moment `GEO_RULES_ENABLED` is flipped without the enforcement work, and its failure message names the five-step work list. That is a constraint encoded where it cannot be forgotten.

The reversal loop is now complete end to end. The appeal link points at a real `/contact` route, tenant pages correctly keep their own copy instead of a relative link that would 404, and `spoo_unblock_links` in the ops tooling covers all three collections with `reason_stamped_only` defaulting on, so an appeal can never undo a manual operator ban.

## Blocker 1: the legacy preview page 500s for every v2 short code

`UrlV2Doc.effective_status` is a `@property`; the new `LegacyUrlDoc.effective_status` is a plain method. Lines 506 and 521 call the v2 one as a method:

```
v2.effective_status (property) -> UrlStatus.ACTIVE
v2.effective_status()          -> TypeError: 'UrlStatus' object is not callable
legacy.effective_status()      -> UrlStatus.ACTIVE
```

`GET /{short_code}+` is a live public route, and both v2 paths through it hit this: the six-character branch when the v1 lookup misses, and the non-six-character branch which tries v2 first. So the preview page is a 500 for every v2 link. Nothing caught it because the only test touching that route checks registration, not behaviour, and the new leak test exercises the model and resolver rather than the handler.

## Blocker 2: the fourth preview branch still leaks

Three of the four branches now set `blocked`. The v1 fallback, reached when the code is not six characters and the v2 lookup misses, builds `url_data` with only `_id`, `url` and `password`. So `url_data.get("blocked")` is falsy and a blocked v1 link on a custom-length alias still renders its destination on the preview page. The gate is right; one of the four producers does not feed it.

Both of these argue for one route-level test per branch, since the model-level tests cannot see either.

## Still open from earlier rounds

Two I raised last time are unchanged, and I confirmed the first by execution: every L1 counter still gets a 48-hour TTL because `_bump`'s second loop reads a `window` left bound to the daily value (the 600s burst key expires at 172800), and `block_matching` still hands an unchunked 50,000-clause `$or` into two queries while the legacy repo beside it uses `$in`.

Also carried: `link.blocked` is still missing from all three renderer tables while `expand()` auto-delivers it to `link.*` subscribers; the edge `bulk_delete` and `invalidate_many` results are still discarded while the enforcer reports counts as fact; the two 1,000-limit collectors still truncate with no warning unlike their three siblings; Discord embed fields for model-authored `reason` are still unclamped; `_aliases_to_block` is still report-only and `block_aliases` still touches only `urlsV2`, so a non-report trigger blocks nothing and still announces "destination auto-blocked"; admitted deep events still skip the review embed while the API arms the sink on `deep_enabled` alone and the worker also requires `llm.enabled`; PATCH retarget still bypasses L1; `www.` is still a free evasion; provenance `egress` is still hardcoded `None`; and sweeps are still v2-only.

Given the ops tooling now covers reversal, the in-backend `unblock_host` having no caller matters less, but note its legacy and emoji filters still clear manual bans that predate safety while the v2 filter correctly requires `blocked_reason`, and its docstring claims the guard covers all three.

## Smaller things in this delta

The appeal is attached only in `_error_page`, so the three new 451 pages (legacy stats, export, preview) render without it. A blocked link's owner landing on `/stats/<code>` gets no route back, which is the case the appeal exists for.

Worth settling before this ships: 451 is in `REDIRECT_EDGE_INTERCEPTED_STATUSES`, so once `EDGE_COMPOSED_ERRORS` flips the body is dropped and the frontend composes the page. The frontend `BlockedBody` carries an explicit comment that a CTA there was settled against, twice. An appeal is not the same thing as a report CTA, so the decision may still be right, but the two surfaces currently disagree and the backend copy silently stops rendering at flag-flip time.

The geo guard is write-only. It blocks new geo rules, but existing links carrying `geo_rules` still route through them, still promote `geo_redirect` entries to edge KV, and still enumerate publicly, so the hazard the tripwire describes persists for data already written. The tripwire cannot catch that either, since it returns early when the flag is off and asserts on a model field rather than a read path. Worth a one-line note in the tripwire text, or a count of affected links before enabling.

`/export/<code>?format=json` now emits a new public `blocked` key, because the JSON writer dumps the projected dict while csv, xlsx and xml use explicit key lists. Probably fine, but it is a public wire change.

On the comment hook: it reads `git diff --cached`, so under `pre-commit run --all-files` (the usual CI invocation) it passes everything unconditionally, and it is not wired into CI at all. Because a divider line resets the run counter, interleaving `# ─────` defeats it. It also flags multi-line blocks that already exist across roughly forty modules, and it is purely textual, so markdown headings inside a triple-quoted prompt string and stacked tool pragmas both trip it. As a local nudge it is fine; as an enforced rule it needs CI and an AST check.

Finally, CI on this PR still has no test matrix and no formatting job, because neither runs on a non-`main` base. I ran both locally.

## What I would fix before merge

The two blockers only. Everything else is the same hardening tail, and the flag gates still hold.

PR #324, opened 2026-08-29
refactor(safety): one Web Risk client behind both callers

Zingzy:
Two Google Web Risk integrations landed on 29 August a couple of hours apart, so neither review saw the other. `main` currently has:

| | safety analyzer | URL expander |
|---|---|---|
| settings | `SafetySettings` (`SAFETY_`) | `WebRiskSettings` (`WEB_RISK_`) |
| client | `WebRiskProvider` | `UrlExpandService._web_risk` |
| auth | `X-Goog-Api-Key` header | key in the query string |
| threat types | MALWARE, SOCIAL_ENGINEERING | plus UNWANTED_SOFTWARE |

Same API, same credential, two of everything. This consolidates them onto one client in `infrastructure/web_risk.py`, then deals with what sharing a credential implies.

## The consolidation

- **`infrastructure/web_risk.py`** is new. `lookup()` returns the matched threat types, an empty list for a clean URL, and `None` when the lookup could not answer. Callers must read `None` as absence, never as clean.
- **`WebRiskProvider`** becomes a thin adapter that maps a match to a `ProviderVerdict`.
- **`UrlExpandService`** takes a client instead of a raw key.
- **`WebRiskSettings` is deleted**, along with `WEB_RISK_TIMEOUT_SECONDS`, which nothing ever read.
- The expander reads `SAFETY_WEB_RISK_API_KEY`, which prod already has set, so the verdict works on deploy with no new variable.

## The key was going into the logs

The expander passed the key as a query parameter. `httpx` logs full request URLs at INFO, prod runs `LOG_LEVEL=INFO`, and unlike `urllib3`, `pymongo` and `botocore`, the `httpx` logger is not demoted in `infrastructure/logging.py`. `redact_sensitive_fields` matches event-dict keys, not substrings inside a message, so it would not have caught it. Every lookup would have written the key into the log store in plaintext.

Nothing has leaked, because `WEB_RISK_API_KEY` was never set in prod. Worth knowing that setting it would have started the leak rather than fixing the feature.

`test_the_key_never_rides_the_query_string` guards it, asserting against the whole recorded call rather than just `params`, so it survives a change in call shape.

## One credential means one quota

The consequence of consolidating: the expander is public and the analyzer is not, but they now spend the same project quota. The free tier is roughly 3,300 lookups a day. The expander allows 300/day per anonymous caller, so a handful of IPs using distinct URLs can exhaust it.

That is worse than a billing problem. When the quota is gone Google returns non-200, `lookup()` returns `None`, and the analyzer abstains, so the safety pipeline silently loses its only online reputation source and the trace looks identical to a transient error. A public endpoint should not be able to blind the abuse pipeline.

So the expander yields first. `WebRiskBudget` is a global daily counter in Redis, capped at `SAFETY_WEB_RISK_EXPANDER_DAILY_BUDGET` (default 1,000), leaving the analyzer roughly 2,300. Past the cap the expander stops asking and its verdict is absent, which is a state the tool already renders. Per-caller rate limits bound one IP; this bounds the bill and the analyzer's headroom. Without Redis the cap is unenforced, matching every other cache here.

Caching now distinguishes the two cases. An ask Google did not answer still is not cached, since that would hide the signal for a whole TTL. A lookup we declined to make is cached, because a spent budget is a sustained state, and without that the tool would re-ask and pay the 4s timeout on every request precisely when it is busiest.

## Behaviour kept deliberately

- Threat lists are explicit at both call sites now, as `ENFORCEMENT_THREAT_TYPES` and `DISPLAY_THREAT_TYPES`, with no default. Adding a type to the display list can no longer silently widen auto-blocking.
- The expander gets a 4s timeout against the analyzer's 10s, because a user is waiting on the response.
- A match Google declines to categorise still counts, as `["UNKNOWN"]`, matching the old provider's behaviour.
- `openapi.json` regenerates unchanged, so the wire contract is untouched.

## Verification

Full suite in a clean worktree at this head: **3423 passed, zero failures.**

An earlier revision of this description claimed 20 environment-dependent failures. That was wrong, and the cause is worth recording: `BaseSettings` reads `.env` directly, so a checkout that has one picks up real local configuration during tests. A clean worktree has no `.env` and the whole suite passes.

The header auth form was checked against the live API and reaches Google's IP check, which confirms the form is accepted. An end-to-end verdict could not be observed locally, because the key is IP-restricted and returns `API_KEY_IP_ADDRESS_BLOCKED` for this machine's current address. Both the header and query forms fail identically, which is what isolates the 403 to the restriction rather than to the change. A live verdict is worth confirming on prod after deploy.

### DIFF
--- a/dependencies/wiring.py
+++ b/dependencies/wiring.py
@@ -19,9 +19,11 @@ from infrastructure.cache.url_cache import UrlCache
 from infrastructure.captcha.hcaptcha import HCaptchaProvider
 from infrastructure.cloudflare_client import CloudflareClient
 from infrastructure.cloudflare_kv import CloudflareKVClient
+from infrastructure.http_client import HttpClient
 from infrastructure.logging import get_logger
 from infrastructure.ops_notify import DiscordOpsNotifier
 from infrastructure.storage.r2 import R2StorageClient
+from infrastructure.web_risk import DEFAULT_THREAT_TYPES, WebRiskClient
 from repositories.api_key_repository import ApiKeyRepository
 from repositories.app_grant_repository import AppGrantRepository
 from repositories.blocked_domain_repository import BlockedDomainRepository
@@ -117,6 +119,29 @@ from services.webhooks.consumers import WebhookFanoutClickSink
 from services.webhooks.renderers import default_renderers
 
 log = get_logger(__name__)
+
+# The expander answers a waiting request, so it can't sit on the safety
+# analyzer's queue-side timeout.
+_EXPANDER_WEB_RISK_TIMEOUT = 4.0
+
+
+def build_expander_web_risk(
+    settings: AppSettings, http_client: HttpClient
+) -> WebRiskClient | None:
+    """The URL expander's Web Risk client, on the same credential as the
+    safety analyzer. Its threat list is wider because this verdict is
+    displayed, never enforced.
+    """
+    safety = settings.safety
+    if not safety.web_risk_enabled:
+        return None
+    return WebRiskClient(
+        http_client,
+        api_key=safety.web_risk_api_key,
+        api_base=safety.web_risk_api_base,
+        threat_types=(*DEFAULT_THREAT_TYPES, "UNWANTED_SOFTWARE"),
+        timeout=_EXPANDER_WEB_RISK_TIMEOUT,
+    )
 
 
 def build_click_service(
@@ -349,9 +374,11 @@ def wire_services(app: FastAPI, settings: AppSettings, redis_client) -> None:
         # Online lookup: analyzer only, never the create path.
         analyzer_providers.append(
             WebRiskProvider(
-                http_client,
-                api_key=sf_settings.web_risk_api_key,
-                api_base=sf_settings.web_risk_api_base,
+                WebRiskClient(
+                    http_client,
+                    api_key=sf_settings.web_risk_api_key,
+                    api_base=sf_settings.web_risk_api_base,
+                )
             )
         )
         log.info("safety_web_risk_enabled")
@@ -540,8 +567,7 @@ def wire_services(app: FastAPI, settings: AppSettings, redis_client) -> None:
         MetaFetchCache(redis_client, prefix="url_expand"),
         regex_timeout=settings.blocked_url_regex_timeout,
         user_agent=settings.meta_tags.fetch_user_agent,
-        http_client=http_client,
-        web_risk_api_key=settings.web_risk.api_key,
+        web_risk=build_expander_web_risk(settings, http_client),
     )
     app.state.domain_intel_service = DomainIntelService(
         MetaFetchCache(redis_client, prefix="domain_intel", ttl_seconds=86_400),
--- a/dependencies/wiring.py
+++ b/dependencies/wiring.py
@@ -16,6 +16,7 @@ from infrastructure.cache.feature_flag_cache import FeatureFlagCache
 from infrastructure.cache.meta_fetch_cache import MetaFetchCache
 from infrastructure.cache.onboarding_cache import OnboardingCache
 from infrastructure.cache.url_cache import UrlCache
+from infrastructure.cache.web_risk_budget import WebRiskBudget
 from infrastructure.captcha.hcaptcha import HCaptchaProvider
 from infrastructure.cloudflare_client import CloudflareClient
 from infrastructure.cloudflare_kv import CloudflareKVClient
@@ -23,7 +24,11 @@ from infrastructure.http_client import HttpClient
 from infrastructure.logging import get_logger
 from infrastructure.ops_notify import DiscordOpsNotifier
 from infrastructure.storage.r2 import R2StorageClient
-from infrastructure.web_risk import DEFAULT_THREAT_TYPES, WebRiskClient
+from infrastructure.web_risk import (
+    DISPLAY_THREAT_TYPES,
+    ENFORCEMENT_THREAT_TYPES,
+    WebRiskClient,
+)
 from repositories.api_key_repository import ApiKeyRepository
 from repositories.app_grant_repository import AppGrantRepository
 from repositories.blocked_domain_repository import BlockedDomainRepository
@@ -139,7 +144,7 @@ def build_expander_web_risk(
         http_client,
         api_key=safety.web_risk_api_key,
         api_base=safety.web_risk_api_base,
-        threat_types=(*DEFAULT_THREAT_TYPES, "UNWANTED_SOFTWARE"),
+        threat_types=DISPLAY_THREAT_TYPES,
         timeout=_EXPANDER_WEB_RISK_TIMEOUT,
     )
 
@@ -378,6 +383,7 @@ def wire_services(app: FastAPI, settings: AppSettings, redis_client) -> None:
                     http_client,
                     api_key=sf_settings.web_risk_api_key,
                     api_base=sf_settings.web_risk_api_base,
+                    threat_types=ENFORCEMENT_THREAT_TYPES,
                 )
             )
         )
@@ -568,6 +574,9 @@ def wire_services(app: FastAPI, settings: AppSettings, redis_client) -> None:
         regex_timeout=settings.blocked_url_regex_timeout,
         user_agent=settings.meta_tags.fetch_user_agent,
         web_risk=build_expander_web_risk(settings, http_client),
+        web_risk_budget=WebRiskBudget(
+            redis_client, limit=settings.safety.web_risk_expander_daily_budget
+        ),
     )
     app.state.domain_intel_service = DomainIntelService(
         MetaFetchCache(redis_client, prefix="domain_intel", ttl_seconds=86_400),
--- a/infrastructure/cache/web_risk_budget.py
+++ b/infrastructure/cache/web_risk_budget.py
@@ -1,0 +1,56 @@
+"""Daily cap on the URL expander's share of the Web Risk quota.
+
+The expander and the safety analyzer run on one project's quota, but only
+the expander is a public endpoint. Without a cap, traffic on the tool can
+spend the free tier and leave abuse analysis with no online reputation
+source for the rest of the day, looking exactly like a transient error.
+So the expander yields first: past its cap it stops asking and its verdict
+is simply absent, which is a state the tool already renders.
+
+Global, not per caller. Per-caller rate limits bound one IP; this bounds
+the bill and the analyzer's headroom.
+
+Without Redis the cap is unenforced, matching every other cache here.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime, timezone
+
+import redis.asyncio as aioredis
+
+from infrastructure.logging import get_logger
+
+log = get_logger(__name__)
+
+_KEY_TTL_SECONDS = 172_800
+
+
+class WebRiskBudget:
+    def __init__(
+        self,
+        redis_client: aioredis.Redis | None,
+        *,
+        limit: int,
+        prefix: str = "web_risk_budget",
+    ) -> None:
+        self._redis = redis_client
+        self._limit = limit
+        self._prefix = prefix
+
+    async def take(self) -> bool:
+        """Claim one lookup. False once the day's cap is spent."""
+        if self._redis is None:
+            return True
+        key = f"{self._prefix}:{datetime.now(timezone.utc):%Y-%m-%d}"
+        try:
+            used = await self._redis.incr(key)
+            if used == 1:
+                await self._redis.expire(key, _KEY_TTL_SECONDS)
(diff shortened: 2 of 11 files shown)


Zingzy (commented):
**Good refactor, mergeable. The security fix is real and this is the better half of a problem I flagged twice on the safety stack. One consequence of sharing the credential is worth a decision before deploy, plus two nits and a correction to the PR body's test note.**

The consolidation is well done. `lookup()` returning matched threats, `[]` for clean, and `None` for "could not answer" is the right shape for a safety lookup, and both callers honour it: the expander uses an explicit `if threats is None`, and the provider's truthiness check is safe because a clean result and a failed one both abstain, which is what its docstring promises. The client logs `type(exc).__name__` rather than `str(exc)`, which matters here specifically, because an httpx error string embeds the request URL and that is exactly what used to carry the key. The deliberate divergences are preserved and explained rather than flattened: separate threat lists, separate timeouts, one credential.

On the key: this is the second of two call sites. I raised the query-string-plus-unsilenced-httpx problem on #300 and #305, and the analyzer half was fixed in the safety stack, which left this one behind. Your read that nothing leaked because `WEB_RISK_API_KEY` was never set is right, and the framing is the useful part: setting it would have started the leak rather than enabling the feature. The regression test is well chosen.

Deleting `WebRiskSettings` is clean. `AppSettings` carries `extra="ignore"`, so a stale `WEB_RISK_API_KEY` left in an environment is ignored rather than a boot failure, and the removed docstring's real safety property (no key means no fabricated verdict) survives in `build_expander_web_risk` returning `None`.

## Should-decide: one credential now means one quota, and the expander can starve the analyzer

`web_risk_enabled` is `bool(web_risk_api_key)`, and prod has `SAFETY_WEB_RISK_API_KEY` set, so on deploy the expander goes from inert to live. That is stated in the PR body as a convenience, and it is, but it is also the moment a second consumer starts drawing on the same quota.

The sizes are not comparable to what the analyzer was budgeted against. `WebRiskProvider`'s own docstring says the 100k/month free tier "covers report-triggered volume by orders of magnitude", and that reasoning was written when reports were the only source. `/api/v1/expand` is public with optional auth, at 15/min and 300/day anonymous, 60/min and 2000/day authenticated, per limiter key. The free tier is roughly 3,300 lookups a day. So a handful of callers working through their own daily budgets against distinct URLs can consume the platform's entire allowance.

What makes that more than a billing note is the failure mode. When the quota is out Google returns a non-200, `lookup()` returns `None`, and the analyzer's `WebRiskProvider` abstains. The safety pipeline silently loses its only online reputation source, and the only trace is a `web_risk_lookup_failed` warning that looks identical to a transient error. A public endpoint can now degrade the abuse pipeline, and nothing connects the two in the logs.

Second-order, and it comes from a decision that is otherwise correct: skipping the cache write when a configured lookup returns `None` is exactly right, because caching an unanswered verdict would hide the signal for the whole TTL. But it also means that while the quota is exhausted, every expand request re-asks Google and re-fails, so the cache stops absorbing load during precisely the window when it would help, and each request now carries the 4s timeout. The expander degrades hardest at the moment it is under most load.

None of this argues against the consolidation, which is the right call. It argues for one of: a separate key for the expander so the two budgets are independent, a cheap per-day counter that stops the expander calling Web Risk once it has spent its share, or an accepted decision that the expander may exhaust the tier with an alert on `web_risk_lookup_failed` so the analyzer's blind spot is at least visible.

## Nits

The enforcement-critical threat list is the implicit one. The analyzer constructs `WebRiskClient` without `threat_types` and inherits `DEFAULT_THREAT_TYPES`, while the expander passes its wider list explicitly. That is backwards for safety: the list that decides what gets auto-blocked is the invisible default, and the list that only renders in a response is the visible one. Someone later adding `UNWANTED_SOFTWARE` to the module constant, reasonably believing it is a display default, would silently widen enforcement. Passing the analyzer's list explicitly at its call site puts the enforcement decision where a reviewer will see it, and costs one line.

The regression test asserts the key is absent from `params`. Since the query string is the only place it ever went, that is the right guard, but asserting it appears nowhere in the outgoing call (URL included) would keep holding if the call shape ever changes.

## On the test note in the PR body

The 20 failures are not environment-dependent in a mysterious way, they are the known `.env` gotcha: settings read `env_file` directly, so a `.env` present in the repo root breaks that set of smoke and integration tests. Your main checkout has one; a clean detached worktree does not. I ran the full suite in a fresh worktree at this head and got **3416 passed, zero failures**, which is your 3396 plus exactly those 20. Worth correcting in the body so nobody spends time chasing them, and it also means this branch has no known-failing tests at all.

The rest of the verification section is the right shape, including isolating the 403 to the IP restriction by confirming both auth forms fail identically. Agreed that a live verdict should be confirmed on prod after deploy, and that is also the moment to watch quota consumption for the first day.

dependencies/wiring.py
```diff
@@ -349,9 +374,11 @@ def wire_services(app: FastAPI, settings: AppSettings, redis_client) -> None:
         # Online lookup: analyzer only, never the create path.
         analyzer_providers.append(
             WebRiskProvider(
-                http_client,
-                api_key=sf_settings.web_risk_api_key,
-                api_base=sf_settings.web_risk_api_base,
+                WebRiskClient(
```
Zingzy, line 382: **Nit, but it points the wrong way for safety: this is the list that causes blocking, and it is the implicit one.**

The analyzer builds the client without `threat_types`, so it inherits `DEFAULT_THREAT_TYPES` from the module. The expander, whose verdict is only displayed, passes its wider list explicitly.

That inverts the visibility you want. A future contributor adding `UNWANTED_SOFTWARE` to `DEFAULT_THREAT_TYPES`, reasonably reading it as a shared default, would silently widen what the analyzer auto-blocks. The PR body is explicit that this is an enforcement decision and deliberately out of scope, which is right, and passing the list here makes that decision visible at the place it takes effect:

```python
WebRiskClient(
    http_client,
    api_key=sf_settings.web_risk_api_key,
    api_base=sf_settings.web_risk_api_base,
    threat_types=DEFAULT_THREAT_TYPES,
)
```

Same behaviour, one line, and the next person editing the constant sees both call sites.

services/url_expand_service.py
```diff
@@ -75,30 +68,18 @@ async def expand(self, url: str) -> dict:
         # A configured-but-failing Web Risk call returns None, and caching
         # that would hide the safety signal for the whole TTL. Skip the
         # write so the next caller asks again.
-        if not (self._web_risk_api_key and web_risk is None):
+        if not (self._web_risk and web_risk is None):
```
Zingzy: **This decision is right, and it has a consequence worth knowing about under quota exhaustion.**

Not caching an unanswered verdict is correct: caching `None` would hide the safety signal for the full hour TTL, and the comment says so.

The interaction is with the shared credential. Once the Web Risk free tier is spent, every lookup returns `None`, so nothing here is ever cached, so every expand request re-asks Google, re-fails, and pays the 4s timeout first. The cache stops absorbing load during exactly the window when the endpoint is degraded, and the expander is slowest when it is busiest.

The analyzer feels this too: it shares the key, so an exhausted quota makes `WebRiskProvider` abstain, and the safety pipeline quietly loses its only online reputation source with nothing but a `web_risk_lookup_failed` warning that looks like any transient error.

Nothing to change in this line. It is worth deciding whether the expander gets its own key or a spend cap, so a public endpoint cannot consume the analyzer's budget. Detail in the review body.

infrastructure/cache/web_risk_budget.py
```diff
@@ -0,0 +1,56 @@
...
+        prefix: str = "web_risk_budget",
+    ) -> None:
+        self._redis = redis_client
+        self._limit = limit
+        self._prefix = prefix
+
+    async def take(self) -> bool:
+        """Claim one lookup. False once the day's cap is spent."""
+        if self._redis is None:
+            return True
```
Zingzy, line 44: **Should-fix: this counter is on the Redis tier whose compose block says "Lose-cache-on-restart is acceptable".**

`redis_client` here resolves to `REDIS_URI`, which in prod is `spoo_redis`: `--maxmemory 512mb`, `--maxmemory-policy allkeys-lru`, `--save ""`, `--appendonly no`.

So the day's spend is an ordinary LRU eviction candidate competing with URL cache entries, and it is lost outright on a container restart. Verified against a real Redis with this class:

```
first 7 takes (limit 5): [True, True, True, True, True, False, False]
--- key evicted / redis restarted ---
takes after eviction   : [True, True, True, True, True]   <- full budget again
```

No log line marks the reset, and the moment it is most likely to happen is a busy day, which is exactly when the analyzer needs its headroom.

The repo already answers this: `CreationPatternScorer` keeps its fixed-window counters on the durable queue Redis specifically because the cache Redis would evict them, and that instance runs `noeviction` with AOF. A budget is a counter rather than a cache, so the docstring's "matching every other cache here" is the line I would reconsider.

Related, the `except` arm below returns True as well. Failing open is right for a cache and questionable here, because what fails is the expander's displayed verdict while what is protected is the analyzer's quota. This module already argues the first is cheap, since absence "is a state the tool already renders" and the expander is meant to yield first. A losable store plus a fail-open reader stacks the two failure modes; pick either a durable store or a closed failure and the pairing stops mattering.

Zingzy (commented):
**All four points addressed, two of them better than what I suggested. The cap works, and I verified it. One thing about the new code is worth changing before this ships, and one is a judgement call.**

Confirming the fixes:

The threat lists are now `ENFORCEMENT_THREAT_TYPES` and `DISPLAY_THREAT_TYPES`, both passed explicitly at both call sites. That is better than the explicit-argument fix I proposed, because naming them by role means the next person editing one can see from the name which side they are on, rather than having to trace the caller.

The cache interaction is resolved precisely. `_web_risk_verdict` now returns whether Google was actually asked, and the write skips only on an answered-but-failed lookup while a declined one caches as a sustained state. That closes the amplification path exactly: once the budget is spent the endpoint stops re-asking and keeps serving from cache, instead of losing both.

The budget itself is right in scope and mechanics. Only the expander takes from it, the analyzer stays unbudgeted, and 1000 of a ~3.3k tier leaves real headroom. `INCR` is atomic, the TTL is set only on first increment, and the exhaustion warning fires exactly once at the crossing because no `INCR` return value is skipped. I ran it against a real Redis:

```
first 7 takes (limit 5): [True, True, True, True, True, False, False]
counter now            : 7
web_risk_budget_exhausted  limit=5     <- logged once
```

And the PR body correction is the right call, including recording why rather than just deleting the claim. Full suite in a clean worktree at this head: 3423 passed, which matches your number exactly.

## Should-fix: the counter lives on the tier designed to lose it

`WebRiskBudget` is wired with `redis_client`, which comes from `REDIS_URI`. In prod that is `spoo_redis`, and its compose block is explicit about what it is for: `--maxmemory 512mb`, `--maxmemory-policy allkeys-lru`, `--save ""`, `--appendonly no`, with the comment "Lose-cache-on-restart is acceptable."

So the day's spend can vanish two ways. Under memory pressure the counter is an ordinary eviction candidate competing with URL cache entries, and on any restart of that container it is gone outright. Continuing the same run:

```
--- key evicted / redis restarted ---
takes after eviction   : [True, True, True, True, True]   <- full budget again
```

No log, no signal, and the next 1000 lookups are granted. The failure is quiet in the direction that matters, because the whole point of the cap is protecting the analyzer's headroom, and the moment it silently resets is the moment traffic is high enough to be evicting things.

This codebase already settled this question the other way. `CreationPatternScorer` puts its fixed-window counters on the durable queue Redis precisely because the cache Redis would evict them, and that instance runs `--maxmemory-policy noeviction` with `--appendonly yes`. A budget is a counter, not a cache, and it belongs on the same tier as the other counters. The docstring's "matching every other cache here" is the reasoning I would revisit, since it classifies this as a cache when its own purpose is accounting.

## Should-decide: failing open contradicts what the cap is for

`take()` returns True when Redis is absent and when `INCR` raises, on the grounds that a broken counter must not take the feature down. That reasoning holds for a cache. It does not obviously hold here, because the thing being protected is not this feature.

If the counter is unavailable, the choice is between the expander losing its displayed verdict and the analyzer losing its online reputation source for the rest of the day. The module docstring already argues the first is cheap: absence "is a state the tool already renders", and the design principle it states is that the expander yields first. Failing open inverts that in the one case where you cannot measure the spend.

Failing closed also composes better with the storage question above: if the counter is on a durable tier, an error genuinely is exceptional rather than routine, so refusing is rare. If you keep it on the cache tier and fail open, then a Redis restart plus a busy day is a plausible route to spending the whole tier with nothing in the logs to explain it afterwards.

Either direction is defensible. What I would not leave is the current pairing of a losable store with a fail-open reader, since the two failure modes stack.

Neither of these blocks the merge. The consolidation and the key fix are the valuable parts and both are done.

PR #331, opened 2026-08-30
fix(safety): stop paying a model to agree with a feed

Zingzy:
## What happened

The first fishfish sync after the deep tier went live ran 53 investigations. 51 of them were feed escalations, and the outcome was:

```
scam_host  block_host     auto=True   corroborated=True   45
scam_host  review         auto=False  corroborated=True    5
scam_host  block_aliases  auto=True   corroborated=True    1
benign     benign         auto=True   corroborated=False   2
```

Zero changed an outcome. 45 called `block_host` on a host screening had already blocked, which is idempotent. The 5 medium-confidence ones were already blocked too. About $1.15 spent confirming decisions the feed had already made and enforced.

## Change

`_handle_toxic` escalated unconditionally. A feed naming a host is the host-wide answer, so there is no reach question to hand a model. Host-scope feed hits now skip the deep tier.

Narrower verdicts still escalate. A link-scoped or pattern-scoped finding genuinely raises the question of how far it reaches, and that path is unchanged.

## The one result worth keeping

`e.vg` came back `scam_host`, high confidence, scope `links`. That was correct: `e.vg` is a general-purpose shortener, and it sits on fishfish because scammers routed through it, not because the domain is the scam. It is in this repo's own `shorteners` feed as well.

It arrived after the host block, so it changed nothing, but it was the only reason anyone found out. Dropping the escalation drops that signal too.

So a feed block landing on a domain in the `shorteners` or `redirectors` feed now logs `safety_feed_block_on_shared_carrier` with the host and sample URL. Enforcement is unchanged and still follows the feed. The warning is how an operator learns the block reached wider than the evidence, at no cost per event.

## Notes

The module docstring claimed screening never originates a host-wide block. That has not been true since feed providers landed, since `FeedDomainProvider` returns host scope. Corrected to describe what the code does.

## Summary by CodeRabbit

* **New Features**
  * Improved safety analysis for host-wide verdicts from curated feeds.
  * Shared link carriers, including shorteners and redirect wrappers, now receive additional warnings when blocks may affect unrelated links.
  * Added resilient carrier detection that safely handles unavailable feed data.

* **Bug Fixes**
  * Prevented unnecessary deep investigations for trusted host-level feed verdicts.
  * Preserved escalation for non-feed sources and review flags when escalation is unavailable.
  * Sanitized warning details to exclude query strings and fragments.

### DIFF
--- a/services/safety/analyzer.py
+++ b/services/safety/analyzer.py
@@ -249,7 +249,8 @@ class SafetyAnalyzer:
                 host=event.host,
                 registrable_domain=event.registrable_domain,
                 source=source,
-                sample_url=event.url,
+                # Query strings on these carry recipient ids and tokens.
+                sample_path=event.url.split("?", 1)[0].split("#", 1)[0],
             )
 
     async def _reenforce(self, event: SafetyAnalyzeEvent, existing: VerdictDoc) -> bool:
--- a/services/safety/analyzer.py
+++ b/services/safety/analyzer.py
@@ -10,8 +10,10 @@ process hosts it (worker consumer or inline sink). Semantics:
 - First non-abstaining provider wins. No provider judging means tier
   UNCERTAIN and a human review embed — BENIGN is never inferred from
   absence of evidence.
-- Screening never originates a host-wide block: it blocks only the links
-  its evidence covers and escalates the host question to the deep tier.
+- Screening enforces exactly as far as its signal reaches. A curated feed
+  naming a host is a host-wide call and needs no second opinion; anything
+  narrower blocks the links it covers and escalates the host question to
+  the deep tier.
 - Everything is best-effort: analysis failures log and drop, they never
   propagate into the trigger path (a report must store even when analysis
   is broken).
@@ -33,6 +35,7 @@ from services.safety.events import SafetyAnalyzeEvent
 from services.safety.providers import (
     AnalysisProvider,
     ProviderVerdict,
+    SharedCarrierLookup,
     verdict_covers,
 )
 from shared.datetime_utils import as_aware_utc
@@ -59,6 +62,10 @@ _TRIGGER_AUTHORITY = {
 # Unresolved machine-volume triggers stay silent; review pings would drown the channel.
 _SILENT_TRIGGERS = frozenset({"sweep", "hot", "redirect"})
 
+# A curated feed naming a host IS the host-wide answer, so there is no reach
+# question left to bill a model for. FeedDomainProvider names itself this way.
+_FEED_SOURCE_PREFIX = "feed_"
+
 
 class SafetyAnalyzer:
     def __init__(
@@ -71,6 +78,7 @@ class SafetyAnalyzer:
         reverdict_ttl_hours: int = 24,
         admission: AdmissionPolicy | None = None,
         deep_sink: DeepAnalysisSink | None = None,
+        carriers: SharedCarrierLookup | None = None,
     ) -> None:
         self._providers = providers
         self._verdict_repo = verdict_repo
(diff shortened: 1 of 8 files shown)


services/safety/providers.py
```diff
@@ -46,6 +46,41 @@ async def analyze(
...
+    lookup only marks the blocks that therefore reached every unrelated
+    link routed through the same domain.
+    """
+
+    def __init__(self, repo: FeedDomainRepository, *, feeds: tuple[str, ...]) -> None:
+        self._repo = repo
+        self._feeds = feeds
+
+    async def covers(self, host: str, registrable_domain: str) -> bool:
+        """Never raises: an unreachable feed store must not turn a narrow
```
Zingzy: **Should-fix:** this states a constraint the guard does not protect. `covers()` cannot widen or narrow a block: by the time it runs, `block_host` and the verdict upsert already happened. What a raise would actually cost is the `notifier.safety_action` call right after it, so the operator would never hear about a block that just fired. The never-raise guard is load-bearing for that notification. Say that instead, otherwise the next editor reasons from the wrong invariant.

Zingzy (commented):
Nothing blocking. The skip lands at the right choke point: enforcement and the verdict upsert happen before it, `_enact` on a benign result never touches enforcement, and no unblock path exists, so escalating a feed host hit could only re-affirm the block. Ran the safety suite on the branch, 187 passed.

One should-fix (a docstring stating the wrong constraint) and four small nits, all inline.

services/safety/analyzer.py
```diff
@@ -225,6 +235,24 @@ async def _handle_toxic(
...
+        if self._carriers is None:
+            return
+        if await self._carriers.covers(event.host, event.registrable_domain):
+            log.warning(
+                "safety_feed_block_on_shared_carrier",
+                host=event.host,
+                registrable_domain=event.registrable_domain,
+                source=source,
+                # Query strings on these carry recipient ids and tokens.
+                sample_path=event.url.split("?", 1)[0].split("#", 1)[0],
```
Zingzy: **Nit:** this duplicates `_without_query` in `providers.py`, which does the same split plus `rstrip("/")`. This module already imports from there. Reusing it also fixes the name for free: `sample_path` holds scheme, host and path, not a path.

services/safety/feeds.py
```diff
@@ -58,6 +58,10 @@
 # The RESOLVE class: never refused, never a judgment source — membership only
 # marks a created link for terminal-URL resolution of the chain's endpoint.
 REDIRECTOR_FEED = "redirectors"
+
+# Domains that carry OTHER people's links: listed for what routed through
+# them, so a host-wide block reaches every unrelated link too.
+CARRIER_FEEDS = (SHORTENER_FEED, REDIRECTOR_FEED)
```
Zingzy: **Nit:** placement. This splits `REDIRECTOR_FEED` from its `_REDIRECTOR_SEED_PATH` two lines down. Move it below the redirector block so the constant and its seed path stay together.

services/safety/providers.py
```diff
@@ -46,6 +46,41 @@ async def analyze(
...
+        block into a host-wide one."""
+        candidates = [d for d in (host, registrable_domain) if d]
```
Zingzy: **Nit:** when host equals registrable domain, the common apex case (`e.vg` included), this runs the same point read twice per feed, and the miss test pins the 4 calls. `dict.fromkeys((host, registrable_domain))` dedupes while keeping order. Rare path, so cost is near zero, hence nit.

tests/unit/services/safety/test_analyzer.py
```diff
@@ -411,17 +411,102 @@ async def test_toxic_escalates_the_host_decision_to_investigation(self):
...
+
+        await analyzer.analyze(_event())
+
+        deep_sink.emit.assert_not_awaited()
+        admission.decide.assert_not_awaited()
+        reason = notifier.safety_action.await_args.kwargs["reason"]
+        assert "feed listing is itself the host-wide answer" in reason
+
+    @pytest.mark.asyncio
+    async def test_host_scoped_hit_from_a_non_feed_source_still_escalates(self):
```
Zingzy, line 440: **Nit:** the review fallback lost its positive test. The replaced test was the only one asserting `needs review` appears; what remains is the negative assertion at line 411. A non-feed host verdict with escalation denied, or the deep tier unwired, no longer has a test proving the operator still gets the review flag.

PR #335, opened 2026-09-02
fix(safety): make L2 decisive, honest about renders, and quiet on sweeps

Zingzy:
Everything here came from replaying the first days of review pings against the hosts that produced them.

## Renders

`networkidle0` waits for zero in-flight requests, so one analytics beacon kept a live page "loading" until it timed out. `nauratimm.net`, an artist's portfolio, was escalated as a possible scam that way; it renders 88 KB under `networkidle2`. The client now tries `networkidle2` then `domcontentloaded`.

The screenshot was decoded and discarded. It now rides along as image content via `ToolReturn`, and the untrusted-data fence names it, since a page can put instruction-shaped text in an image as easily as in markup.

Frameset and meta-refresh targets are surfaced (`edanmed.com` was 340 bytes of frameset reported as "no content"). A bot challenge page is named as missing evidence instead of left to inference.

## Noise

`_enact` pinged on every review verdict while `_record_and_review` skipped sweeps, so a successful investigation woke an operator and a crashed one stayed quiet. 30 of the first 35 pings were sweeps whose page had died. Coverage triggers now record and stay silent.

A `redirector_service` verdict with an empty proposal list pinged with nothing to approve. It is now a recorded gray observation.

## Enforcement

Under the `confident` policy a sweep-found scam at links scope called `block_aliases` with an empty list, because only a report carries `reported_codes`. It now falls back to blocking the judged URL. Without this, flipping `SAFETY_DEEP_AUTOBLOCK` would have enforced nothing for most L2 findings.

## Proposals

`waa.ai` came back `redirector_service / high` and went nowhere. The prompt only proposed cloaks with no product while the list holds `bit.ly` and `tinyurl.com`; proposals were rendered into Discord and never stored; nothing could apply one.

The prompt now requires a proposal for every `redirector_service`, into `shorteners` (refuses future links) or `redirectors` (resolve only) by blast radius. A high-confidence `redirectors` proposal applies itself. A `shorteners` proposal still waits for a human. Proposals persist on the verdict, and `FeedDomainRepository` gains `add()`. The ops-mcp side (`spoo_feed_add`) is in the ops-mcp repo.

## Not in this PR

Two env decisions: `SAFETY_DEEP_AUTOBLOCK=confident` and `SAFETY_SHORTENERS_ENABLED=true`.

## Summary by CodeRabbit

- **New Features**
  - Web page rendering retries alternate wait conditions to improve successful captures.
  - Page analysis can include screenshots and identifies embedded frames, refresh destinations, and common anti-bot challenge pages.
  - Redirector proposals can be applied automatically when policy allows; shortener proposals require review.
  - Proposed domain-list actions are saved for traceability.
  - Exact URLs can be blocked when alias codes are unavailable.

- **Bug Fixes**
  - Redirector findings without proposals are treated as benign gray results.
  - Silent sweep, hot-trigger, and redirect checks no longer send unnecessary review notifications.

### DIFF
--- a/infrastructure/browser_run.py
+++ b/infrastructure/browser_run.py
@@ -8,7 +8,9 @@ the render, because "the page looked clean" and "the page looked clean
 to a scanner IP" are different pieces of evidence.
 
 Failures return None rather than raising: a dead render is an absent
-piece of evidence, not a broken investigation.
+piece of evidence, not a broken investigation. But a render that failed
+because we asked for too strict a wait is NOT absent evidence, it is a
+wrong answer, so each wait condition is tried in turn before giving up.
 """
 
 from __future__ import annotations
@@ -22,6 +24,9 @@ from infrastructure.logging import get_logger
 log = get_logger(__name__)
 
 _API_BASE = "https://api.cloudflare.com/client/v4"
+# networkidle0 wanted ZERO in-flight requests, so one analytics beacon kept
+# a live page "loading" until it timed out. First success wins.
+_WAIT_LADDER = ("networkidle2", "domcontentloaded")
 EGRESS_LABEL = "cloudflare datacenter (Browser Run)"
 
 
@@ -31,6 +36,14 @@ class RenderResult:
     html: str
     screenshot: bytes  # webp
     egress: str = EGRESS_LABEL
+
+
+def _is_wait_timeout(exc: Exception) -> bool:
+    """Browser Run reports a goto timeout as error code 6002 in the body;
+    a dead host is 5006 and never benefits from a looser wait."""
+    response = getattr(exc, "response", None)
+    text = getattr(response, "text", "") or ""
+    return "6002" in text or "timeout was reached" in text.lower()
 
 
 class BrowserRunClient:
@@ -46,18 +59,33 @@ class BrowserRunClient:
         self._account_id = account_id
         self._token = api_token
         self._timeout = timeout_seconds
+        self._last_was_wait_timeout = False
 
     @property
     def configured(self) -> bool:
         return bool(self._account_id and self._token)
 
     async def snapshot(self, url: str) -> RenderResult | None:
-        """Render *url* and return HTML + screenshot, or None on any
-        failure (logged). The browser follows JS/meta redirects the plain
-        resolver cannot see — what it lands on is the truth of the page."""
+        """Render *url* and return HTML + screenshot, or None once every
+        wait condition has failed (each logged). The browser follows
+        JS/meta redirects the plain resolver cannot see — what it lands on
+        is the truth of the page."""
         if not self.configured:
             log.warning("browser_run_unconfigured")
             return None
+        for attempt, wait_until in enumerate(_WAIT_LADDER, start=1):
+            result = await self._attempt(url, wait_until)
+            if result is not None:
+                if attempt > 1:
+                    log.info("browser_run_recovered", url=url, wait_until=wait_until)
+                return result
+            if not self._last_was_wait_timeout:
+                # A dead host fails the same way under every condition; only
+                # a wait timeout is something the next rung can fix.
+                return None
+        return None
+
+    async def _attempt(self, url: str, wait_until: str) -> RenderResult | None:
         endpoint = f"{_API_BASE}/accounts/{self._account_id}/browser-rendering/snapshot"
         try:
             response = await self._http.post(
@@ -67,7 +95,7 @@ class BrowserRunClient:
                     "url": url,
                     "screenshotOptions": {"type": "webp"},
                     "gotoOptions": {
-                        "waitUntil": "networkidle0",
+                        "waitUntil": wait_until,
                         "timeout": int(self._timeout * 1000),
                     },
                 },
@@ -76,13 +104,17 @@ class BrowserRunClient:
             response.raise_for_status()
             body = response.json()
         except Exception as exc:
+            self._last_was_wait_timeout = _is_wait_timeout(exc)
             log.warning(
                 "browser_run_snapshot_failed",
                 url=url,
+                wait_until=wait_until,
                 error=str(exc),
                 error_type=type(exc).__name__,
+                wait_timeout=self._last_was_wait_timeout,
             )
             return None
+        self._last_was_wait_timeout = False
         result = body.get("result") or {}
         html = result.get("content") or ""
         shot_b64 = result.get("screenshot") or ""
@@ -91,6 +123,6 @@ class BrowserRunClient:
         except Exception:
             screenshot = b""
         if not html and not screenshot:
-            log.warning("browser_run_empty_result", url=url)
+            log.warning("browser_run_empty_result", url=url, wait_until=wait_until)
             return None
         return RenderResult(url=url, html=html, screenshot=screenshot)
--- a/repositories/feed_domain_repository.py
+++ b/repositories/feed_domain_repository.py
@@ -20,6 +20,7 @@ from datetime import datetime, timezone
 from pymongo import UpdateOne
 
 from repositories.base import BaseRepository
+from shared.url_utils import registrable_domain
 
 _BULK_BATCH = 5_000
 
@@ -73,6 +74,28 @@ class FeedDomainRepository(BaseRepository[None]):
             )
             purged = int(result.deleted_count)
         return kept, purged, new_domains
+
+    async def add(self, feed: str, domain: str) -> bool:
+        """Add one domain to *feed*. True when it was new. Seeds only run on
+        an empty feed and syncs only replace feeds that have an upstream, so
+        an add here survives both."""
+        # A model-emitted string becomes a Mongo key: reduce it to the
+        # registrable domain or refuse, so contains() can actually match it.
+        domain = registrable_domain(domain.strip().lower())
+        if not domain or "/" in domain or " " in domain or "." not in domain:
+            return False
+        result = await self._col.update_one(
+            {"_id": self._key(feed, domain)},
+            {
+                "$set": {
+                    "feed": feed,
+                    "domain": domain,
+                    "synced_at": datetime.now(timezone.utc),
+                }
+            },
+            upsert=True,
+        )
+        return result.upserted_id is not None
 
     async def contains(self, feed: str, domain: str) -> bool:
         doc = await self._col.find_one(
--- a/services/safety/investigation.py
+++ b/services/safety/investigation.py
@@ -22,16 +22,20 @@ from __future__ import annotations
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
+from typing import Literal
 
 from pydantic import BaseModel, Field
 
 from infrastructure.llm import LlmTask, LlmTaskFailed, LlmTaskRunner, load_prompt
 from infrastructure.logging import get_logger
+from repositories.feed_domain_repository import FeedDomainRepository
 from repositories.url_repository import UrlRepository
 from repositories.verdict_repository import VerdictRepository
 from schemas.enums.safety import VerdictTier
 from services.safety.enforcer import SafetyEnforcer
-from services.safety.events import SafetyAnalyzeEvent
+from services.safety.events import SILENT_TRIGGERS, SafetyAnalyzeEvent
+from services.safety.feeds import REDIRECTOR_FEED
+from services.safety.providers import without_query
 from services.safety.tools import reset_hard_hit, saw_hard_hit
 from shared.validators import is_valid_pattern, matching_blocked_pattern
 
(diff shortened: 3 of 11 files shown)


services/safety/tools.py
```diff
@@ -269,16 +288,33 @@ def trim_html(html: str) -> str:
...
+    if _is_bot_challenge(parser.title, visible):
+        parts.append(
+            "NOTE: this is an anti-bot challenge page, not the site. The "
+            "render was refused, so this is MISSING evidence about the "
+            "destination and says nothing about whether it is safe."
+        )
     return "\n".join(parts)
 
 
+def _is_bot_challenge(title: str, visible: str) -> bool:
```
Zingzy: **Should-fix:** this note is spoofable by the page it describes. The haystack is attacker-controlled (title + visible text), and on a hit the code appends a trusted-voice instruction saying the render was refused and the evidence is missing. A scam page that tucks "verifying you are human" into a corner of its visible text gets the model told to discount the forms and text shown right above the note. That converts a blockable page into uncertain.

Bounded (under-blocking is the accepted failure direction here, and reports still reach a human with the screenshot), but cheap to harden: only emit the note when the page also has no forms and thin visible text (the real kisalt.com case is 27KB of challenge and nothing else), and phrase it as "matches anti-bot challenge markers" rather than asserting the render was refused.

repositories/feed_domain_repository.py
```diff
@@ -74,6 +74,26 @@ async def replace_feed(
             purged = int(result.deleted_count)
         return kept, purged, new_domains
 
+    async def add(self, feed: str, domain: str) -> bool:
```
Zingzy, line 78: **Should-fix, two halves of one contract:**

1. `domain` here is a model-emitted string written straight into a Mongo key. A proposal of `waa.ai/path` stores an inert row `contains()` can never match, while `safety_list_applied` logs it as applied. Run it through `registrable_domain()` (or reject non-hostname shapes) before the write. Relatedly, `ListProposal.list` is a bare `str` whose comment still says `"shorteners" | "manual"`, contradicting the prompt this PR rewrote; `Literal["shorteners", "redirectors"]` makes pydantic reject garbage at parse, which degrades to the existing `LlmTaskFailed` path. The ops-mcp input already patterns on exactly this set.

2. This method is why codecov/patch is red: every caller mocks it, so its normalization and was-it-new semantics are untested. Same shape as the SharedCarrierLookup gap fixed in #331's third commit; a direct test class settles it. (The `feed_repo is None` break in `_enact` is uncovered too.)

The docstring preemptively answering the purge question (seeds only fire on empty, syncs only replace upstream-backed feeds) is exactly right, by the way. That was the one scary interaction.

services/safety/investigation.py
```diff
@@ -144,8 +154,16 @@ def decide_authority(
         return AuthorityDecision("benign", VerdictTier.BENIGN, auto=True)
 
     if cls == Classification.REDIRECTOR_SERVICE:
-        # Adding a whole service to a block list reaches every FUTURE link
-        # to it — always a human tap, regardless of confidence.
+        if not verdict.proposals:
+            # Naming a service is an observation; without a list proposal
+            # there is nothing for an operator to approve or refuse.
+            return AuthorityDecision("benign", VerdictTier.GRAY, auto=True)
+        if high and all(p.list == RESOLVE_ONLY_LIST for p in verdict.proposals):
```
Zingzy: **Should-decide:** this branch checks confidence but never the policy, so even under `SAFETY_DEEP_AUTOBLOCK=off` the model autonomously writes to the redirectors feed. Defensible (a resolve-only add is not a block, and the knob is named autoblock), but an operator who sets `off` plausibly means "the model takes no autonomous action at all." One `policy is not AutoBlockPolicy.OFF` guard settles it either way; pick which contract `off` promises.

services/safety/investigation.py
```diff
@@ -398,7 +434,32 @@ async def _enact(
...
+                if await self._feed_repo.add(proposal.list, proposal.domain):
+                    added.append(f"{proposal.list}:{proposal.domain}")
+            log.info(
+                "safety_list_applied",
+                host=event.host,
+                trigger=event.trigger,
+                added=added,
+                proposed=len(verdict.proposals),
+            )
         elif decision.action in ("review", "propose"):
```
Zingzy, line 454: **Should-decide:** the silence covers `propose` as well as `review`, and a persisted proposal has no discovery path: ops-mcp has `spoo_feed_add` (apply) but nothing lists verdicts with unapplied proposals. The test docstring in this PR states the doctrine itself: a propose ping carries something to approve. That is exactly what distinguishes it from the dead-page review asks that motivated the silence. Either exempt `propose` from the silent-trigger rule, or add a pending-proposals listing on the ops side. As written, a waa.ai-class discovery arriving via sweep gets recorded and never seen.

services/safety/investigation.py
```diff
@@ -42,6 +44,14 @@
 
 _DEFAULT_PROMPT = (Path(__file__).parent / "prompts" / "investigate_v1.md").read_text()
 
+# Nobody asked these, so nobody is waiting on the answer: an unresolved
+# coverage sweep is a record, not an escalation.
+_SILENT_TRIGGERS = frozenset({"sweep", "hot", "redirect"})
```
Zingzy: **Nit:** second identical definition (`analyzer.py:64` has the first), and the comment in `_enact` even says it is the same rule the screening tier uses. Two copies of a trigger set drift; one home, probably `events.py` next to the event they classify.

services/safety/investigation.py
```diff
@@ -42,6 +44,14 @@
...
+
+# The one list a model verdict may grow on its own: membership only marks a
+# link for chain resolution, never refuses it.
+RESOLVE_ONLY_LIST = "redirectors"
```
Zingzy: **Nit:** third copy of this feed name (`feeds.py` has `REDIRECTOR_FEED`, `tools.py:504` has a literal tuple). No import cycle stops you: `feeds` pulls only sweeps/events/sinks. Import the existing constant instead of minting another.

infrastructure/browser_run.py
```diff
@@ -52,12 +57,22 @@ def configured(self) -> bool:
...
-        failure (logged). The browser follows JS/meta redirects the plain
-        resolver cannot see — what it lands on is the truth of the page."""
+        """Render *url* and return HTML + screenshot, or None once every
+        wait condition has failed (each logged). The browser follows
+        JS/meta redirects the plain resolver cannot see — what it lands on
+        is the truth of the page."""
         if not self.configured:
             log.warning("browser_run_unconfigured")
             return None
+        for attempt, wait_until in enumerate(_WAIT_LADDER, start=1):
```
Zingzy, line 76: **Nit:** the ladder retries on every failure kind, not just wait timeouts, so a genuinely dead host (per this PR's own data, the most common case in sweeps) now costs two Browser Run API calls and up to ~90s. Bounded by the daily budgets, so a nit, but matching on the timeout error class would halve the cost of exactly the pages the second condition cannot help.

tests/unit/services/safety/test_investigation.py
```diff
@@ -392,3 +399,236 @@ async def test_pattern_reaches_the_operator_and_the_verdict(self):
...
+
+    @pytest.mark.asyncio
+    async def test_apply_list_writes_the_feed_and_stays_silent(self):
+        inv, verdict_repo, _enforcer, notifier = _investigator(
+            _verdict(
+                Classification.REDIRECTOR_SERVICE,
+                proposals=[self._proposal("redirectors", "forms.gle")],
+            )
+        )
+        inv._feed_repo = AsyncMock()
```
Zingzy: **Nit:** reaching into the private attribute instead of the constructor. `_investigator` taking a `feed_repo` kwarg matches how `_build` grew `**kwargs` in test_analyzer for the same need.

Zingzy (commented):
Nothing blocking, and this is the strongest PR in the safety series: every change traces to a named host from the first days of real traffic, and all four moves serve the framework's actual objective (minimize reputation damage per unit of traffic sacrificed; the render is the evidence; a ping must be a 5-second decision). Ran the safety suite on the branch: 439 passed.

Two should-fixes, two should-decides, three nits, inline. The red codecov/patch check is legitimate: `FeedDomainRepository.add()` is only ever mocked.

Merge ships nothing, but on the next release four behaviors go live with no flag: the wait ladder, screenshots riding to the model, sweep silence, and redirectors self-apply (note: the last one is active even under the default `corroborated` policy, see the `decide_authority` comment). The note-hardening and `add()` coverage are the tail I'd finish before flipping `SAFETY_DEEP_AUTOBLOCK=confident`.

services/safety/investigation.py
```diff
@@ -398,7 +434,32 @@ async def _enact(
                 legacy_count=legacy_count,
                 sample_url=event.url,
             )
+        elif decision.action == "apply_list":
+            added: list[str] = []
+            for proposal in verdict.proposals:
+                if self._feed_repo is None:
+                    break
+                if await self._feed_repo.add(proposal.list, proposal.domain):
+                    added.append(f"{proposal.list}:{proposal.domain}")
```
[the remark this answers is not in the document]
Zingzy: Fixed. A missing feed repo now logs safety_list_apply_unwired with the proposals it could not apply, and returns before anything claims success.

services/safety/tools.py
```diff
@@ -269,16 +288,33 @@ def trim_html(html: str) -> str:
...
         f"external script hosts: {', '.join(hosts) or 'none'}",
+        f"embedded destinations: {'; '.join(embedded) or 'none'}",
         f"visible text (capped): {visible or '(none)'}",
     ]
+    if _is_bot_challenge(parser.title, visible):
+        parts.append(
+            "NOTE: this is an anti-bot challenge page, not the site. The "
+            "render was refused, so this is MISSING evidence about the "
+            "destination and says nothing about whether it is safe."
+        )
```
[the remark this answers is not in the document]
Zingzy: Fixed. The note now says the page matches anti-bot challenge markers, not that the render was refused, and it only fires when the page also has no forms and almost no text.

services/safety/tools.py
```diff
@@ -210,6 +224,11 @@ def handle_starttag(self, tag, attrs):
         elif tag == "form":
             self._form_fields = []
             self._form_action = _one_line(a.get("action", ""), 200)
+        elif tag in ("frame", "iframe") and a.get("src"):
+            # A frameset has no text of its own; the target IS the page.
+            self.embedded.append(f"{tag} src={_one_line(a['src'], 200)}")
+        elif tag == "meta" and a.get("http-equiv", "").lower() == "refresh":
+            self.embedded.append(f"meta refresh={_one_line(a.get('content', ''), 200)}")
```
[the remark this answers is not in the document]
Zingzy: Fixed. The parser keeps at most _MAX_EMBEDDED entries and counts the total separately for the overflow suffix, so an attacker page cannot grow the list during parse.

services/safety/tools.py
Zingzy: Fixed. The note now requires the interstitial shape as well as the markers: no forms and under 400 chars of visible text. A scam page hiding "verifying you are human" next to a credential form keeps the form in evidence and gets no note. Reworded to "matches anti-bot challenge markers". Test added for exactly that page.

repositories/feed_domain_repository.py
Zingzy: Fixed both halves. add() reduces the input through registrable_domain() and refuses anything that is not a hostname, so waa.ai/path becomes waa.ai rather than an inert row. ListProposal.list is Literal["shorteners", "redirectors"], so an unknown list fails parse into the LlmTaskFailed path. Direct tests for add() live in the existing test_feed_domain_repository.py.

services/safety/investigation.py
Zingzy: Decided: off means no autonomous action at all. apply_list now also requires policy is not OFF; under OFF it falls through to propose. Test added.

services/safety/investigation.py
Zingzy: Decided: propose is exempt from the silent-trigger rule. The silence exists for dead-page review asks that carry nothing to act on; a proposal carries something to approve, so it reaches the operator on any trigger. Test asserts a sweep-sourced shorteners proposal still pings with the proposal in context.

services/safety/investigation.py
Zingzy: Fixed. One definition, SILENT_TRIGGERS in events.py; analyzer and investigation both import it.

services/safety/investigation.py
Zingzy: Fixed. RESOLVE_ONLY_LIST is gone; investigation imports REDIRECTOR_FEED from feeds, and the literal tuple in tools.py now uses the feed constants too. No cycle, as you said.

infrastructure/browser_run.py
Zingzy: Fixed. The ladder only advances when the failure body carries Browser Run's 6002 goto-timeout code; a 5006 dead host stops after one call. Tests cover both.

tests/unit/services/safety/test_investigation.py
Zingzy: Fixed. _investigator takes a feed_repo kwarg; no test reaches into _feed_repo any more.

PR #336, opened 2026-09-02
fix(safety): a fake verification gate is the payload, not a wall

Zingzy:
## Why

Of 18 medium-confidence scam verdicts since go-live, 7 were fake verification gates: a hand-drawn "Security Check / I'm not a robot / 3-D Secure" card on a numeric or .cfd/.shop throwaway domain, days old, no MX, nothing at the root. The model described them accurately, named ClickFix in its own reason, and said `medium`, so under `SAFETY_DEEP_AUTOBLOCK=confident` nothing happened.

It was following the prompt. The only confidence rule is "high means you would stake the block on it", and every `high` example shows the model seeing the payload directly. Nothing said a fake gate on a young spoofing domain **is** the payload.

## Change

One rule in the cloaking guidance and one worked example (Example 10) built from `5347567.shop`, the only one of the seven whose gate is still up. A fake gate that is not a real challenge provider (no reCAPTCHA, hCaptcha, Turnstile or `cdn-cgi/challenge-platform`) on a young, brand-imitating, MX-less domain with no legitimate content anywhere is `scam_host / high`. A real provider challenge on an established domain stays missing evidence, as before.

A test pins the rule and the example so a later prompt edit cannot quietly drop them.

## Eval

Same 17 medium-confidence hosts, real tools, real model, enforcement and notifications off, both prompts run side by side via `LLM_PROMPT_DIR`:

```
host                                        CURRENT prompt            NEW prompt
5347567.shop                                scam_host/medium          scam_host/high        UP  (the live gate)
p0shmark.437905182.shop                     scam_host/medium          scam_host/high        UP
eucuffexbo.temp.swtest.ru                   scam_host/medium          scam_host/high        UP
cpotyrover.temp.swtest.ru                   scam_host/high            scam_host/medium      down
s.id                                        scam_host/medium          redirector_service/medium
premiumtvs.space                            scam_host/medium          uncertain/low
www.link-for-unsubscribe.site               scam_host/medium          spam_gray/medium
cd.runzehui.net                             compromised_legit/medium  spam_gray/medium
id8754285.cfd                               uncertain/medium          uncertain/low
www.carmelites.org.au                       benign/high               benign/high
(7 others unchanged)

scam-class HIGH:   5/17 -> 7/17
benign -> scam:    0
```

The current prompt's verdict on the live gate reads "a known pattern for malware delivery (ClickFix)" and still says medium. The new prompt's reads "hand-drawn fake 'I'm not a robot' verification card with no real captcha provider" and says high.

Two caveats. Six of the seven gate hosts are already dead (404/422), so under either prompt they correctly return uncertain; the rule could only be tested on one. And the model shows run-to-run variance regardless of prompt (`s.id` came back high, medium and redirector_service across three runs today), so single-row moves like `cpotyrover` are within noise. What is not noise: nothing benign or gray was condemned.

## Also in this PR: sampling at temperature 0

The runner set a timeout and the cache flags and nothing else, so every investigation sampled at Anthropic's default of 1.0. The same evidence could land on `high` one run and `medium` the next by chance alone; `s.id` came back `high`, `medium` and `redirector_service` across three runs of one day. `LLM_TEMPERATURE` now defaults to 0 and is passed to the agent. This removes sampling noise on identical evidence. It does not remove noise from evidence that genuinely changed between runs, which for a shortener fronting a rotating cloaker is real.

## Also: the challenge note no longer contradicts the rule

The `#335` challenge note fired on marker text alone, so a ClickFix kit that copies Cloudflare's wording verbatim, formless and thin by construction, got told "missing evidence" by the tool while this rule said judge it. The parser now records whether a real challenge provider script loaded (`/recaptcha/`, `hcaptcha.com`, `challenges.cloudflare.com`, `cdn-cgi/challenge-platform`, matched on the URL since a Google clone loads gstatic too). Markers plus provider keeps the missing-evidence note. Markers with no provider emits the opposite: a real challenge always loads its provider, so this is a hand-drawn fake gate and positive evidence.

Known hole, by design: a kit that includes an inert Turnstile tag next to its card passes the provider check and drops back to medium. Under-block direction, and the screenshot still reaches a human on review.

## Not in this PR

Whether medium-confidence `scam_host` should block at links scope. Six of seven gates rotated out within two days of being judged at medium, which is the argument for acting at judgment time.


services/safety/prompts/investigate_v1.md
```diff
@@ -65,6 +65,7 @@ Judge by these pairs — this is where the errors happen:
...
 - A render served from a datacenter/scanner egress that looks clean while the report text is specific and damning is a CLOAKING hypothesis, not an acquittal — good kits serve scanners a clean page. Weigh the report against the clean render; do not treat "looked fine" as proof.
+- A fake verification gate is the payload, not a wall in front of it. A "Security Check", "I'm not a robot", "verify you are human" or "checking your browser" card that is NOT a real challenge provider (no reCAPTCHA, hCaptcha or Turnstile script host, no `cdn-cgi/challenge-platform`) on a domain that is days old, is a numeric or random string, imitates a brand at its root, or has no MX and no legitimate content on any path, is the ClickFix pattern: the gate itself is the social-engineering step. You have seen the harm. That is `scam_host`, `high`, and the screenshot is your evidence. Do not downgrade to `medium` because you "could not see behind" it; there is nothing behind it you need. A REAL provider challenge on an established domain (a Cloudflare "Just a moment..." with the challenge-platform script, on a domain with history) is still missing evidence, as before.
```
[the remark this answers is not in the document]
Zingzy: Fixed in 7da3c18. The rule now needs two things together: a RENDERED non-provider gate (a failed render or 404 is explicitly not a gate) and at least two of {registered or first cert within ~30 days, numeric/random hostname, brand imitation, no MX, no legitimate content at root or reported path}. It also states the negative: a fake gate on an established domain with a real root is compromised_legit or uncertain, never scam_host high. The first draft's 'or' is exactly why p0shmark went high on a dead page from hostname shape alone.

services/safety/prompts/investigate_v1.md
```diff
@@ -118,6 +119,10 @@ Verdict: `spam_gray`, `medium`. Reason: "Aggressive affiliate content with ad wa
...
+Verdict: `scam_host`, `high`. Reason: "Numeric throwaway domain serving a hand-drawn 'I'm not a robot' card with no real challenge provider and nothing at its root: the gate is the ClickFix lure, not a barrier in front of one." Evidence: ["fake verification card, no captcha provider script", "'3-D Secure' misused on a robot check", "2-day-old cert, no MX, empty root", "no legitimate content on any path"]. Scope: `host`. — Contrast with Example 9: there the gate was a real maintenance page on a 2-year-old domain and the harm was only alleged. Here the domain has no history, no product, and the card is the harm. `medium` would be the wrong answer; you are not waiting to see a payload, you are looking at it.
```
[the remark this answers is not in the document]
Zingzy: Fixed in 7da3c18. Example 10 now says 'first TLS certificate 2 days old (RDAP unavailable)' rather than a registration age, claims only what the bundle checked (root returns 'Cannot GET /', the reported path), drops 'on any path', and counts its signals explicitly: gate rendered plus four second-part signals.

tests/unit/services/safety/test_investigation.py
```diff
@@ -706,3 +706,20 @@ def test_a_proposal_for_an_unknown_list_fails_parse(self):
...
+
+    def test_rule_and_worked_example_are_present(self):
+        from services.safety.investigation import _DEFAULT_PROMPT as p
+
+        assert "A fake verification gate is the payload" in p
+        assert "ClickFix" in p
+        assert "Example 10" in p
+        assert "5347567.shop" in p
+        # The counter-case stays: a real provider challenge is still missing evidence.
+        assert "cdn-cgi/challenge-platform" in p
```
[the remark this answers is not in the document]
Zingzy: Fixed in 7da3c18. Two new assertions pin the qualifiers: the two-things-together structure, the rendered-gate requirement, the non-provider requirement, the dead-page exclusion, the two-of threshold, the never-high negative; and Example 10 must carry scam_host/high, the observed cert and root facts, the signal count, and must not contain 'on any path'.

services/safety/prompts/investigate_v1.md
```diff
@@ -65,6 +65,7 @@ Judge by these pairs — this is where the errors happen:
...
...ogether. First, you RENDERED the gate: a "Security Check", "I'm not a robot", "verify you are human" or "checking your browser" card that is NOT a real challenge provider (no reCAPTCHA, hCaptcha or Turnstile script host, no `cdn-cgi/challenge-platform`). A page that failed to render, or a dead 404, is not a gate; it is missing evidence and stays `uncertain`. Second, at least two of: domain registered or first certificate within the last ~30 days; a numeric or random-string hostname; a brand name imitated in the host or at the root; no MX; no legitimate content at the root or the reported path. With both, the gate itself is the social-engineering step (ClickFix). You have seen the harm. That is `scam_host`, `high`, and the screenshot is your evidence; do not downgrade to `medium` because you "could not see behind" it. Name the signals you counted in `evidence`. Without the second part, a fake gate on an established domain with a real root is `compromised_legit` or `uncertain`, never `scam_host` `high`. A REAL provider challenge on an established domain (a Cloudflare "Just a moment..." with the challenge-platform script, on a domain with history) is still missing evidence, as before.
```
Zingzy, line 68: **Should-fix:** the challenge note in `tools.py` undermines this rule for the strongest lookalikes. `_is_bot_challenge` (tools.py:319) fires on marker text alone: a formless, thin page saying "Verifying you are human" gets the code-voiced note "likely the interstitial rather than the site: MISSING evidence." A real ClickFix gate is formless and thin by construction, and the kits that imitate Cloudflare verbatim use exactly those marker phrases. For those pages the tool output says missing evidence while this rule says judge it high.

Example 10 dodges the collision only because 5347567.shop's card happens to use non-marker wording ("Please confirm you are not a robot", "Security Check"), so the eval never exercised it. The distinguishing fact is one the parser already collects: `script_hosts`. When markers match but no real provider script is present (no reCAPTCHA/hCaptcha/Turnstile host, no `cdn-cgi/challenge-platform`), emit the opposite hint (challenge-shaped wording with no challenge provider, see the fake-gate rule); with a provider script present, keep the current note. Miss direction if unfixed is under-blocking, so not a blocker, but this PR created the contradiction and is the natural place to close it.

Separate known limitation, no ask: the provider-script test is spoofable in the attacker's favor. A kit that includes an inert Turnstile script tag next to its hand-drawn card fails the "NOT a real challenge provider" condition and drops back to medium. Under-block direction, consistent with the framework's economics, and the screenshot still reaches a human on review. Inherent to any published criterion; just know the hole exists.

config.py
```diff
@@ -514,6 +514,8 @@ class LlmSettings(BaseSettings):
     # chain) before concluding. 10 cut those off mid-investigation.
     max_tool_calls_per_run: int = Field(default=18, ge=1)
     max_total_tokens_per_run: int = Field(default=120_000, ge=1000)
+    # 0: the same evidence should land on the same verdict. Provider default is 1.0.
+    temperature: float = Field(default=0.0, ge=0.0, le=1.0)
```
Zingzy, line 518: **Note, nothing to do now:** this is process-global. `LlmTask` has no per-task temperature, so when the report-inbox scanner (the docstring's named next consumer) arrives, it inherits 0. Right default for that task too, most likely; the seam moves to `LlmTask` the day a task wants sampling.

Zingzy (commented):
Nothing blocking. This is the right kind of fix for the miss it describes: the model was obedient, not wrong, since the only confidence rule was "high means you'd stake the block on it" and every high example showed a directly observed payload. Fixing that in the prompt, pinning the rule and example with tests, and letting the v1+sha8 provenance hash mark the boundary is how prompt changes should land. Temperature 0 for a judgment task is overdue, and the test proving the setting actually reaches the agent is the part most people skip. Ran the affected suites on the branch: 66 passed.

One should-fix that belongs in this PR's scope (inline on the rule): the challenge note in tools.py now contradicts this rule for exactly the most Cloudflare-imitating gates.

One note with no diff line to hang it on: the eval predates temperature 0 (or doesn't say otherwise) and is one run per cell with admitted variance. The aggregate claims hold anyway, and 0 benign-to-scam is the one that matters. Now that runs are deterministic and cheap to repeat, re-running the same 17-host harness at temp 0 would make the next calibration PR's table mean more.

Merge ships nothing, but unlike the last two safety PRs there is no flag between this prompt and prod behavior: AUTOBLOCK has been `confident` since Sep 3, so the two extra highs in the eval table are two real host-wide auto-blocks. That's the stated purpose, and the two-signal floor plus the 0-condemnations eval is what makes it defensible. The deliberately excluded question (medium scam_host acting at links scope, given six of seven gates rotated out within two days) deserves its own PR soon.

services/safety/prompts/investigate_v1.md
Zingzy: Fixed in 3deaffd. The parser now records whether any <script src> is a real challenge provider (matched on the URL, not the host, since a Google-homepage clone loads gstatic too: /recaptcha/, hcaptcha.com, challenges.cloudflare.com, cdn-cgi/challenge-platform). _is_bot_challenge became _challenge_shape returning real, fake or nothing. Markers plus provider keeps the current missing-evidence note. Markers with NO provider now emits the opposite: 'a real challenge always loads its provider; this is a hand-drawn fake gate: POSITIVE evidence, see the fake-gate rule.' Tests cover both shapes plus the three providers, and the kisalt fixture now includes the challenge-platform script it really has. Mutation-checked: ignoring the provider flag turns the fake-gate test red. The inert-Turnstile-tag hole you named is real and stays: under-block direction, and the screenshot still reaches a human on review; noted in the PR body.

config.py
Zingzy: Agreed, leaving it process-global. LlmSettings owns the plumbing and 0 is the right default for the inbox scanner too; the seam moves to LlmTask the day a task wants sampling.

Zingzy:
Follow-up: the false-positive half of the eval, run after merge against the deployed prompt (`v1+87da35e9`), same harness, real tools and model, enforcement off.

20 benign hosts: 13 sampled from verdicts already judged `benign / high` (Anthropic, NVIDIA developer, Al Arabiya, x.com, Meesho, Pixeldrain, a GitHub Pages tenant, a chatgpt.site tenant, and others) plus seven known-good destinations chosen as the hardest cases for a rule about gates and redirectors: github.com, linkedin.com, docs.google.com, instagram.com, bit.ly, tinyurl.com, forms.gle.

```
scam-class verdicts:        0 / 20
benign or legit_relay:     16
redirector_service/high:    2   forms.gle, tinyurl.com   (correct class, host scope)
uncertain/low:              2   docs.google.com (sign-in gated), a lead-gen tenant
```

Nothing benign was condemned. The two `uncertain` are the safe direction and do not act under `confident`.

PR #337, opened 2026-09-03
feat(safety): say how far a block reached, and show what the model saw

Zingzy:
## Two gaps in the Discord embeds

**Scope was invisible.** The auto-block embed said what host and how many links, but not whether that meant the whole domain, one path pattern, or a single judged link. On some paths the scope was a parenthetical inside the reason; on `block_host` it was absent entirely. It is now a dedicated `Scope` field on every block embed (`host-wide`, `pattern: <regex>`, `the judged link only`, `… (re-enforced existing verdict)`), and `Reason` carries only the model's reason. The screening tier's next-step status (sent to investigation, needs review, feed listing is the answer) is a separate `Follow-up` field, so a link-scoped block can never read "host-wide decision" beside a Scope that says otherwise.

**No evidence.** The screenshot the model judged was captured, handed to the model, and discarded at the end of the run. `fetch_page` now leaves the last rendered screenshot in a mutable contextvar holder, the same shape as the hard-hit flag and for the same reason (tools run in a child task and must mutate, never rebind). The investigator attaches it to both the block embed and the review embed. Every render is kept, and the one attached is the render of the **judged URL**, not the last one: the prompt tells the model to fetch the destination and then the domain root, so last-wins would have put a blank root page beside a reason describing the gate. Discord webhooks take it as a `multipart/form-data` file with the embed pointing at `attachment://evidence.webp`; when there is no render the embed stays plain JSON.

Screening-tier blocks (feed hits, blocklist patterns) have no render and attach nothing.

## Where it lives

Review pointed out that `infrastructure/ops_notify.py` is a shared operator-channel module and the safety-shaped parameters had no business there. The shared module now exposes one generic `send_embed(channel, title, color, fields, image=None)` that owns everything Discord: routing, footer, the 1024-character field cap (long patterns clip on display, the verdict keeps the full text), an image as a multipart attachment, https-only webhook URLs, and no redirect following. `safety_action` and `safety_review` moved to `services/safety/notify.py` as `SafetyNotifier`, composed over an `OpsNotifier`; that is where Scope, Follow-up, legacy count and the screenshot are decided. Contact and report-intake keep the raw notifier.

## Tests

Scope field present and named when missing; multipart path with attachment reference vs plain JSON; review embed carries the image; `_enact` passes scope and the rendered screenshot (mutated from a child task, as the agent does); `fetch_page` stashes and `reset_last_render` clears. Mutation-checked: removing the stash line turns three tests red.


infrastructure/ops_notify.py
```diff
@@ -215,12 +220,17 @@ async def safety_action(
...
         sample_url: str | None,
+        scope: str = "",
+        screenshot: bytes | None = None,
     ) -> bool:
         """Enforcement already happened — this states the ACTION TAKEN, it
         is not a request for one. ``legacy_count`` is v1/emoji links
-        blocked via their safety flag (same 451 as v2, reversible)."""
+        blocked via their safety flag (same 451 as v2, reversible).
+        ``scope`` is its own field: "host-wide" and "one link" are read
+        very differently by the person deciding whether to intervene."""
```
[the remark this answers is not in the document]
Zingzy: Fixed in c365a8f. Reason now carries only verdict.reason on the screening path too. The follow-up status (sent to investigation / needs review / feed listing is the answer) is its own Follow-up field, rendered only when present, so a link-scoped block can never read 'host-wide decision' next to a Scope that says otherwise. Three analyzer tests moved from asserting on reason text to asserting on follow_up, plus two notifier tests for the field.

infrastructure/ops_notify.py
```diff
@@ -215,12 +221,18 @@ async def safety_action(
...
         sample_url: str | None,
+        scope: str = "",
+        screenshot: bytes | None = None,
+        follow_up: str = "",
     ) -> bool:
         """Enforcement already happened — this states the ACTION TAKEN, it
         is not a request for one. ``legacy_count`` is v1/emoji links
-        blocked via their safety flag (same 451 as v2, reversible)."""
+        blocked via their safety flag (same 451 as v2, reversible).
+        ``scope`` is its own field: "host-wide" and "one link" are read
```
Zingzy: ops_notify is supposed to be a shared package for multiple ops right? so why are we hardcoding it around link block notifications? things like, blocked_count, screenshot, follow_up seems to be specific to safety ops only?

services/safety/tools.py
```diff
@@ -523,6 +548,9 @@ async def fetch_page(url: str) -> str | ToolReturn:
         text = f"rendered via {result.egress}\nurl: {url}\n{trimmed}"
         if not result.screenshot:
             return text
+        held = _last_render.get()
+        if held is not None:
+            held.screenshot, held.url = result.screenshot, url
```
Zingzy: **Should-fix:** last-render-wins attaches the wrong screenshot on the flagship flow. The prompt instructs the model to fetch the destination AND, separately, the domain root, so on exactly the Example 10 sequence (fetch the fake gate, then fetch the root, which returns "Cannot GET /") the embed attaches the blank root page next to a scam_host verdict whose reason describes the gate. The docstring's justification for the stash ("the model's reason describes what was in it") is precisely what last-wins fails to guarantee, and it degrades the 5-second decision this PR exists to enable: an operator sees an empty page beside "fake verification card" and has to go look anyway.

The fix is half-written already: `held.url` is stashed here and read nowhere (as merged it is a dead field). Prefer the render whose URL matches the judged event URL (`without_query` comparison, same as everywhere else), fall back to the last one. One small change in `last_render_screenshot` plus a test where the model fetches destination-then-root and the destination's shot wins.

tests/unit/services/safety/test_tools.py
```diff
@@ -557,3 +557,31 @@ async def test_wait_timeout_earns_the_next_rung(self):
...
+        fetch_page = next(
+            t for t in build_investigation_tools(deps) if t.__name__ == "fetch_page"
+        )
+
+        await fetch_page("https://x.example")
+
+        assert last_render_screenshot() == b"px"
+
+    @pytest.mark.asyncio
+    async def test_reset_clears_the_previous_investigation(self):
```
Zingzy, line 621: **Nit:** this never sets a value before resetting, so it passes trivially on a fresh contextvar and doesn't test what its name claims. Stash something, then reset, then assert empty.

Zingzy (commented):
Nothing blocking. This closes the loop the strategy doc demanded (the human gets screenshot + verdict + reason attached, a 5-second decision instead of an investigation) and kills the "host-wide decision" wording next to link-scoped blocks. The scope/follow-up split lands at the right place: the notifier owns the embed shape, callers state facts, Reason carries only the model's reason. The multipart/plain-JSON fork is tested in both directions, the child-task mutation test mirrors how pydantic-ai actually dispatches tools, and "unspecified" names a missing scope instead of hiding it. Verified `HttpClient.post` forwards **kwargs to httpx, so the `data=`/`files=` path is sound; a Discord rejection degrades to the existing False-and-log path. Ran the affected suites on the branch: 253 passed.

One should-fix on the evidence-selection rule, one test nit, inline.

Merge ships nothing; on release this is operator-facing only, no flags, model-facing behavior untouched.

infrastructure/ops_notify.py
Zingzy: You're right, and I'd made it worse by adding more safety-shaped parameters to a shared module. Fixed in 4681d62: `infrastructure/ops_notify.py` now exposes one generic `send_embed(channel, title, color, fields, image=None)` that owns everything Discord (routing, footer, the 1024-char field cap, image as a multipart attachment). `safety_action` and `safety_review` moved to a new `services/safety/notify.py` (`SafetyNotifier`, composed over an `OpsNotifier`), along with the two safety colours; that is where Scope, Follow-up, legacy count and screenshot live now. Analyzer takes a `SafetyNotifier`; the worker and wiring wrap `DiscordOpsNotifier` at the three safety call sites. Contact and report-intake keep the raw notifier. The Discord-format tests stay in `test_ops_notify.py` against `send_embed`; the field-composition tests moved to `tests/unit/services/safety/test_notify.py`.

infrastructure/ops_notify.py
```diff
@@ -215,12 +221,18 @@ async def safety_action(
...
+        very differently by the person deciding whether to intervene."""
         fields: list[dict[str, Any]] = [
             {"name": "Destination Host", "value": f"```{host}```"},
+            {"name": "Scope", "value": f"```{scope or 'unspecified'}```"},
```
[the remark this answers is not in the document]
Zingzy: Fixed in 4681d62. `send_embed` bounds every field value to Discord's 1024 chars via `_bound_field`, keeping the head and the code fences, so a long pattern clips on display while the verdict keeps the full text. Test at 2000 chars asserts the result is under the cap and still fenced; mutation-checked by removing the bound.

infrastructure/ops_notify.py
```diff
@@ -280,16 +299,35 @@ async def safety_review(
...
         try:
-            response = await self._http.post(url, json=payload)
+            if screenshot:
+                # Discord webhooks take the image as a multipart file; the
+                # embed refers to it by attachment name.
+                payload["embeds"][0]["image"] = {"url": f"attachment://{_SHOT_NAME}"}
+                response = await self._http.post(
+                    url,
+                    data={"payload_json": json.dumps(payload)},
+                    files={"files[0]": (_SHOT_NAME, screenshot, "image/webp")},
```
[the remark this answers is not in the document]
Zingzy: Fixed in 4681d62. `_deliver` refuses any non-https webhook URL before sending (logged, returns False) and passes `follow_redirects=False` on both the JSON and multipart posts. Test asserts an http URL is refused with no post made, and both paths assert the redirect flag; mutation-checked by disabling the guard.

services/safety/tools.py
Zingzy: Fixed in 334b613. The holder now keeps every render as (url, screenshot) in fetch order, and `last_render_screenshot(prefer_url)` returns the render whose URL matches the judged event URL by `without_query`, scanning most-recent-first, falling back to the last render only when nothing matches. The investigator passes `event.url`. Test reproduces the Example 10 sequence exactly: fetch the gate, then fetch the root, assert the gate's shot is attached (and that a `?utm=1` variant of the judged URL still matches); a second test covers the fallback. Mutation-checked: reverting to last-wins turns both the tools test and the `_enact` test red. `held.url` is no longer dead, it is the key.

tests/unit/services/safety/test_tools.py
Zingzy: Fixed in 334b613. The test now stashes a render through `fetch_page`, asserts it is there, resets, then asserts empty for both the plain and the preferred-URL lookup.

infrastructure/ops_notify.py
```diff
@@ -206,90 +211,71 @@ async def report_summary(
...
+                payload["embeds"][0]["image"] = {"url": f"attachment://{_IMAGE_NAME}"}
+                response = await self._http.post(
+                    url,
+                    data={"payload_json": json.dumps(payload)},
+                    files={"files[0]": (_IMAGE_NAME, image, "image/webp")},
+                    follow_redirects=False,
+                )
+            else:
+                response = await self._http.post(
+                    url, json=payload, follow_redirects=False
```
Zingzy, line 290: **Blocker (this is the whole CI failure):** the two hand-rolled integration fakes pin the old HTTP contract exactly: `_CapturingHttp.post(self, url, json)` at `tests/integration/api_v1/test_contact.py:53` and `tests/integration/api_v1/test_reports.py:93`. This new kwarg raises `TypeError: unexpected keyword argument 'follow_redirects'` inside `_deliver`, the notify fails, the contact/report endpoints 500, and every matrix job goes red (`ops_notify_request_failed` in the CI logs names it).

Fix is two lines: `async def post(self, url, **kwargs)` on both fakes, ideally with `assert kwargs.get("follow_redirects") is False` so the redirect guard stays pinned instead of just tolerated. To be fair to the fake: its strict signature did exactly what the class docstring promises (pin the shape, break on change). It caught a real contract change; it just needs updating with intent.

infrastructure/ops_notify.py
```diff
@@ -206,90 +211,71 @@ async def report_summary(
         }
         return await self._deliver(self._report_url, payload, kind="report_summary")
 
-    async def safety_action(
+    async def send_embed(
```
Zingzy, line 223: **Nit:** the module docstring is now false for this method. It still says callers state WHAT happened and the implementation owns "embed structure, colors, footer", but `send_embed` callers own the title, color, field dicts and code fences; only the footer, bounds, routing and image plumbing stay here. The class docstring's "embed shapes are pinned by the integration tests" is also half-true now that the safety shapes are pinned in `tests/unit/services/safety/test_notify.py`. This file's docstring is its contract statement; update it to describe the hybrid (semantic methods for contact/report, generic embed for composers).

Zingzy (commented):
Re-review of the rewrite (SafetyNotifier split + send_embed + field bounding + https guard + URL-matched screenshot selection).

**One blocker: CI is red on every Python version, and the cause is two test fakes, not the feature.** Details inline on the `follow_redirects` line.

Both findings from my previous review are fixed and I verified them in the code: `last_render_screenshot(prefer_url)` picks the judged URL's render via `without_query` with fall-back to the most recent (the `?utm` case is tested), and the reset test now stashes before asserting empty. The dead `held.url` field became the ordered `shots` list that makes the selection possible.

The `SafetyNotifier` split is a clean answer to the maintainer comment: safety vocabulary out of `infrastructure`, colors moved with the composer, no callers left on the removed methods (verified by grep: all eight call sites are analyzer/investigator, both wired through `SafetyNotifier` at both production sites). `_bound_field`'s math is right (6 fence chars + 1017 + ellipsis = 1024), and the https refusal plus `follow_redirects=False` matches what the comment claims. Trade-off accepted by this design, worth stating once: `send_embed` callers now compose Discord-shaped field dicts, fences and colors, so the notifier protocol is no longer swappable in principle. At one operator on one Discord, that costs nothing today.

infrastructure/ops_notify.py
```diff
@@ -206,90 +211,71 @@ async def report_summary(
...
-                }
-            )
-        if sample_url:
-            fields.append({"name": "Sample URL", "value": f"```{sample_url}```"})
+        """The generic operator embed. Callers own WHAT the fields say; this
+        owns Discord: routing, the footer, field limits, and how an image
+        rides along. Field values are bounded here because Discord rejects
+        the whole message over 1024 characters and a lost notification is
+        worse than a clipped one."""
+        url = self._report_url if channel == "report" else self._contact_url
```
Zingzy, line 242: **Nit:** an unknown `channel` silently routes to the contact webhook; a typo like `"reports"` would deliver safety block embeds to the contact channel with no signal. Cheap to close: `Literal["report", "contact"]` on the parameter, or refuse-and-log like the empty-URL case above.

Zingzy:
CI was red on every Python version while my local run was green: I had only run `tests/unit`. The integration suites fake `HttpClient` with a `_CapturingHttp` whose `post(url, json)` was narrower than the `post(url, **kwargs)` it stands in for, so the new `follow_redirects=False` raised `TypeError` and the contact and report endpoints returned 500. Widened both doubles to the real contract in 35baf66. Nothing in production code changed for this.

infrastructure/ops_notify.py
Zingzy: Fixed in 35baf66 (the push before this one): both `_CapturingHttp.post` fakes widened to `post(self, url, json=None, **kwargs)`, the surface of the `HttpClient.post` they stand in for. Also the reason I now run `tests/unit tests/integration` before pushing rather than unit alone.

infrastructure/ops_notify.py
Zingzy: Fixed in 7200f3b. Module docstring now describes both shapes: the named methods are semantic and own their embeds; `send_embed` is the generic where the caller owns title, color and field text and this module owns only Discord's business (routing, footer, field cap, image attachment, https-only, no redirects), with a pointer to `services/safety/notify.py`. Class docstring now says the integration tests pin the named methods' shapes and that `send_embed`'s field text is pinned by its callers' tests.

infrastructure/ops_notify.py
Zingzy: Fixed in 7200f3b. `channel` is `Literal["report", "contact"]` on both the Protocol and the implementation, and `send_embed` refuses anything else at runtime with `ops_notify_unknown_channel` logged and `False` returned, same shape as the empty-URL case. Test asserts `"reports"` is refused with no post made.

PR #338, opened 2026-09-03
fix(safety): say what a block does to future links, drop blocklist claim

Zingzy:
## What

The Discord block embed gets a **Future links** field, derived from the block's scope kind:

| scope kind | Future links |
|---|---|
| host | new links to this host are REFUSED at create (verdict store, not the blocklist) |
| pattern | new links matching the pattern are REFUSED at create; other paths on the host stay open |
| links | only this exact URL is refused at create; the host stays open |

The pattern branch of `_enact` also claimed the pattern was "proposed for the blocklist". That was never true: nothing in the pipeline writes to the operator blocklist. The verdict store is the create-time gate and it refuses future matches on its own. The claim is removed and a test asserts no future-links text mentions the blocklist.

## Why

The embed answered "how far did this reach on existing links" and nothing else. The operator's real question after a block ping is what happens to the next link, and whether any list changed. Now the embed answers both, and the wording is derived in one place so call sites cannot drift.

## Verification

- `uv run pytest tests/unit tests/integration` (minus the env-only app-token scopes file): 3571 passed, 5 new in `test_notify.py`
- Rendered all three scope kinds locally, fields print as in the table above

## Summary by CodeRabbit

- **Improvements**
  - Safety notifications now clearly identify whether an action applies to an entire host, a URL pattern, or specific links.
  - Notifications can include relevant “Future links” information based on the action’s scope.
  - Pattern-based notifications now use clearer wording and no longer imply that blocklist changes have been proposed.
  - Redirect screening and re-enforced verdict notifications now display more accurate scope details.

### DIFF
--- a/services/safety/analyzer.py
+++ b/services/safety/analyzer.py
@@ -233,6 +233,9 @@ class SafetyAnalyzer:
             legacy_count=result.legacy_count,
             sample_url=event.url,
             scope=scope_note,
+            scope_kind={"host": "host", "path_pattern": "pattern", "links": "links"}[
+                scope
+            ],
             follow_up=follow_up,
         )
 
@@ -272,7 +275,7 @@ class SafetyAnalyzer:
                 reason=reason,
             )
             await self._notify_reenforced(
-                event, result, reason, scope=f"pattern: {pattern}"
+                event, result, reason, scope=f"pattern: {pattern}", scope_kind="pattern"
             )
             return matching_blocked_pattern(event.url, (pattern,)) is not None
         # links scope: already blocked and the create gate refuses the exact URL.
@@ -326,10 +329,16 @@ class SafetyAnalyzer:
                 legacy_count=result.legacy_count,
                 sample_url=event.url,
                 scope=f"the judged link only (reached through {event.host})",
+                scope_kind="links",
             )
 
     async def _notify_reenforced(
-        self, event, result, reason: str, scope: str = "host-wide"
+        self,
+        event,
+        result,
+        reason: str,
+        scope: str = "host-wide",
+        scope_kind: str = "host",
     ) -> None:
         """Blocked something: the operator hears about it. Zero blocks stays quiet."""
         if result.blocked_count + result.legacy_count == 0:
@@ -342,6 +351,7 @@ class SafetyAnalyzer:
             legacy_count=result.legacy_count,
             sample_url=event.url,
             scope=f"{scope} (re-enforced existing verdict)",
+            scope_kind=scope_kind,
         )
 
     async def _escalate(
--- a/services/safety/notify.py
+++ b/services/safety/notify.py
@@ -8,12 +8,22 @@ and that the screenshot the model judged travels with the verdict.
 
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, Literal
 
 from infrastructure.ops_notify import OpsNotifier
 
 ACTION_COLOR = 15548997  # red: automatic enforcement happened
 REVIEW_COLOR = 16705372  # yellow: a human decision is needed
+
+ScopeKind = Literal["host", "pattern", "links"]
+
+# What the verdict store, which is the create-time gate, does from now on.
+# Nothing here touches the operator blocklist; saying so is the point.
+_FUTURE: dict[str, str] = {
+    "host": "new links to this host are REFUSED at create (verdict store, not the blocklist)",
+    "pattern": "new links matching the pattern are REFUSED at create; other paths on the host stay open",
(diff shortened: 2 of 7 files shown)


services/safety/analyzer.py
```diff
@@ -233,6 +233,9 @@ async def _handle_toxic(
             legacy_count=result.legacy_count,
             sample_url=event.url,
             scope=scope_note,
+            scope_kind={"host": "host", "path_pattern": "pattern", "links": "links"}[
```
Zingzy, line 236: **Nit:** this lookup adds a crash path that didn't exist before. `ProviderVerdict.scope` is a bare `str` (providers.py:37), so a fourth scope value from a future provider lands here as a `KeyError` after enforcement but before the notifier, losing the operator ping. Previously any string flowed through as display text. The root fix is the same move `ListProposal.list` got: `Literal["host", "links", "path_pattern"]` on `ProviderVerdict.scope`, after which this mapping is total by construction and the next wrong scope fails at the provider, not in the notify path.

services/safety/notify.py
```diff
@@ -8,13 +8,23 @@
...
 REVIEW_COLOR = 16705372  # yellow: a human decision is needed
 
+ScopeKind = Literal["host", "pattern", "links"]
+
+# What the verdict store, which is the create-time gate, does from now on.
+# Nothing here touches the operator blocklist; saying so is the point.
+_FUTURE: dict[str, str] = {
+    "host": "new links to this host are REFUSED at create (verdict store, not the blocklist)",
+    "pattern": "new links matching the pattern are REFUSED at create; other paths on the host stay open",
+    "links": "only this exact URL is refused at create; the host stays open",
```
Zingzy: **Nit:** "only this exact URL" understates the gate. `verdict_covers` matches links scope through `without_query`, so query, fragment, and trailing-slash variants of the judged URL are refused too. In a PR whose whole point is precise operator wording, say "this URL (query variants included) is refused at create".

Zingzy (commented):
Nothing blocking, CI green, 248 safety tests pass on the branch. Two nits inline.

The important part of this PR is the retraction, and it's correct: I traced the create gate and `ToxicVerdictProvider` reads the verdict store, `verdict_covers` handles all three scopes, and nothing in the pipeline writes the operator blocklist — so the old "proposed for the blocklist" text was indeed a false promise, and each of the three new Future-links texts matches what the gate actually does. The wording living in one `_FUTURE` map with a pin test (including the note that the false claim survived two review rounds) is the right way to retire it.

Zingzy:
Retrospective on this follow-up series (#329, #331, #335, #336, #337, this one), since the run of fixes deserves an explanation.

The follow-ups come from three different places, and only two need fixing.

**Calibration against live traffic.** #331 (feed escalations were almost entirely redundant), the render fixes in #335 (networkidle0 starving on an analytics beacon, framesets read as empty), and #336 (fake verification gates landing at medium confidence) could not have been designed in up front. The system shipped dark, went live, and the first days of real output were replayed against the hosts that produced them. Each of those PRs names its measured failure population and pins the lesson with a test or an eval. This class is the tuning loop working, not churn, and it should keep happening.

**The operator embed was never specified, so it got designed one ping at a time.** #335 decided when pings fire, #337 decided what fields a ping carries and where embed code lives, and this PR adds the field answering what happens to the next link while retracting a wrong claim from the previous one. That is one design conversation (what questions must a block ping answer: what happened, how far, what happens next, what changed) spread across four diffs. The false blocklist claim surviving two review rounds is the receipt. Fix tracked in a separate issue: write the embed contract down and check notification changes against it.

**Contract changes discovered by collision instead of by search.** #329 (the worker's own wiring crash-looped) and the CI break in #337 (a new HTTP kwarg no test fake accepted) are the same failure shape: a contract changed and its implementers were not enumerated first. The feed-provider half is already structural (build_feed_providers is the single composition point). The remaining piece, one shared HTTP fake for integration tests, is tracked in a separate issue.

The series is trending the right way: each PR is smaller and better evidenced than the last. The first class is the plan; the count drops on its own once the two structural fixes land.

services/safety/notify.py
Zingzy: Done. Links text now reads: this URL (query variants included) is refused at create; the host stays open. Test asserts the wording.

services/safety/analyzer.py
Zingzy: Done. ProviderVerdict.scope is now VerdictScope = Literal["host", "links", "path_pattern"], so the mapping here is total by type and a wrong scope fails at the provider. Same move ListProposal.list got.

PR #341, opened 2026-09-03
feat(safety): probe browser that clicks, and sibling links in the bundle

Zingzy:
## What

Two things the investigator could not do before: see what a page **does**, and see the **other links** on the same host.

### 1. Probe browser (`browser/`, `infrastructure/browser_probe.py`)

A small Playwright service in its own container. For each `fetch_page` it loads the URL, waits for JS and meta redirects, screenshots, then clicks up to three of the page's most prominent controls (keyword-ranked: play, verify, continue, download, copy, login...) and records what each click caused:

- pop-ups and where they went, navigations, downloads (cancelled), dialogs (dismissed)
- clipboard writes and notification-permission prompts, via hooks installed before any page script runs
- what appeared on the page after the click: modal text, new form fields, new frames. This walks open shadow roots, because that is where at least one live card-harvesting kit renders its form.

The record goes to the model under `## Observed behaviour`. Both screenshots (as loaded, after the clicks) are attached. Cloudflare snapshots remain the fallback when the probe is down or the page will not load.

Safety: the container joins only its own docker network (the worker joins both), and Playwright aborts any request whose host resolves to a non-public address. Discord attachments now take their name and type from the image bytes, since the probe sends JPEG.

### 2. Sibling links in the evidence bundle

For a host with 15 links or fewer, the bundle lists every link we hold on it, who made each (anonymous or account), the creation window, and where each resolves over HTTP. Shared platforms are skipped: their siblings belong to strangers.

### 3. Prompt

- What a click does is a fact from `## Observed behaviour`, never an inference from markup or script hosts.
- Ad-network and push-SDK script tags are `spam_gray` evidence, never `scam_host` evidence.
- A probe-recorded clipboard write of a command from a verification card is the ClickFix payload itself.
- Example 11: the fake video player that is bait for pop-under ads.

## Why

The ourl.jp verdict this week called a fake video player "designed to trigger malicious downloads or push-notification scams". Nobody had clicked it. Clicking it opens a Monetag pop-under and the page later redirects to a Telegram channel. That is `spam_gray` by our own rubric. The model inferred harm from vendor names because inference was all it had. It also never saw that the same anonymous creator made nine ourl.jp links in three minutes.

## Verified against live pages (image built and run on the VPS)

ourl.jp/EJTQX:
```
loaded: https://ourl.jp/EJTQX → https://ourl.jp/go.php?code=EJTQX&to=... (1 HTTP redirect hops) → landed on https://aceuimg.pages.dev/ [cross-domain] (JS or meta refresh)
click 1: button 'Continue Watching' → opened pop-up https://aceuimg.pages.dev/?player=true; navigated to https://iserinekugel.com/ikqwIHnPoWnLyH/141690 [cross-domain]
```

whatnot.ver314186.cfd (a live phishing kit, `scam_host` in our store):
```
click 1: a 'Verify Account button' → revealed on the page: 'Add bank card' | 'All operations comply with PCI DSS' | 'Card number' | 'MM/YY' | 'CVV' | 'Name on card'; new form fields: text:pm-card, text:pm-exp, text:pm-cvv, text:pm-holder, ...
```

clipboardjs.com (download capture and scroll fallback):
```
click 1: a 'download a ZIP' → started download clipboard.js-master.zip from https://codeload.github.com/...
```

A real Cloudflare challenge and a dead host both come back as honest "nothing observable" or a probe failure that falls back to the snapshot.

Not verified end to end: a live ClickFix clipboard write. No live sample was reachable. The hook is exercised by construction and the formatting by tests.

Probe time per page: 7 to 28 seconds. An investigation calls `fetch_page` two to four times, so expect 30 to 60 seconds more per investigation.

## Rollout

- The browser joins only its own docker network (`probenet`); the worker joins both. `infrastructure/probenet-egress.sh <subnet>` adds DOCKER-USER drops for every private range from that network (with intra-network traffic allowed), so a DNS-rebinding page cannot reach the host or other bridges. Applied on the VPS for 172.31.0.0/24.

- Images build under the same tags as the app (`ghcr.io/spoo-me/spoo-browser`). The deploy pulls and starts `browser` before `app`.
- Prod compose on the VPS already has the `browser` service, the `probenet` network, the worker joined to it, and `SAFETY_BROWSER_PROBE_URL` staged. `docker compose config` validates. Nothing starts until the next release deploys.

## Tests

`uv run pytest tests/unit tests/integration` (minus the env-only app-token scopes file): 3591 passed, 25 new.


docker-compose.yml
```diff
@@ -105,13 +105,28 @@ services:
...
       test: ["CMD", "curl", "-fsS", "http://localhost:8001/health"]
       interval: 15s
       timeout: 5s
       retries: 3
       start_period: 15s
 
+  # Scripted browser probe for the safety deep tier: loads a destination,
+  # clicks its prominent controls, reports what happened. Own network in
+  # prod so a hostile page cannot reach the other services.
+  browser:
```
Zingzy, line 120: **Should-fix: the isolation this comment promises does not exist in this file.** There is no `networks:` stanza anywhere in the repo compose, so the browser joins the default network with mongo, redis and the app — and this is the file self-hosters run. The network boundary is load-bearing, not belt: the route guard resolves hosts itself while Chromium re-resolves to connect, so a DNS-rebinding page (public answer for the guard's getaddrinfo, private answer with TTL 0 for Chromium) can land a request on whatever the container can reach. Chromium's restricted-port list covers the classic cross-protocol targets like 6379, but Mongo on 27017 and any container HTTP endpoint are one rebind away on a shared network.

The fix is small and belongs in this file so the docstring is true of what ships: a dedicated network for `browser`, the worker joining both. On the VPS, a DOCKER-USER iptables rule dropping RFC1918/link-local egress from that network is the cheap belt on top.

Zingzy (commented):
Mergeable; the probe stays inert until the browser service starts at the next release deploy, and the fallback to Cloudflare snapshots means a dead probe costs nothing but the clicks. Ran the full suite on the branch (3608 passed); the codecov red is a config artifact, noted inline.

The epistemics are the best part of this PR and worth naming: the prompt now distinguishes what a click DID (read from the record) from what markup suggests, the ourl.jp case is exactly the miss that distinction prevents, and the record hooks have a useful asymmetry — a hostile page can only fabricate observations that incriminate it, never exonerate it. The layered fallback (probe, then snapshot, then honest missing-evidence), the first-shot-as-the-evidence-image choice, cancelled downloads, dismissed dialogs, service workers blocked at both the context and the hook, popups closed after capture, the guard applied to every request on the context including popups, the paired image tags with browser swapping first ungated, and the live verification against two real kits in the PR body: all right. The Discord attachment sniffing closing the JPEG/webp mismatch end to end is the kind of loose end most PRs leave.

Two should-fixes (one security posture, one flood cap), one should-decide, and nits, inline.

browser/probe.py
```diff
@@ -0,0 +1,500 @@
...
+    clicks: list[_Click] = field(default_factory=list)
+    popups: list[str] = field(default_factory=list)
+    downloads: list[str] = field(default_factory=list)
+    dialogs: list[str] = field(default_factory=list)
+    events: list[list[str]] = field(default_factory=list)
+    blocked: list[str] = field(default_factory=list)
+    tasks: list = field(default_factory=list)
+    current: _Click | None = None
+
+    def event(self, kind: str, detail: str) -> None:
```
Zingzy, line 216: **Should-fix: cap the event count.** Everything else in the record is bounded (detail at 400 chars, blocked at 20, revealed text at 8) but a hostile page can call `window.__probeRec` in a loop and grow `events` without limit: memory in the probe, a giant JSON response, and an `## Observed behaviour` section that stuffs the model's context on the attacker's terms. A cap here (say 100, with a final "...truncated" entry) closes it in three lines. The fabrication direction is already safe — a page can only invent observations that incriminate it — so the flood is the only lever this binding hands the page.

infrastructure/browser_probe.py
```diff
@@ -0,0 +1,190 @@
...
+import base64
+from dataclasses import dataclass
+
+from infrastructure.http_client import HttpClient
+from infrastructure.logging import get_logger
+from shared.url_utils import registrable_domain
+
+log = get_logger(__name__)
+
+EGRESS_LABEL = "our own browser (Hetzner datacenter IP), scripted probe"
```
Zingzy, line 24: **Should-decide: probing from the prod IP.** The label is honest, but this egress is a single stable datacenter IP tied to production. Two costs accrue: kits that fingerprint scanners can note the IP once and cloak for it forever (a static IP burns much faster than Cloudflare's pool), and the traffic profile of clicking scam pages now originates from the same address that serves every redirect. The safety research settled on keeping render egress off the prod box for exactly this reason, and Cloudflare Browser Run was partly chosen for it. Accepting the trade for the click capability is defensible — the fallback still exists when a page cloaks — but it should be a recorded decision, and a cheap egress proxy for the probe is the eventual answer if cloak-rate against this IP climbs.

pyproject.toml
```diff
@@ -61,6 +61,8 @@ omit = [
     "*/conftest.py",
     ".venv/*",
     "main.py",
+    # Separate service with its own runtime (Playwright); not importable here.
+    "browser/*",
```
Zingzy, line 65: **Nit (this is the red codecov/patch check):** this omit silences local coverage for `browser/*`, but there is no `codecov.yml`, so codecov still counts those 500 patch lines as uncovered. An `ignore: ["browser/**"]` in a codecov.yml aligns the two. The remaining misses in `infrastructure/browser_probe.py` (the fallback branches of `format_observations`, `_b64`'s except) are worth a couple of small cases while you're there.

repositories/url_repository.py
```diff
@@ -415,6 +415,32 @@ async def list_by_dest_host_with_urls(
             )
         return [(d["alias"], d.get("domain", ""), d.get("long_url", "")) for d in docs]
 
+    async def list_recent_by_dest_host(
```
Zingzy, line 418: **Nit, merge-order interplay with #342:** this matches `dest.host` only. Once #342 lands, a sibling that reaches this host through a geo rule is invisible here; the semantic part of the rebase is switching this to `dest_host_filter` so the campaign shape includes rule-hidden siblings. Both PRs also edit `build_evidence_bundle`, so the textual conflict will force the visit anyway.

browser/probe.py
```diff
@@ -0,0 +1,500 @@
...
+        rec = _Record(start=time.monotonic())
+        guard = _Guard()
+
+        async def route(r, request):
+            host = urlparse(request.url).hostname or ""
+            if not _is_public_url(request.url) or not await guard.public(host):
+                rec.blocked.append(request.url[:200])
+                await r.abort()
+                return
+            await r.continue_()
```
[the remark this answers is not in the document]
Zingzy: Agreed that the route guard alone cannot pin what Chromium connects to. Enforced at the network layer instead (c54132d): the repo compose now gives the browser its own network with only the worker joining it, and infrastructure/probenet-egress.sh adds DOCKER-USER drops for every private range from that subnet (with a RETURN for traffic inside it, so the worker's calls and the browser's replies pass). Applied on the VPS for 172.31.0.0/24. A rebinding page can then reach nothing private regardless of what it resolves to.

services/safety/tools.py
```diff
@@ -529,38 +548,77 @@ async def resolve_chain(url: str) -> str:
...
+        business with one compromised path from a parked or purpose-built
+        domain."""
         refusal = _fetch_refusal(url)
         if refusal is not None:
             return refusal
+        if deps.probe is not None:
+            probed = await deps.probe.probe(url)
+            if probed is not None:
+                return _render_return(
+                    url,
```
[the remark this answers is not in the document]
Zingzy: Done in c54132d: the tool output keeps `url:` as requested (the _last_render key) and adds `landed on: <final_url> (everything below is that page)` when they differ. Test: TestFetchPageLandedUrl.

docker-compose.yml
Zingzy: Fixed in c54132d. docker-compose.yml now has a `probenet` network: `browser` joins only it, `click-worker` joins default and probenet, so the docstring is true of what ships. infrastructure/probenet-egress.sh is the belt: DOCKER-USER drops from the probe subnet to every private range (RFC1918, link-local, CGNAT, loopback), with a RETURN for intra-subnet traffic first so the worker and browser still talk. Applied on the VPS for 172.31.0.0/24; the chain order is RETURN then six DROPs, ahead of the existing Cloudflare rule.

browser/probe.py
Zingzy: Done in c54132d: MAX_EVENTS = 100; the record appends one final ["truncated", ...] entry and ignores the rest.

pyproject.toml
Zingzy: Done in c54132d: codecov.yml ignores browser/**, and test_browser_probe.py gained cases for the fallback lines of format_observations (missing fields, downloads, dialogs, blocked requests, notification prompt, click failure, probe note) and _b64's except.

infrastructure/browser_probe.py
Zingzy: Recorded as a decision in the PR: we accept probing from the prod IP for the click capability, with Cloudflare snapshots as the fallback when a page cloaks. Two things make it cheap to revisit: the egress label in every render tells us which path produced the evidence, and a cloak-rate climb against this IP shows up as probe results that disagree with the snapshot. If that happens, an egress proxy for the probe container is the next step; the container already has its own network, so routing it through a proxy is a compose change, not a code change. Leaving this open for your call.

repositories/url_repository.py
Zingzy: Agreed. This branch is based on main, which has no dest_host_filter yet. Once #342 merges I will rebase this on top and switch list_recent_by_dest_host to dest_host_filter so rule-hidden siblings appear; the build_evidence_bundle conflict forces that visit anyway. Leaving open as the reminder.

PR #343, opened 2026-09-03
feat(safety): feed-delta sweep and L1 counters see secondary destinations

Zingzy:
Beads ticket spoo-v3f.39. Stacked on #342 (`feat/safety-secondary-destinations`); merge that first.

## What

Two gaps the first PR left on record.

**Feed-delta sweep.** It fans out by `dest.registrable_domain`, so a geo or variant destination under a feed-listed domain whose main host sits elsewhere was never enqueued. `UrlDestination` now stamps `secondary_registrable`, index-aligned with `secondary_hosts`, at every write point and in the backfill, with a sparse index. The sweep matches either field, zips `(host, registrable)` pairs inside Mongo and keeps only pairs under the feed domain, so the clean main host is never enqueued for a domain it does not belong to. Both sweeps now sample a URL on the host they emit instead of the link's main URL, or the analyzer would screen the wrong page.

**L1 burst counters.** Create records one L1 event per distinct destination of the link, so a campaign that hides its host in geo rules crosses thresholds like any other.

Destination sources come from #342: geo rules and the scheduling `pre_start_url`; both flow through here unchanged.

The backfill docstring says to run it from the project env (`uv run python scripts/backfill_url_dest.py`); the inline script header needs PyPI.

## Proof (worktree app with safety and L1 wired to the local queue Redis)

Geo link `cqmh5Ft`: `long_url` on `clean-main-proof.com`, geo rule to `kit.evil-feed-proof.com`.

```
stored dest: {host: clean-main-proof.com, secondary_hosts: [kit.evil-feed-proof.com], secondary_registrable: [evil-feed-proof.com]}
L1 counters: l1:dom:600:evil-feed-proof.com:* = 1, l1:dom:86400:evil-feed-proof.com:* = 1 (and the same for clean-main-proof.com)
feed-delta sweep evil-feed-proof.com  -> [(kit.evil-feed-proof.com, https://kit.evil-feed-proof.com/kit)]
feed-delta sweep clean-main-proof.com -> [(clean-main-proof.com, https://clean-main-proof.com/landing)]
recent-screen listing includes (kit.evil-feed-proof.com, https://kit.evil-feed-proof.com/kit)
backfill after unsetting both fields: "geo links needing secondary fields: 2 ... stamped 2 ... remaining 0"; second run 0
index dest_secondary_registrable (sparse) created at startup
```

## Tests

`GEO_RULES_ENABLED=true uv run pytest tests/unit tests/integration` (app-token scopes file deselected): 3802 unit and integration passed, plus 3 real-Mongo cases in tests/db.

## Deploy order

`$zip` in the feed sweep stops at the shorter list, so a link stamped by #342-era code (secondary_hosts without secondary_registrable) yields no pairs until the backfill's second pass re-stamps it. Merge #342 and #343 before one release and run `uv run python scripts/backfill_url_dest.py` once after deploy. `tests/db/test_secondary_destination_sweeps.py` pins the pipeline against a real Mongo, including that truncation.

### DIFF
--- a/repositories/url_repository.py
+++ b/repositories/url_repository.py
@@ -18,6 +18,7 @@ from infrastructure.logging import get_logger
 from repositories.base import BaseRepository
 from schemas.models.base import ANONYMOUS_OWNER_ID
 from schemas.models.url import UrlStatus, UrlV2Doc
+from shared.url_utils import parse_destination
 
 log = get_logger(__name__)
 
@@ -32,6 +33,43 @@ def dest_host_filter(host: str) -> dict:
 
 # Every host a link can route to, for grouping: main plus secondary.
 _ALL_HOSTS = {"$setUnion": [["$dest.host"], {"$ifNull": ["$dest.secondary_hosts", []]}]}
+# Every URL the link routes to; the sweeps pick the one on the host they emit.
+_ALL_URLS = {
+    "$concatArrays": [
+        ["$long_url"],
+        {
+            "$map": {
+                "input": {"$objectToArray": {"$ifNull": ["$geo_rules", {}]}},
+                "as": "rule",
+                "in": "$$rule.v",
+            }
+        },
+        {
+            "$cond": [
+                {"$eq": [{"$type": "$pre_start_url"}, "string"]},
+                ["$pre_start_url"],
+                [],
+            ]
+        },
+    ]
+}
+# (host, registrable) pairs for every destination; secondary_registrable is
+# index-aligned with secondary_hosts by construction (UrlDestination.for_link).
+# $zip stops at the shorter list: a doc stamped before secondary_registrable
+# existed yields no pairs until the backfill's second pass re-stamps it.
+_HOST_PAIRS = {
+    "$concatArrays": [
+        [["$dest.host", "$dest.registrable_domain"]],
+        {
+            "$zip": {
+                "inputs": [
+                    {"$ifNull": ["$dest.secondary_hosts", []]},
+                    {"$ifNull": ["$dest.secondary_registrable", []]},
+                ]
+            }
+        },
+    ]
+}
 
 
 # Mongo caps a single query document at 16MB; a host-wide block can name
@@ -44,6 +82,24 @@ def _alias_chunks(
 ) -> list[list[dict[str, str]]]:
     clauses = [{"alias": a, "domain": d} for a, d in pairs]
     return [clauses[i : i + _ALIAS_CHUNK] for i in range(0, len(clauses), _ALIAS_CHUNK)]
+
+
+def _host_samples(docs: list[dict]) -> list[tuple[str, str]]:
+    """(host, sample URL on that host) per grouped row. A secondary host's
+    sample must be the geo or variant URL, not the link's main destination,
+    or the analyzer screens the wrong page."""
+    out: list[tuple[str, str]] = []
+    for d in docs:
+        host = d.get("_id")
+        if not host:
+            continue
+        urls = [u for u in d.get("urls") or [] if isinstance(u, str) and u]
+        sample = next(
+            (u for u in urls if (parse_destination(u) or {}).get("host") == host),
+            urls[0] if urls else "",
+        )
+        out.append((host, sample))
+    return out
 
 
 class UrlRepository(BaseRepository[UrlV2Doc]):
@@ -562,16 +618,20 @@ class UrlRepository(BaseRepository[UrlV2Doc]):
         pipeline = [
             {
                 "$match": {
-                    "dest.registrable_domain": registrable_domain,
+                    "$or": [
+                        {"dest.registrable_domain": registrable_domain},
+                        {"dest.secondary_registrable": registrable_domain},
+                    ],
                     "status": UrlStatus.ACTIVE.value,
                 }
             },
-            # Main host only: a secondary host's own registrable domain is not
-            # the matched one, so fanning it out here would tag it wrongly.
+            {"$project": {"pairs": _HOST_PAIRS, "urls": _ALL_URLS}},
+            {"$unwind": "$pairs"},
+            {"$match": {"pairs.1": registrable_domain}},
             {
                 "$group": {
-                    "_id": "$dest.host",
-                    "sample_url": {"$first": "$long_url"},
+                    "_id": {"$arrayElemAt": ["$pairs", 0]},
+                    "urls": {"$first": "$urls"},
                 }
             },
             {"$limit": limit},
@@ -584,7 +644,7 @@ class UrlRepository(BaseRepository[UrlV2Doc]):
                 registrable_domain=registrable_domain,
                 limit=limit,
             )
-        return [(d["_id"], d.get("sample_url", "")) for d in docs if d.get("_id")]
+        return _host_samples(docs)
 
     async def list_recent_destination_hosts(
         self, since: datetime, *, limit: int = 20_000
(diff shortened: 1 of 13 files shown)


services/url_service.py
```diff
@@ -940,9 +940,12 @@ async def create(
         inserted_id = await self._url_repo.insert(doc)
         url_doc.id = inserted_id
 
-        # L1 accumulation: count the successful create (best-effort, the
-        # policy service never raises from here).
-        await self._url_policy.record_create(request.long_url)
+        # L1 accumulation for every destination the link routes to
+        # (best-effort, the policy service never raises from here).
+        for destination in link_destination_urls(
```
Zingzy, line 1036: **Should-fix: this can cross the burst threshold with one create.** `link_destination_urls` dedupes by URL STRING, but the L1 counters key on REGISTRABLE DOMAIN (`l1:dom:{window}:{domain}`). One create with a long_url on `x.com` plus geo rules to 50 distinct `x.com` paths bumps the same domain counter 51 times — and `l1_domain_burst_threshold` is 25, so a single link can trip a burst on any domain, including an innocent one someone geo-targets by path. It also drives the redirect probe once per URL, so redirector links get duplicate `trigger="redirect"` events for one host.

Dedupe by registrable domain (or at least host) before this loop, the same way report intake and hot screening already do. The PR's own proof used two distinct domains, which is why this never showed.

repositories/url_repository.py
```diff
@@ -30,6 +31,34 @@ def dest_host_filter(host: str) -> dict:
...
+                "input": {"$objectToArray": {"$ifNull": ["$geo_rules", {}]}},
+                "as": "rule",
+                "in": "$$rule.v",
+            }
+        },
+    ]
+}
+# (host, registrable) pairs for every destination; secondary_registrable is
+# index-aligned with secondary_hosts by construction (UrlDestination.for_link).
+_HOST_PAIRS = {
```
Zingzy, line 60: **Deploy-order note for the runbook (plus a doc line here):** `$zip` without `useLongestLength` truncates to the shortest input, so a doc carrying `secondary_hosts` but not `secondary_registrable` (anything written by #342-era code, or #342's backfill) yields empty pairs — its secondary hosts are silently invisible to the feed-delta sweep until this branch's backfill re-runs (the `_SECONDARY_FILTER` keying on `secondary_registrable` absent is what makes that re-run work). Since deploys are release-gated, merging #342 and #343 before one release and running the backfill once makes the window vanish; the constraint just needs to be written down, because nothing in the code or the script says it. Also worth knowing: no test executes this pipeline ($zip/$objectToArray/pairs.1 are only ever seen by AsyncMock) — the PR body's live run is currently the only thing pinning it, and a tests/db case would keep it pinned.

scripts/backfill_url_dest.py
```diff
@@ -133,26 +150,33 @@ def backfill_secondary(coll, dry_run: bool) -> None:
...
+                        ).items()
                     }
                 },
             )
             for d in batch
         ]
         try:
-            coll.bulk_write(ops, ordered=False)
+            skipped += len(ops) - coll.bulk_write(ops, ordered=False).matched_count
         except BulkWriteError as exc:
```
Zingzy, line 208: **Nit:** this branch doesn't increment `skipped`, so the "remaining (expect failed + skipped)" reconciliation line lies on any partial bulk failure. The main `backfill()` handles the same case correctly; mirror it.

repositories/url_repository.py
```diff
@@ -30,6 +31,34 @@ def dest_host_filter(host: str) -> dict:
 
 # Every host a link can route to, for grouping: main plus secondary.
 _ALL_HOSTS = {"$setUnion": [["$dest.host"], {"$ifNull": ["$dest.secondary_hosts", []]}]}
+# Every URL the link routes to; the sweeps pick the one on the host they emit.
+_ALL_URLS = {
```
Zingzy, line 37: **Nit, for the A/B tripwire:** this reads `long_url` + `geo_rules` only, which is complete today, but when `ab_variants` gets a write path this is one of the places that must grow a third arm — and the A/B launch-guard checklist doesn't name it. Add it to that tripwire's list now while the knowledge is fresh.

Zingzy (commented):
Reviewed as the stack tip (delta over #342 only). Suite green with GEO_RULES_ENABLED=true (3624 in a clean worktree). The `$match pairs.1` repair of #342's fan-out over-reach and the sample-on-the-emitted-host fix are both exactly right — the second one quietly prevents the analyzer screening the wrong page for every secondary-host sweep hit, which would have poisoned verdicts. The optimistic-concurrency guards the backfill grew (re-assert what was read, count skips) are better than the first version.

One real should-fix (the L1 fanout can self-trigger a burst), one deploy-order note that must make it into the runbook, and nits, inline.

services/url_service.py
Zingzy: Fixed in 739bb58: create now records one L1 event per registrable domain (the counter key), deduping before the loop, so fifty geo paths on one host bump the domain once and the redirect probe runs once per domain. Test: test_many_geo_paths_on_one_domain_count_once (ten same-domain rules, one record_create).

repositories/url_repository.py
Zingzy: Both done in 5cf34d1. A comment on _HOST_PAIRS states the $zip truncation and that the backfill's second pass (keyed on the missing secondary_registrable) repairs pre-field docs; the deploy order (merge #342 and #343 before one release, run the backfill once) is in the PR body. New tests/db/test_secondary_destination_sweeps.py runs the real pipeline against Mongo: hidden hosts under the feed domain come out with their own URL, the clean main host does not, the recent-screen listing carries secondary hosts with their own URL, and a pre-registrable stamp is invisible to the feed sweep while still reachable by host verdicts. _ALL_URLS also gained the pre_start_url from #342.

scripts/backfill_url_dest.py
Zingzy: Fixed in 5cf34d1: on a BulkWriteError both passes add len(ops) - nMatched - len(writeErrors) to skipped, so the reconciliation line holds on partial failures. Test: test_backfill_secondary_partial_bulk_failure_reconciles (one failed, one written, one skipped, remaining 2).

repositories/url_repository.py
Zingzy: Added in 5cf34d1 as item 4 of the A/B tripwire's _WORK: variant URLs must go into _ALL_URLS so the sweeps sample a variant host's own URL.

Zingzy (commented):
Re-reviewed the restack (now two commits on the amended #342). All four findings fixed and verified in the diff:

- The L1 loop dedupes by registrable domain before counting, with the fifty-paths-one-domain rationale in the comment, and now covers pre_start_url too. The one-create-crosses-the-burst-threshold lever is gone.
- The $zip truncation is documented at the pipeline itself, stating exactly which docs yield no pairs and that the backfill's second pass is the cure — the deploy-order constraint now lives where the next reader will trip on it.
- skipped is counted in the BulkWriteError branch of BOTH backfill passes (nMatched-based), so the reconciliation line is honest on partial failures.
- The A/B tripwire checklist gained the _ALL_URLS item. Two more belong on it per the #344 review (variants in the L1 loop, secondary_registrable for variant hosts) — #344's rebase is the natural moment.

Suite green on the tip: 3821 with GEO_RULES_ENABLED=true in a clean worktree. Runbook line for ship day stands: merge #342+#343, one release, run the backfill once.

PR #344, opened 2026-09-03
feat(links): A/B split destinations with per-variant stats

Zingzy:
## Why

The dashboard editor already sends `ab_variants` and the API silently dropped it. This is the backend for link variants, end to end, kept dark behind `AB_VARIANTS_ENABLED=false` until the flag is registered.

## What

Contract, fixed by the frontend: `ab_variants: [{url, weight}] | null`, integer weights summing to at most 100, `long_url` keeps the remainder.

- Schema and DTOs: `AbVariant` on `UrlV2Doc` and the redirect cache, echoed on every link response, hard cap 50 in the DTO and product cap `ab_variants_max` (10) in settings.
- Service: same shape as geo rules. Every variant URL runs the full destination policy on create and patch; identical patches are no-ops; `null` or `[]` clears.
- Gates: `ab_testing` feature flag on create and patch, anonymous creates rejected.
- Redirect: one roll from 0 to 99 per request walks the weight bands in list order; the remainder goes to `long_url`; a matched geo rule wins over the split. Responses carry `Cache-Control: no-store`.
- Click attribution: `variant_index` rides `ClickEvent` through both sinks onto `ClickDoc`.
- Stats: `variant` as a filter and group-by on the account and per-link endpoints. Values are the index as a string and `(default)` for clicks that went to `long_url`; filters coerce indices back to the stored int. `openapi.json` regenerated.
- Edge cache: variant links are never promoted, since the Worker cannot make the pick.
- Safety: variant hosts flow into `dest.secondary_hosts` at create, update and backfill, and into report intake and hot screening, so a host block reaches a link that hides the host in a variant. The public preview API and the legacy preview page list every variant with its share.

## Proof

Live against the compose stack with the flag on:

- A 60/40 link over 200 redirects split 118/82; `ClickDoc` counts and `group_by=variant` agree exactly.
- After a 30/30 patch, 200 more requests produced a `(default)` bucket of 76; filtering by `(default)` or a variant index works.
- The real cached entry is refused by edge promotion with reason `ab_variants`.
- Preview API and legacy preview list both variants; a foreign variant host lands in `dest.secondary_hosts`.
- Anonymous create 401, weights over 100 422, disabled flag 403 once the flag cache expires.
- `uv run pytest`: 3818 passed.

Stacked on `feat/safety-secondary-destinations`.

## Summary by CodeRabbit

* **New Features**
  * Added weighted A/B testing destinations for shortened links.
  * Users can configure variants when creating or updating links, with validation and feature controls.
  * Redirects now distribute visitors among configured destinations and expose variant details in previews.
  * Statistics and exports support filtering and grouping by variant.
  * Webhook link snapshots include configured variants.

* **Bug Fixes**
  * Geo-targeting continues to take precedence over A/B variants.
  * Variant destinations are included in safety checks and destination discovery.

### DIFF
--- a/config.py
+++ b/config.py
@@ -759,7 +759,7 @@ class AppSettings(BaseSettings):
     geo_rules_enabled: bool = False
     geo_rules_max_countries: int = 50
     # Dark like geo_rules until safety stamps secondary hosts; see the
-    # tripwire in tests/unit/test_ab_variants_launch_guard.py.
+    # tripwire in tests/unit/test_ab_testing_launch_guard.py.
     ab_variants_enabled: bool = False
     ab_variants_max: int = 10
     url_password_min_length: int = 8
--- a/config.py
+++ b/config.py
@@ -758,6 +758,10 @@ class AppSettings(BaseSettings):
     # tests/unit/test_geo_launch_guard.py.
     geo_rules_enabled: bool = False
     geo_rules_max_countries: int = 50
+    # Dark like geo_rules until safety stamps secondary hosts; see the
+    # tripwire in tests/unit/test_ab_variants_launch_guard.py.
+    ab_variants_enabled: bool = False
+    ab_variants_max: int = 10
     url_password_min_length: int = 8
     account_password_min_length: int = 8
     account_password_max_length: int = 128
@@ -770,6 +774,7 @@ class AppSettings(BaseSettings):
         "max_emoji_alias_length",
         "emoji_generated_alias_length",
         "geo_rules_max_countries",
+        "ab_variants_max",
         "account_erasure_batch_limit",
         "account_erasure_time_budget_seconds",
         "account_erasure_claim_lease_seconds",
--- a/schemas/enums/stats.py
+++ b/schemas/enums/stats.py
@@ -44,6 +44,8 @@ class StatsDimension(str, Enum):
     # resolves them (by name or id) to the owner's url_ids before the $match.
     TAG = "tag"
     TAG_ID = "tag_id"
+    # A/B variant index served ("0", "1", ...); "(default)" = long_url.
+    VARIANT = "variant"
 
 
 class StatsMetric(str, Enum):
@@ -83,6 +85,7 @@ ALLOWED_FILTERS = frozenset(
         StatsDimension.UTM_CAMPAIGN,
         StatsDimension.TAG,
         StatsDimension.TAG_ID,
+        StatsDimension.VARIANT,
     }
 )
 # Per-link endpoints pre-select the link in the path, so slicing or
--- a/services/report_intake_service.py
+++ b/services/report_intake_service.py
@@ -319,7 +319,8 @@ class ReportIntakeService:
         self, domain: str | None, code: str
     ) -> list[str] | None:
         """Domain-scoped existence check that also yields every destination
-        the link routes to (long_url, geo overrides, pre-start and after-expiry pages).
+        the link routes to (long_url, geo overrides, A/B variants, pre-start
+        and after-expiry pages).
 
         System-domain codes resolve via the shared PublicLinkResolver —
         the same generation dispatch (v1/v2/emoji) as the redirect, and
--- a/services/safety/hot.py
+++ b/services/safety/hot.py
@@ -81,7 +81,7 @@ class HotLinkScreen:
             )
 
     async def _destinations(self, hot: HotUrl) -> list[str]:
-        """Every destination the link routes to: main, geo overrides, both fallbacks."""
+        """Every destination the link routes to: main, geo overrides, variants, both fallbacks."""
         domain = self._system_domain if hot.domain in ("default", "") else hot.domain
         doc = await self._url_repo.find_by_alias(hot.short_code, domain)
         if doc is not None:
--- a/services/url_service.py
+++ b/services/url_service.py
@@ -47,6 +47,7 @@ from repositories.legacy.emoji_url_repository import EmojiUrlRepository
 from repositories.legacy.legacy_url_repository import LegacyUrlRepository
 from repositories.url_repository import UrlRepository
 from schemas.dto.requests.url import (
+    AbVariantRequest,
     ClaimItemRequest,
     CreateUrlRequest,
     ListUrlsQuery,
@@ -77,6 +78,7 @@ if TYPE_CHECKING:
     from services.tag_service import TagService
 from schemas.models.base import ANONYMOUS_OWNER_ID
 from schemas.models.url import (
+    AbVariant,
     EmojiUrlDoc,
     LegacyUrlDoc,
     LinkMetaTags,
@@ -85,6 +87,7 @@ from schemas.models.url import (
     UrlStatus,
     UrlV2Doc,
 )
+from shared.ab_variants import variant_urls
 from shared.alias_dispatch import (
     emoji_lookup_candidates,
     resolution_order,
@@ -140,6 +143,35 @@ def _validate_geo_rules_shape(
             raise ValidationError(
                 f"'{code}' is not a valid ISO 3166-1 alpha-2 country code",
                 field=f"geo_rules.{code}",
+            )
+
+
+def _validate_ab_variants_shape(
+    variants: Sequence[AbVariantRequest],
+    *,
+    max_variants: int,
+    enabled: bool = True,
+) -> None:
+    """Feature gate + entry cap. Weights and URL safety are checked
+    elsewhere (DTO and ``url_policy.check`` respectively)."""
+    if not enabled:
+        raise ValidationError("A/B variants are not available yet", field="ab_variants")
+    if len(variants) > max_variants:
+        raise ValidationError(
+            f"ab_variants cannot exceed {max_variants} entries", field="ab_variants"
+        )
+
+
+async def _check_variant_destinations(
+    service: UrlService, variants: Sequence[AbVariantRequest]
+) -> None:
+    """Every variant URL runs the full L0 gate, same as long_url and geo."""
+    for index, variant in enumerate(variants):
+        rejection = await service._url_policy.check(variant.url)
+        if rejection is not None:
+            log.info("url_create_rejected", reason=rejection.code)
+            raise ValidationError(
+                rejection.public_message, field=f"ab_variants.{index}.url"
             )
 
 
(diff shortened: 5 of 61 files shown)


tests/unit/test_ab_testing_launch_guard.py
```diff
@@ -0,0 +1,40 @@
...
+from __future__ import annotations
+
+import inspect
+
+from config import AppSettings
+from schemas.models.url import UrlDestination
+
+_FLAGS = ("ab_testing_enabled", "ab_variants_enabled")
+
+_WORK = """
```
Zingzy, line 20: **Should-fix: this checklist doesn't know about #343's half of the machinery.** Items 1-3 cover stamping, the event fanout and the backfill — all done in this PR. But #343 (parallel on the same base) adds three more places variant URLs must reach, and whichever branch merges second has to reconcile them: the L1 `record_create` loop (currently `long_url` + geo only — and note my #343 review's dedupe-by-registrable-domain fix applies to variants doubly, since 10 same-domain variants plus geo rules compound the same counter), the `_ALL_URLS` / `_HOST_PAIRS` sweep pipelines, and `secondary_registrable` for variant hosts (this branch's `for_link` predates that field). Add items 4-6 here so the tripwire carries the whole list; the guard is the one place the merge-seam knowledge survives a rebase.

config.py
```diff
@@ -758,6 +758,10 @@ def blocked_self_domains(self) -> tuple[str, ...]:
     # tests/unit/test_geo_launch_guard.py.
     geo_rules_enabled: bool = False
     geo_rules_max_countries: int = 50
+    # Dark like geo_rules until safety stamps secondary hosts; see the
+    # tripwire in tests/unit/test_ab_variants_launch_guard.py.
```
Zingzy: **Nit:** the comment names `test_ab_variants_launch_guard.py`; the file is `test_ab_testing_launch_guard.py`. Also a heads-up rather than a finding: rebasing onto the amended #342 base will conflict in `for_link` and the backfill (the amendment folded `pre_start_url` into the same seams), and it brings the legacy-preview fix that correctly withholds this PR's variant list for scheduled links too — worth confirming after the rebase.

schemas/enums/stats.py
```diff
@@ -40,6 +40,8 @@ class StatsDimension(str, Enum):
     UTM_SOURCE = "utm_source"
     UTM_MEDIUM = "utm_medium"
     UTM_CAMPAIGN = "utm_campaign"
+    # A/B variant index served ("0", "1", ...); "(default)" = long_url.
+    VARIANT = "variant"
```
Zingzy, line 48: **Nit, semantics worth one documented sentence:** the variant dimension is positional. Replacing the list re-keys history — yesterday's index 0 clicks attach to today's index 0 URL. The frontend handles dangling indices from a SHRUNK list gracefully, but a swapped list silently reassigns credit. "Edits re-key past clicks" in the API description (or the docs) makes it a known property instead of a support ticket.

Zingzy (commented):
Mergeable, dark behind both the config kill-switch and the allowlist flag, exactly the geo layering. Full suite green on the branch (3701 in a clean worktree). The hot-path work is right: geo-wins precedence is documented and deterministic, the pick is O(variants) with no allocation, `no-store` correctly extends to variant links, the edge skip has the honest reason (the Worker has no weighted-pick entry type), and `variant_index` rides the click pipeline with backward-compatible defaults at every layer. I also traced the password path: the form re-enters the GET, so variants apply there without special-casing. Both preview surfaces (new endpoint and the legacy template) list every variant with its share inside the active-and-unlocked guard — the anti-cloaking doctrine held on the first try, which past feature PRs needed a review round to get.

The launch guard upgrading from a field-name check to a behavioral stamp assertion follows the geo guard's improvement. One should-fix about what that guard's checklist doesn't yet know, one rebase note, one semantics nit, inline.

Zingzy (commented):
**Verdict: not mergeable as it stands. The code is sound; the rebase is the work.** GitHub reports this branch as conflicting with main, and a test merge onto main confirms it: 15 files collide, including `services/url_service.py`, `services/safety/hot.py`, `services/report_intake_service.py`, `services/public_preview_service.py`, `scripts/backfill_url_dest.py`, `schemas/models/url.py`, `openapi.json` and the launch guard. Those are the files #349, #350 and #352 rewrote after this branch forked. CI is green on the branch itself, but there is no green test-merge to lean on.

Two of the findings below are best fixed *during* the rebase, because main moved the seam they sit on.

**Should fix**

1. **The L1 burst counter never sees variant hosts.** In `create`, the per-domain counter loop still enumerates `long_url`, `geo_rules` and `pre_start_url` only. Fifty links each hiding the same phishing host in a variant never trip the counter. `dest.secondary_hosts` gets the host (so a block reaches it), but the burst detector, which is the thing that finds the campaign in the first place, does not. Detail inline.

2. **Feed variants through `link_destination_urls_for`, not per-reader kwargs.** #350 ended the drift class where every destination reader re-listed the fields: readers now call `link_destination_urls_for(doc)` and that function is the single place that knows which fields hold destinations. This branch predates that and re-adds `variants=variant_urls(doc.ab_variants)` at three call sites, which is exactly the pattern main just deleted. On rebase, add `variants=variant_urls(getattr(link, "ab_variants", None))` inside `link_destination_urls_for` and drop the three kwargs. The L1 loop, hot screening and report intake all pick variants up in one move, and finding 1 disappears with it.

3. **`_ALL_URLS` in `repositories/url_repository.py` needs a variants arm.** This is item 4 of the launch guard's checklist and the one item this branch does not do. The feed-delta sweep samples a URL per host from that pipeline; a host that only appears in a variant yields no URL to check. One arm alongside the geo `$map`:

```python
{"$map": {"input": {"$ifNull": ["$ab_variants", []]}, "as": "v", "in": "$$v.url"}},
```

`$ifNull` covers the cleared state, which stores `null`. Take main's version of the guard text; it already names this.

**Checked and fine**

- Redirect: one roll per request, geo match wins, remainder to `long_url`, `no-store` on split links, edge promotion refused with its own reason. `pick_variant` bands are correct at the boundaries (roll 0 lands in variant 0; roll equal to the summed weights lands on the default).
- Gating mirrors geo exactly: route requires the `ab_testing` flag only when the list is non-empty, so a de-allowlisted owner can still clear; the config kill switch sits under it; anonymous creates 401.
- Update handler: identical list is a no-op, `null` and `[]` clear, every variant URL runs the L0 gate, and the `dest` restamp key set includes `ab_variants`.
- Cache: `UrlCacheData.ab_variants` defaults to None so old entries decode; no version bump needed.
- Stats: `variant` filter coerces indices back to int and `(default)` falls into the null-or-missing branch, so pre-feature clicks count as default. Export and per-link dimension sets include it.
- Preview transparency holds in both directions: variants listed only while active and unlocked, which is the only state in which the redirect serves them.
- Webhook `LinkSnapshot` carries the list; sample payload updated.

**Notes**

- A plain link already promoted to the edge that gains variants through PATCH keeps serving the cached `long_url` redirect until the KV entry expires (about five minutes). The purge condition on update covers deactivation, `starts_at` and key changes only. Same behaviour as a `long_url` edit today, so I am not asking for a change, just naming it.
- `openapi.json` will need regenerating after the rebase.
- `_check_variant_destinations` logs `url_create_rejected` from the update path too. Nit.

services/url_service.py
```diff
@@ -986,6 +1051,7 @@ async def create(
             dest=UrlDestination.for_link(
                 request.long_url,
                 geo_rules=request.geo_rules,
+                variants=variant_urls(request.ab_variants),
```
Zingzy, line 1093: **Should fix.** `for_link` sees the variants here, so `dest.secondary_hosts` is right. But the L1 burst counter loop about fifty lines above this (`for destination in link_destination_urls(request.long_url, geo_rules=..., pre_start_url=...)`) does not, so a burst of links that hide one host in a variant never trips the per-domain counter. On main that loop is now `link_destination_urls_for(request)`; teach that function about `ab_variants` (see the review body) and this is fixed without touching the loop.

services/safety/hot.py
```diff
@@ -77,12 +78,15 @@ async def on_hot(self, hot: HotUrl) -> None:
...
-        """Every destination the link routes to: main, geo overrides, pre-start page."""
+        """Every destination the link routes to: main, geo overrides, variants, pre-start page."""
         domain = self._system_domain if hot.domain in ("default", "") else hot.domain
         doc = await self._url_repo.find_by_alias(hot.short_code, domain)
         if doc is not None:
             return link_destination_urls(
-                doc.long_url, geo_rules=doc.geo_rules, pre_start_url=doc.pre_start_url
+                doc.long_url,
+                geo_rules=doc.geo_rules,
+                variants=variant_urls(doc.ab_variants),
```
Zingzy: **Should fix during the rebase.** Main replaced this whole call with `link_destination_urls_for(doc)` in #350, so this hunk conflicts. Resolve by taking main's one-liner and adding the variants inside `link_destination_urls_for` itself, not here. Same for the twin at `report_intake_service.py`.

services/report_intake_service.py
```diff
@@ -329,18 +331,20 @@ async def _resolve_destinations(
...
             return None
+        return self._v2_destinations(doc)
+
+    @staticmethod
+    def _v2_destinations(doc: UrlV2Doc) -> list[str]:
         return link_destination_urls(
-            doc.long_url, geo_rules=doc.geo_rules, pre_start_url=doc.pre_start_url
+            doc.long_url,
+            geo_rules=doc.geo_rules,
+            variants=variant_urls(doc.ab_variants),
```
Zingzy: Same as `hot.py`: main's version is `link_destination_urls_for(doc)`; put the variants in that function and delete `_v2_destinations`.

tests/unit/test_ab_testing_launch_guard.py
```diff
@@ -2,8 +2,8 @@
 
 Variants add destinations exactly like geo rules do. Safety enforcement
 matches ``dest.host`` or ``dest.secondary_hosts``; a variant host that is
-never stamped there matches nothing. The flag does not exist yet; this
-test names the work the moment someone adds it and flips it on.
+never stamped there matches nothing. AB_VARIANTS_ENABLED keeps the feature
+dark; this test fails the moment it is flipped without the stamping work.
```
Zingzy, line 6: Take main's text here (it already lists `SINGLE_DESTINATION_FIELDS` in item 4). And item 4 itself is still open on this branch: `_ALL_URLS` in `repositories/url_repository.py` has no variants arm. The guard only trips when the config default flips, so CI stays green while the sweep cannot sample a variant-only host. Add the arm in this PR so the checklist is actually complete when the flag flips.

Zingzy:
Re-checked acc2a7c. All three should-fixes are in: `link_destination_urls_for` reads `ab_variants` (duck-typed, which sidesteps the import cycle), `hot.py` and report intake are back to the one-liner main uses, the L1 loop picks variants up through the same call, and `_ALL_URLS` has the variants arm. The launch guard now asserts the seam with a fake link; it still only names `_ALL_URLS` in the checklist text, and an assertion that the pipeline mentions `$ab_variants` would close that too, but it is optional. Nits (update-path event name, config filename, positional-index sentence) all done. Rebase resolved, GitHub reports mergeable, CI green. I ran unit and integration on this head in a clean worktree: 3922 passed. The 17 failures I saw first were my own environment (I had copied a real .env in, which changes the token settings that `test_app_token_scopes` depends on); without it that file passes 17/17. Nothing open from my side.

PR #345, opened 2026-09-03
chore(flags): mongosh seed for the ab_testing feature flag

Zingzy:
## What

`scripts/seed_ab_testing_flag.js`: a mongosh script that registers the `ab_testing` feature flag as an empty allowlist. The flag then exists in prod without enabling anyone; accounts are added with `$addToSet` afterwards.

Insert-only and idempotent: rerunning never rewrites a doc a human has already edited.

## Proof

Run twice against the local compose Mongo: the first run inserts, the second run after adding an allowlist email reports the flag as already registered and leaves the email in place. `FeatureFlagDoc` validates the stored document.

The prod run stays a manual step. Stacked on `feat/ab-variants`.

### DIFF
--- a/scripts/seed_ab_testing_flag.js
+++ b/scripts/seed_ab_testing_flag.js
@@ -1,0 +1,44 @@
+// Registers the ab_testing feature flag without enabling it for anyone.
+//
+// Run against prod from a machine with the Mongo URI:
+//   mongosh "$MONGODB_URI/url-shortener" scripts/seed_ab_testing_flag.js
+//
+// Idempotent: only inserts when no doc exists, so a flag a human has
+// already edited (allowlist, rollout) is never rewritten. Add accounts
+// afterwards with:
+//   db.feature_flags.updateOne({name: "ab_testing"},
+//     {$addToSet: {allowlist_emails: "someone@example.com"}})
+// and unregister with:
+//   db.feature_flags.deleteOne({name: "ab_testing"})
+// The app picks changes up within the flag cache TTL (about 30s).
+
+const now = new Date();
+const result = db.feature_flags.updateOne(
+  { name: "ab_testing" },
+  {
+    $setOnInsert: {
+      name: "ab_testing",
+      enabled: true,
+      rollout_type: "allowlist",
+      allowlist_user_ids: [],
+      allowlist_emails: [],
+      percentage: 0,
+      enabled_digits: [],
+      tier: null,
+      description:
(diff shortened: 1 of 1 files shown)


Zingzy (commented):
The seed script itself is right, and small enough to say so completely: `$setOnInsert` under upsert makes it genuinely insert-only, so a re-run can never rewrite a flag a human has edited; `enabled: true` with `rollout_type: allowlist` and empty lists registers the flag dark (default-deny via the empty allowlist, matching the flag doctrine); the field names match `schemas/models/feature_flag.py` exactly (checked each one); and the header documents add, remove, unregister and the ~30s cache TTL, which is everything an operator needs at 2am. The trailing verification print is a nice touch — the run's output is its own receipt.

**One blocker-shaped chore before merge: restack.** This branch carries stale pre-fix copies of all four #344 commits (different SHAs than #344's current head), which is why the diff shows ~2600 additions for a 44-line script and why spec-drift is red. Rebase onto the current `feat/ab-variants` so the PR is the one commit it claims to be; the seed itself needs no changes.

Runbook note to carry into the ship-day list: run the seed before adding anyone to the allowlist, and after the #344 rebase lands the amended #342 base, the ship-day order is backfill, then seed, then allowlist.

Zingzy (commented):
**Verdict: mergeable** once #344 lands and this retargets to main. Nothing to fix.

**What I verified against main**

- Every field matches `FeatureFlagDoc`: `name`, `enabled`, `rollout_type`, `allowlist_user_ids`, `allowlist_emails`, `percentage`, `enabled_digits`, `tier`, `description`, timestamps. `"allowlist"` is a valid `RolloutType`. The collection name matches `repositories/indexes.py`.
- `$setOnInsert` with upsert is idempotent and never overwrites a flag a human has already edited, as the header claims.
- Behaviour: `is_enabled` already returns False for a missing flag, so registering with an empty allowlist changes nothing for any user. The script exists so an operator can `$addToSet` emails, and that is the right shape.
- `AB_TESTING_FLAG` is `"ab_testing"` in the service and sits in `EXPOSED_FEATURES`, so the dashboard's `/me/features` gate lights up for allowlisted accounts with no further work.

**Optional.** This is the first flag seed in `scripts/`, and `link_scheduling` and `expired_fallback` owe the same one. If you would rather not keep three copies, one `seed_flag.js` that reads the name and description from `--eval 'const FLAG = {...}'` does all three. Not asking for it here.

Zingzy:
Re-checked 42488ec: rebased onto the new #344 head, script unchanged, mergeable once #344 lands.

PR #347, opened 2026-09-03
perf(webhooks): O(1) pending cap via a counter on the endpoint doc

Zingzy:
## What

- Each webhook endpoint document now carries `pending_count`. Dispatch takes a slot with one conditional increment (the same write also bumps `total_deliveries`), and the executor releases the slot on every terminal outcome: success, ladder exhausted, 410, endpoint inactive, secret unreadable, event expired, payload over cap.
- An over-cap verdict is re-verified against one real count before anything is dropped. The counter is overwritten with the real value, and the recount is throttled to once per 30 seconds per endpoint so an endpoint that is genuinely at the cap does not pay a count per event.
- Test sends never touch the counter. Manual retries of an already finished row never release a slot. Reschedules and 429 deferrals keep the slot.
- Endpoint documents created before this change have no counter field; the first over-cap verdict on them recounts and writes the field.

## Why

The dispatcher enforced the per-endpoint pending cap with a `count_documents` per matched endpoint per event. That count grows with the endpoint's backlog and runs on the serial stream consumer, so one backed-up endpoint slowed dispatch for every owner queued behind it. The count-then-insert check was also not atomic, so concurrent dispatchers could overshoot the cap.

## How proven

Rig against local Mongo with the real repositories, matcher, dispatcher and executor: one dead endpoint pre-loaded with a backlog for owner A, one healthy endpoint for owner B, events for A and B dispatched serially in alternation, runs of `main` and this branch interleaved on the same host.

- 20k backlog with the cap raised to 50k, 400 events: on `main` the dead endpoint's events dispatch at 9.4 to 11.5ms p50 and the count alone costs 5 to 6ms. On this branch they dispatch at 5.6 to 5.9ms p50, the same as the healthy endpoint.
- 1000 backlog at the default cap on a quiet host, 2000 events: 3.9s total on this branch against 4.7s on `main`.
- 1000 healthy deliveries drained through the real executor to a local receiver: receiver saw 1000 hits, real pending 0, counter 0.
- Counter forced to 5000 with 900 real pending rows: the next dispatch repaired it to 900 and inserted its row (901 and 901).
- 50 concurrent dispatches with the endpoint at cap minus 5: exactly 5 rows inserted, 45 dropped, counter equal to the real count, zero overshoot.
- Endpoint document with the counter field unset and 1000 real pending rows: first event recounted, wrote 1000, dropped; 100 events later still 1000 real, counter 1000, 100 dropped.
- `uv run pytest`: 3724 passed, 5 skipped.

## Summary by CodeRabbit

- **New Features**
  - Webhook endpoints now enforce pending-delivery limits more reliably, preventing excess queued deliveries.
  - Pending delivery counts are tracked and corrected automatically when needed.
  - Test deliveries no longer consume endpoint pending-capacity slots.
  - Delivery completion and failure handling now release pending capacity accurately, including for non-HTTP and terminal outcomes.

- **Bug Fixes**
  - Improved handling of duplicate or already-completed deliveries to prevent incorrect capacity updates.
  - Pending counters are protected from becoming negative or drifting from actual delivery state.

### DIFF
--- a/services/webhooks/dispatcher.py
+++ b/services/webhooks/dispatcher.py
@@ -10,6 +10,7 @@ multi-hour retry ladder cannot hold a stream entry pending.
 
 from __future__ import annotations
 
+import time
 from datetime import datetime, timezone
 
 from bson import ObjectId
@@ -63,12 +64,15 @@ class WebhookDispatcher:
         endpoint_repo: WebhookEndpointRepository,
         *,
         max_pending_per_endpoint: int = 1000,
+        recount_interval_seconds: float = 30.0,
     ) -> None:
         self._matcher = matcher
         self._event_repo = event_repo
         self._delivery_repo = delivery_repo
         self._endpoint_repo = endpoint_repo
         self._max_pending = max_pending_per_endpoint
+        self._recount_interval = recount_interval_seconds
+        self._recount_not_before: dict[ObjectId, float] = {}
 
     async def dispatch(self, event: DomainEvent) -> None:
         endpoints = await self._matcher.match(event)
@@ -80,14 +84,8 @@ class WebhookDispatcher:
         for endpoint in endpoints:
             # The pending cap protects the queue itself. A subscriber
             # who can't drink max_pending deliveries has already lost the
-            # facts — counting beats pretending. Count-then-insert is
-            # deliberately non-atomic: concurrent dispatchers can overshoot
-            # by a batch, which is fine for a protective backstop; an exact
-            # cap would cost a reservation counter on every dispatch.
-            if (
-                await self._delivery_repo.count_pending(endpoint.id)
-                >= self._max_pending
-            ):
+            # facts — counting beats pretending.
+            if not await self._reserve_slot(endpoint):
                 await self._endpoint_repo.increment_dropped(endpoint.id)
                 log.warning(
                     "webhook_delivery_dropped_over_cap",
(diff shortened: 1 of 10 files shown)


services/webhooks/dispatcher.py
```diff
@@ -98,11 +96,38 @@ async def dispatch(self, event: DomainEvent) -> None:
...
+        while sibling dispatches hold uninserted reservations undercounts
+        by that many; the cap is a backstop, so that overshoot is accepted.
+        """
+        if await self._endpoint_repo.reserve_pending(endpoint.id, self._max_pending):
+            return True
+        now = time.monotonic()
+        if now < self._recount_not_before.get(endpoint.id, 0.0):
+            return False
+        self._recount_not_before[endpoint.id] = now + self._recount_interval
+        counted = await self._delivery_repo.count_pending(endpoint.id)
```
Zingzy, line 122: **Nit.** `count_pending` counts every `status: pending` row, including `is_test` rows. A test send that dies mid-attempt (process killed between `insert_row` and `attempt`) stays pending with `next_attempt_at: None`, is never claimed, and is never released. Each such row inflates every recount by one until the 30-day TTL. Rare, and worst case it costs one slot of cap, but the fix is one clause: exclude `is_test` in `count_pending` (or pass a filter from here) so the repair counts only rows that will ever be released.

Zingzy (commented):
**Verdict: mergeable.** Base of the stack, and the tightest of the three. No blocking findings.

**What I verified**

- Every terminal write goes through `_finish` / `_fail`, so the slot is released exactly once. I inventoried the paths on main: five `mark_failed` sites (endpoint_inactive, secret_unreadable, event_expired, unknown_flavor, payload_over_cap) and four `record_attempt_and_finish` sites (success, single-shot, 410, exhaustion). All nine are rerouted in this diff, and the `ReturnDocument.BEFORE` gate means a row that was already terminal cannot release twice.
- Test sends are symmetric: never reserved by the dispatcher, never released by `_release_slot`.
- Missing-field migration self-heals: a doc with no `pending_count` fails the `$lt` reserve, the throttled recount initialises it, the retry reserves. No backfill needed.
- Drift heals at the only moment it matters. A counter can run high after a TTL-expired pending row, a crash between reserve and insert, or orphan rows after `delete_endpoint`. All three only hurt when the counter reaches the cap, and that is exactly when the recount fires. This is the property that replaces the derived count's self-correction, and it holds.
- `defer` (paused endpoint, 429) keeps rows pending without touching the counter, which matches what `count_pending` reported before.

**Freshness.** The stack branched from an older main. GitHub's test-merge is green and I ran the stack tip against its own base (3627 passed). Nothing here overlaps the safety or scheduling work that landed since.

One nit inline on the recount filter.

services/webhooks/dispatcher.py
Zingzy: Fixed in fa4c937. count_pending now filters is_test with $ne true, so a test send that died between insert_row and attempt no longer inflates the recount. The integration fake mirrors the filter and a repo test pins the query shape.

Zingzy:
Re-checked fa4c937: `count_pending` now excludes `is_test` rows and `test_counts_only_rows_that_hold_a_slot` pins it. CI green including codecov/patch. Nothing open from my side; mergeable.

PR #348, opened 2026-09-03
perf(webhooks): run delivery attempts under a bounded semaphore

Zingzy:
Stacked on #347.

## What

- `DeliveryExecutor.run` claims rows one at a time as before, but each attempt runs as its own task under a semaphore. `WEBHOOKS_EXECUTOR_CONCURRENCY` (default 8) bounds attempts in flight per process.
- `WEBHOOKS_EXECUTOR_PER_ENDPOINT_CONCURRENCY` (default 2) caps how many of those slots one endpoint may hold. Saturated endpoints are excluded from the claim query with `$nin`, so their rows wait while other endpoints' rows keep flowing.
- Per-row atomic claim, lease and outcome writes are unchanged.
- Shutdown still drains: on cancel the loop waits up to one delivery timeout for in-flight attempts, then cancels whatever is left and logs how many were drained or abandoned.
- An attempt that raises logs `webhook_executor_tick_failed` with `stage=attempt` and frees its slot; a claim failure logs the same event with `stage=claim`.

## Why

The claim loop was one serial task. A subscriber endpoint that accepted the connection and then sat on it held the whole executor for the 15s delivery timeout, once per attempt, and every other owner's deliveries queued behind it.

## How proven

Rig against local Mongo with the real dispatcher and executor and a local receiver that answers `/slow` after 14s and `/ok` at once. Slow rows were dispatched first so they sat at the head of the claim order, then 50 healthy rows for a second owner. Four executors ran against the same Mongo.

- `main`, 4 slow rows: the first healthy delivery completed at 14.75s and the last at 15.76s. Every healthy row waited behind the slow ones.
- This branch, 4 slow rows: first healthy at 0.19s, all 50 at 0.51s, all 4 slow rows still pending.
- This branch, 8 slow rows (exactly filling the 4 x 2 per-endpoint slots): first healthy at 0.12s, all 50 at 0.41s, healthy created to completed p50 0.32s and max 0.41s, all 8 slow rows still pending. The slow rows completed at 14.2s.
- Exactly once under 4 claimers: receiver saw 58 hits and 58 distinct webhook ids, 58 rows in SUCCESS, every row with exactly one attempt.
- Shutdown drain: one executor cancelled while its two slow attempts were in flight; `run()` returned after 13.7s, once those attempts had completed, and both rows were delivered once.
- `uv run pytest`: 3730 passed, 5 skipped.

### DIFF
--- a/config.py
+++ b/config.py
@@ -469,6 +469,10 @@ class WebhookSettings(BaseSettings):
     max_pending_per_endpoint: int = Field(default=1000, ge=10)
     executor_poll_seconds: float = Field(default=1.0, gt=0)
     executor_lease_seconds: int = Field(default=60, ge=10)
+    # Attempts in flight per executor process, and how many of those one
+    # endpoint may hold so a hanging receiver cannot occupy every slot.
+    executor_concurrency: int = Field(default=8, ge=1)
+    executor_per_endpoint_concurrency: int = Field(default=2, ge=1)
     # Owner→subscription-count cache in front of the matcher.
     matcher_cache_ttl_seconds: int = Field(default=60, ge=5)
     domain_stream_maxlen: int = Field(default=100_000, ge=1000)
--- a/.env.example
+++ b/.env.example
@@ -145,6 +145,8 @@ GITHUB_REPO="" # org/repo format
 # WEBHOOKS_MAX_CONSECUTIVE_FAILURES="10"         # exhausted deliveries → auto-disable
 # WEBHOOKS_DELIVERY_LOG_TTL_DAYS="30"            # events + delivery log retention
 # WEBHOOKS_MAX_PENDING_PER_ENDPOINT="1000"       # backlog cap; beyond = drop+count
+# WEBHOOKS_EXECUTOR_CONCURRENCY="8"              # delivery attempts in flight per process
+# WEBHOOKS_EXECUTOR_PER_ENDPOINT_CONCURRENCY="2" # of those, max held by one endpoint
 
 # ── URL safety pipeline ──────────────────────────────────────────────────
 # Report-triggered destination analysis, per-host verdicts, and automatic
--- a/dependencies/wiring.py
+++ b/dependencies/wiring.py
@@ -506,6 +506,8 @@ def wire_services(app: FastAPI, settings: AppSettings, redis_client) -> None:
         max_consecutive_failures=wh_settings.max_consecutive_failures,
         poll_interval=wh_settings.executor_poll_seconds,
         lease_seconds=wh_settings.executor_lease_seconds,
+        concurrency=wh_settings.executor_concurrency,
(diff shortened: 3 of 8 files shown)


config.py
```diff
@@ -469,6 +469,10 @@ class WebhookSettings(BaseSettings):
     max_pending_per_endpoint: int = Field(default=1000, ge=10)
     executor_poll_seconds: float = Field(default=1.0, gt=0)
     executor_lease_seconds: int = Field(default=60, ge=10)
+    # Attempts in flight per executor process, and how many of those one
+    # endpoint may hold so a hanging receiver cannot occupy every slot.
+    executor_concurrency: int = Field(default=8, ge=1)
+    executor_per_endpoint_concurrency: int = Field(default=2, ge=1)
```
Zingzy, line 475: **Note, not blocking.** Main's serial loop plus the `next_attempt_at` sort gave subscribers de-facto FIFO per endpoint. Nothing promised it, but at per-endpoint concurrency 2, `link.created` and `link.updated` for the same link can now arrive swapped whenever the first attempt is slower than the second. The payload carries the event timestamp so consumers can order, and at-least-once already required dedup on `webhook_id`. Worth one sentence in the webhooks docs when the flag flips: deliveries are unordered, sort by the event timestamp.

Zingzy (commented):
**Verdict: mergeable** once #347 lands and this retargets to main. No blocking findings.

**What I verified**

- The semaphore lives in `run()`, not `attempt()`. `send_test` and `retry_delivery` call `attempt` inline from a request, so they never queue behind background backlog. Good call, that was the trap here.
- No slot leak on any path: acquire before claim, claim exception leaves `row` None and releases, `_run_attempt` releases in `finally`. The claim-error test pins it.
- Shutdown: drain bounded by the delivery timeout, and a second cancel arriving inside `asyncio.wait` still runs the `finally` that cancels the abandoned tasks. `app.py` cancels the embedded executor with a 10s bound, which is shorter than the 15s drain, so in embedded mode the drain gets cut at 10s. That is fine, just know the log will say abandoned.
- Lease arithmetic still holds: 60s lease against a 15s HTTP timeout with 8 in flight leaves plenty of room. Not a concern at these defaults.
- `run()` is exercised for the first time (six cases: overlap bound, per-endpoint cap, drain, raise frees slot, claim error, abandon past timeout). Main had zero coverage of the loop.

One behaviour note inline, not a blocker.

config.py
Zingzy: Agreed, and it is deliberate: FIFO per endpoint was a side effect of the serial loop, never a contract, and holding it would mean one slot per endpoint, which brings the head-of-line problem back for that endpoint. Consumers already need dedup on webhook_id for at-least-once, and the payload carries the event timestamp for ordering. I will add the "deliveries are unordered, sort by the event timestamp" sentence to the webhooks docs when the flag flips.

PR #349, opened 2026-09-03
feat(links): scheduled start time for links

Zingzy:
## What

Links can now carry a start time. Before that moment the short link is not
live: a visit answers a 404 "not live yet" page, or a 302 to an owner-set
fallback URL when one is given. At the start time the link begins
redirecting on its own. The feature ships dark behind a `link_scheduling`
feature flag (default off) and is authenticated-only, mirroring geo
targeting and custom meta tags. Paid gating comes later.

Two new fields on a link, both optional and clearable:

- `starts_at`: a UTC instant (ISO 8601 or unix seconds), must be in the
  future and before `expire_after`.
- `pre_start_url`: where early visitors land; blank shows the not-yet-live
  page. It runs the same destination safety checks as every other URL.

A future start derives a `SCHEDULED` status the same way a past expiry
derives `EXPIRED`: the stored status never changes, every read folds the
clock in. That keeps the redirect, the list filter, and the public preview
answering from one predicate, so a scheduled link never leaks its
destination early.

## Why

Scheduling a drop, an announcement, or an event link is a common ask, and
none of the big shorteners offer a start time. It pairs naturally with the
expiry we already have.

## How proven

- `uv run pytest` green (3755 passed). New tests cover the derived status
  and its ordering against expiry, the resolve branches (cache hit and
  database, fallback carried, expired and blocked both beat scheduled), the
  redirect responses (404 with its own error code and no-store, 302
  fallback with no-store, HEAD parity, password form not reached before
  start), edge-promotion skip, the list filter arms with a predicate-pin
  test, request validation, the flag gates on create and update, the public
  preview, and the webhook change map.
- `ruff check` and `ruff format --check` clean.
- The start time is compared against the cached timestamp on every resolve,
  so a link with a future start returns the not-yet-live answer from a warm
  cache entry and flips to the redirect at the exact second with no cache
  eviction, and a future-start link is never promoted to the edge.
- `openapi.json` regenerated.

## Summary by CodeRabbit

* **New Features**
  * Added scheduled link support with configurable start times and optional pre-start destinations.
  * Scheduled links now display a `scheduled` status and hide their destination until activation.
  * Added scheduling details to URL responses, previews, lists, and webhook snapshots.
  * Unavailable scheduled links return a no-store 404 or redirect to the configured fallback URL.
  * Scheduling is protected by an available feature flag and required authentication.

* **Bug Fixes**
  * Prevented redirects, previews, and edge caching from serving links before their scheduled start time.


Zingzy (commented):
Merge-safe (dark behind the default-off flag, CI green, 3634 tests pass on the branch in a clean worktree), but two of the findings below are flag-enable gates: both are ways the destination still leaks early, which is the one promise this PR makes.

What's right about the core is worth stating, because it's the part that's hard: the derived status lands in all three places at once (the model property, the Mongo clause with the predicate-pin test, and the cache-shaped hot-path check in both resolve branches), and the per-request clock comparison against the cached timestamp is the only correct shape given the 3600s Redis TTL. I traced the "flips at the exact second with no eviction" claim through the cache layer and it's true, including the strict-greater boundary mirroring `is_time_expired`'s `<=`. Resolve raises before the OG intercept and the password gate, so no card leak and no password-form fishing pre-start. `no-store` on both pre-start answers is ahead of the existing error pages (which have none). Public preview and public stats withhold via the shared active-only rule. The flag gates match geo/meta exactly, including clearing-is-never-gated, and SCHEDULED is unreachable through status PATCH because the DTOs are Literal-locked. v1/emoji are safe by default-None. The Caddy error composition is generic over X-Error-Code, so the new slug composes with no Caddy change (verified).

Inline: two should-fixes (both leaks, both cheap), one should-decide on the abuse posture of `pre_start_url`, one docs-parity nit.

schemas/models/url.py
```diff
@@ -235,6 +243,9 @@ def effective_status(self) -> UrlStatus:
             return UrlStatus.EXPIRED
         if self.max_clicks is not None and self.total_clicks >= self.max_clicks:
             return UrlStatus.EXPIRED
+        starts_at = as_aware_utc(self.starts_at)
+        if starts_at is not None and starts_at > datetime.now(timezone.utc):
+            return UrlStatus.SCHEDULED
```
Zingzy, line 251: **Should-fix (flag-enable gate): the legacy preview page still shows the destination.** `routes/legacy/url_shortener.py` withholds only on BLOCKED (`:500`, `:517`, `:547`); a scheduled v2 link's `long_url` and full geo map render at `/{alias}+`, and that legacy page is the surface actually serving in prod until the Caddy cutover of the new public pages. The new preview and public stats both withhold correctly, which makes this the one surface where "a scheduled link never leaks its destination early" fails. Expired links showing their destination there is a defensible pre-existing choice (it was public once); a scheduled link's destination was never public.

While fixing it, add a scheduled arm to `tests/unit/services/test_blocked_destination_leaks.py` ("no surface reveals it") — that suite is exactly the net that would have caught this.

services/edge_cache/promotion.py
```diff
@@ -57,6 +58,8 @@ def promotion_skip_reason(
         return "block_bots"
     if url.expiration_time:
         return "has_expiration"  # could expire mid-TTL; rare, so skip all
+    if url.is_not_yet_live(time.time()):
+        return "not_yet_live"  # the edge would serve the redirect early
```
Zingzy, line 62: **Should-fix (flag-enable gate): this covers new promotions, but not demotion.** PATCHing `starts_at` onto an already-promoted plain link purges nothing: the plain-link purge in `UrlService.update` (`url_service.py:1240-1249`) fires only on `status == INACTIVE` or a key change, and the og-link `relevant` set (`:1235`) doesn't include `starts_at` either. Redis is invalidated on every update, so origin answers correctly, but the edge keeps serving the real 302 for up to ~6 minutes (300s TTL + jitter) into the link's dark window — on exactly the links hot enough to have been promoted. Add `starts_at` to both conditions; one line each.

services/url_service.py
```diff
@@ -867,6 +916,19 @@ async def create(
                 "expire_after must be in the future", field="expire_after"
             )
 
+        # 4b. Scheduling: same future rule, and the pre-start destination
+        # runs the full gate like every geo destination.
+        if request.starts_at is not None and request.starts_at <= now:
+            raise ValidationError("starts_at must be in the future", field="starts_at")
+        _check_start_before_expiry(request.starts_at, expire_ts)
+        if request.pre_start_url:
+            pre_rejection = await self._url_policy.check(request.pre_start_url)
```
Zingzy, line 939: **Should-decide (before the flag widens): `pre_start_url` is invisible to takedown sweeps.** Only `long_url` is stamped into the sparse `dest.*` indexes; this write-time policy check is the only gate `pre_start_url` ever passes. A far-future `starts_at` plus a `pre_start_url` is functionally a permanent redirect whose destination never appears in the dest indexes, so a feed-delta sweep can't find it when the domain gets listed later — strictly better than `long_url` for evasion, since the clean `long_url` is what gets stamped. Geo destinations share this gap, but they only serve matching countries; the pre-start fallback serves every visitor.

Three honest options: stamp it (a second sparse dest field), teach the sweeps to query it, or document it as an accepted gap while the flag stays allowlist-only. Silence is the only wrong choice given what the safety framework exists to protect.

services/webhooks/payloads.py
```diff
@@ -64,6 +64,8 @@ def link_snapshot(doc: UrlV2Doc) -> dict[str, Any]:
         "block_bots": bool(doc.block_bots),
         "max_clicks": doc.max_clicks,
         "expires_at": doc.expire_after.isoformat() if doc.expire_after else None,
+        "starts_at": doc.starts_at.isoformat() if doc.starts_at else None,
```
Zingzy, line 67: **Nit:** the snapshot dict gained `starts_at`/`pre_start_url` but `LinkSnapshot` and `_sample_link()` in `services/webhooks/registry.py` didn't. `extra="allow"` means nothing breaks, but the registry drives the webhook docs, and the tags PR set the precedent of updating model + sample together. Also fine as-is that `status` stays raw (a scheduled link's webhook says ACTIVE): consistent with the existing raw-status choice, and `starts_at` rides along for consumers to derive.

schemas/models/url.py
Zingzy: Fixed in eab0879. The legacy preview page now withholds a scheduled link the way it withholds a blocked one: the redirect's own 404 with X-Error-Code not_yet_live and no-store, no destination or geo map in the body. Verified against a real scheduled link on a local stack (zero occurrences of the destination in the response) while a live link's preview still renders. The destination-leak suite gained a scheduled arm across the model, the public resolver and the public preview, plus a route test for the page itself.

services/edge_cache/promotion.py
Zingzy: Fixed in eab0879. starts_at in update_ops now purges a plain link's promoted entry, and it joins the og write-through's relevant set. Two tests cover both paths. I deliberately did not carry the expiry TRD's accepted TTL window over here: pushing a start into the future is a deliberate go-dark, and the edge should follow origin at once.

services/webhooks/payloads.py
Zingzy: Fixed in eab0879. LinkSnapshot and _sample_link carry starts_at and pre_start_url, and a new test pins the snapshot's keys to the model so a field cannot ship undocumented again. Status stays raw, as you noted.

services/url_service.py
Zingzy: Not changed yet, this one is a product call. The mechanism that would make the stamp natural, secondary destinations visible to enforcement, is still in #342 and #343, and pre_start_url is the same shape of problem as geo rules there. My recommendation is to accept the gap while the flag stays allowlist-only and add pre_start_url to that stamping when it lands. The alternatives are a second sparse dest field now, or teaching the sweeps to query the field directly. Recorded in the design note; leaving this thread open for the decision.

Zingzy (commented):
Re-reviewed the fix commit. All three fixable findings are closed at their choke points, verified in the diff and with the full suite (3767 passed locally; CI was still running at post time):

- The legacy preview now answers a scheduled link with the redirect's own 404 (`X-Error-Code: not_yet_live`, `Cache-Control: no-store`), placed after the blocked check so the ordering matches everywhere else, and only on the v2 branches where a start time can exist. The leak suite gained the scheduled arm with the "secret.example not in the serialized body" assertion plus a route-level integration test, so the next surface inherits the rule instead of the bug.
- `starts_at` joins both edge conditions: the plain-link purge and the og write-through's relevant set. Purging on a cleared start is harmless (the next hot streak re-promotes), so the unconditional `"starts_at" in update_ops` is the right simplicity.
- Registry snapshot model and sample carry both fields.

Still open, by design: the `pre_start_url` sweep-visibility decision (stamp it, sweep it, or document the accepted gap). Fine to leave past merge; it wants an answer before `link_scheduling` goes beyond the allowlist.

services/url_service.py
Zingzy: Decided: accept the gap while the flag stays allowlist-only, and stamp pre_start_url alongside the geo and variant hosts in the secondary-destinations work once this lands. The fallback only serves before go-live, so that is the whole window to cover.

services/url_service.py
```diff
@@ -1117,6 +1182,12 @@ async def update(
         for handler in FIELD_HANDLERS.values():
             await handler(request, existing, update_ops, self)
 
+        if {"starts_at", "expire_after"} & update_ops.keys():
+            _check_start_before_expiry(
+                update_ops.get("starts_at", existing.starts_at),
+                update_ops.get("expire_after", existing.expire_after),
+            )
```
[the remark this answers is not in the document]
Zingzy: Confirmed and fixed in https://github.com/spoo-me/spoo/pull/356. Both sides are normalised with as_aware_utc, and the tests now use naive stored values so the real path is what is exercised.

infrastructure/cache/url_cache.py
```diff
@@ -52,6 +52,10 @@ class UrlCacheData(BaseModel):
     meta_color: str | None = None
     meta_image_width: int | None = None  # from async image validation
     meta_image_height: int | None = None
+    # Unix timestamp before which the link is not live; None = live now.
+    # Entries cached before this field existed deserialize to None.
+    start_time: int | None = None
```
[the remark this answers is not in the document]
Zingzy: Fixed in https://github.com/spoo-me/spoo/pull/356. The cache projection rounds a fractional start up to the next whole second, so the hot path can never read a link as live before the model does. A test with a .900 start pins it.

services/url_service.py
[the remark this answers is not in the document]
Zingzy: Confirmed and fixed in https://github.com/spoo-me/spoo/pull/356. Both sides are normalised with as_aware_utc, and the tests now use naive stored values so the real path is what is exercised.

infrastructure/cache/url_cache.py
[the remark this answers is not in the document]
Zingzy: Fixed in https://github.com/spoo-me/spoo/pull/356. The cache projection rounds a fractional start up to the next whole second, so the hot path can never read a link as live before the model does. A test with a .900 start pins it.

PR #350, opened 2026-09-03
feat(links): fallback URL served once a link has expired

Zingzy:
## What

Owners can set `expired_redirect_url` on a link. Once the link has expired, by time or by click limit, the redirect answers 302 to that URL with `Cache-Control: no-store` instead of the 410 expired page. Blank keeps today's behaviour. INACTIVE and BLOCKED links never use it.

Dark behind the `expired_fallback` feature flag (a missing flag doc reads as off). Setting the field needs a signed-in account with the flag; clearing it never does. The flag is exposed on `GET /api/v1/me/features` so the frontend can show the field.

## Why

An expired link today dead-ends on our page. Owners running a promo or a limited drop want visitors to land on a page they control: sold out, the next offer, a waitlist. Short.io and Dub both offer this per link.

## How

- New `ExpiredRedirectError(GoneError)`, raised at the three places `resolve` finds a v2 link EXPIRED. Every existing `except GoneError` keeps working; only the redirect route catches the subclass first.
- The value passes the same policy gate as `long_url` (format, self-link, blocklist, feeds) and its host is stamped into `dest.secondary_hosts`, so enforcement, sweeps and report intake see it the way they see a geo host.
- A fallback hit records no click. The owner's limit was reached, and counting would push `total_clicks` past `max_clicks`. The route logs `url_expired_fallback` so the traffic stays countable.
- The edge cache never promotes links with `max_clicks` or `expire_after`, so there is nothing to purge when a link flips.
- Stacked on #342 because the secondary-hosts stamp lives there.

## Proof

- `uv run pytest`: 3756 passed, 5 skipped.
- Live against a throwaway database with the flag seeded:
  - link with `expire_after` three seconds out plus a fallback: 302 to the destination before, 302 to the fallback with `cache-control: no-store` after, from both the DB path and the cache path. Mongo shows `status: EXPIRED` and `dest.secondary_hosts: ["fallback.example"]`.
  - `max_clicks: 1` plus a fallback: click 1 goes to the destination, clicks 2 and 3 go to the fallback, `total_clicks` stays at 1.
  - fallback on a blocklisted host: 400 with `field: expired_redirect_url`. Self-link fallback: 400. Anonymous create with the field: 401. Flag off: 403.
  - PATCH set then clear: the secondary host appears on `dest` and disappears again.
- `openapi.json` regenerated.

## Summary by CodeRabbit

* **New Features**
  * Added optional expired-link fallback URLs for supported authenticated users.
  * Expired links can now redirect visitors to the configured fallback with a temporary redirect.
  * Added feature-flag visibility and controls for this capability.
  * URL responses now display the configured expired-link fallback.

* **Bug Fixes**
  * Expired-link fallback redirects are not cached or counted as clicks.
  * Clearing a fallback URL is supported when updating a link.

* **Tests**
  * Added coverage for validation, permissions, configuration, persistence, and redirect behavior.

### DIFF
--- a/routes/redirect_routes.py
+++ b/routes/redirect_routes.py
@@ -16,6 +16,7 @@ from fastapi.responses import RedirectResponse, Response
 from dependencies import ClickSink, GeoIP, UrlSvc
 from errors import (
     BlockedUrlError,
+    ExpiredRedirectError,
     ForbiddenError,
     GoneError,
     NotFoundError,
@@ -185,6 +186,17 @@ async def redirect_url(
     except BlockedUrlError:
         log.info("url_blocked", short_code=short_code)
         return _error_page(request, "451", "THIS URL HAS BEEN BLOCKED", 451)
+    except ExpiredRedirectError as exc:
+        # Not a click: the owner's limit was reached, so nothing is counted.
+        log.info(
+            "url_expired_fallback",
+            short_code=short_code,
+            fallback_domain=extract_hostname(exc.redirect_url),
+        )
+        resp = RedirectResponse(exc.redirect_url, status_code=302)
+        resp.headers["X-Robots-Tag"] = _NOINDEX_HEADER
+        resp.headers["Cache-Control"] = "no-store"
+        return resp
     except GoneError:
         log.info("url_gone", short_code=short_code)
         return _error_page(request, "410", "SHORT URL EXPIRED", 410)
--- a/services/url_service.py
+++ b/services/url_service.py
@@ -99,7 +99,11 @@ from shared.emoji_policy import (
 )
 from shared.generators import generate_emoji_alias_v2, generate_short_code_v2
 from shared.reserved_aliases import is_reserved_alias
-from shared.url_utils import extract_hostname, link_destination_urls, parse_destination
+from shared.url_utils import (
+    extract_hostname,
+    link_destination_urls_for,
+    parse_destination,
+)
 from shared.validators import (
     validate_alias,
     validate_blocked_url,
@@ -1065,11 +1069,7 @@ class UrlService:
         # One record per registrable domain: the counters key on the domain,
         # so fifty geo paths on one host must not read as a fifty-link burst.
         counted: set[str] = set()
-        for destination in link_destination_urls(
-            request.long_url,
-            geo_rules=request.geo_rules,
-            pre_start_url=request.pre_start_url or None,
-        ):
+        for destination in link_destination_urls_for(request):
             parts = parse_destination(destination)
             domain = parts["registrable_domain"] if parts else destination
             if domain in counted:
--- a/services/url_service.py
+++ b/services/url_service.py
@@ -32,6 +32,7 @@ from errors import (
     AppError,
     BlockedUrlError,
     ConflictError,
+    ExpiredRedirectError,
     ForbiddenError,
     GoneError,
     NotFoundError,
@@ -333,6 +334,21 @@ async def _handle_status(
         ops["status"] = request.status
 
 
+async def _handle_expired_redirect_url(
+    request: UpdateUrlRequest, existing: UrlV2Doc, ops: dict, service: UrlService
+) -> None:
+    if "expired_redirect_url" not in request.model_fields_set:
+        return
+    if not request.expired_redirect_url:
+        if existing.expired_redirect_url:
+            ops["expired_redirect_url"] = None
+        return
+    if request.expired_redirect_url == existing.expired_redirect_url:
+        return
+    await service.check_expired_redirect_url(request.expired_redirect_url)
+    ops["expired_redirect_url"] = request.expired_redirect_url
+
+
 async def _handle_geo_rules(
     request: UpdateUrlRequest, existing: UrlV2Doc, ops: dict, service: UrlService
 ) -> None:
@@ -435,6 +451,7 @@ FIELD_HANDLERS: dict[str, Callable[..., Awaitable[None]]] = {
     "status": _handle_status,
     "geo_rules": _handle_geo_rules,
     "tag_ids": _handle_tag_ids,
+    "expired_redirect_url": _handle_expired_redirect_url,
     # Must follow long_url — the handler validates against ops["long_url"]
     # when the destination changes in the same request.
     "meta_tags": _handle_meta_tags,
@@ -592,7 +609,7 @@ class UrlService:
                     schema=schema,
                     source="cache",
                 )
-                _raise_for_status(cached.url_status)
+                _raise_for_status(cached)
             await self._raise_if_time_expired(cached, schema, short_code, "cache")
             _raise_if_not_yet_live(cached, short_code, "cache")
             if should_sample("cache_operation"):
@@ -633,7 +650,7 @@ class UrlService:
                 schema=schema,
                 source="db",
             )
-            _raise_for_status(url_cache_data.url_status)
+            _raise_for_status(url_cache_data)
 
         # 4b. Raise for v1 URLs whose max-clicks have been exhausted
         if (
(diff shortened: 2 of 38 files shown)


shared/url_utils.py
```diff
@@ -95,17 +95,20 @@ def link_destination_urls(
     geo_rules: dict[str, str] | None = None,
     variants: Iterable[str] | None = None,
     pre_start_url: str | None = None,
+    expired_redirect_url: str | None = None,
```
Zingzy, line 119: **Should-fix: the fallback reaches the stamp but not the events.** On this branch (and on the test-merged result), `expired_redirect_url` flows through `for_link` — so blocks, unblocks and sweep MATCHING see the host — but five sites that enumerate destinations by keyword argument don't pass it: the L1 create loop (`url_service.py:1068`), report intake `_resolve_destinations` (both arms), hot-screen `_destinations` (`hot.py:84`), the `_ALL_URLS` pipeline (so a fallback-host sweep hit samples the WRONG url via `_host_samples`' fallback — the exact bug #343 fixed for geo), and the backfill's `secondary_urls` + `_SECONDARY_FILTER` arm. The PR body's "enforcement, sweeps and report intake see it the way they see a geo host" currently holds only for the stamp.

This is the third feature (pre_start, variants, now this) paying the same 6-seam enumeration tax, and the kwargs design of this function is the drift generator: every caller must re-list every field. Root fix: give this module a doc-shaped entry point (`link_destination_urls_for(doc_or_request)`) that owns the field list once, and pin `_ALL_URLS` equivalent to it with a fixture test, the way `effective_status_clause` is pinned to `effective_status`. Then the next destination-bearing field is a one-place change, and the A/B guard's item-4 parenthetical ("today: long_url, geo rules, pre_start_url") stops going stale per feature.

services/url_service.py
```diff
@@ -1835,9 +1876,11 @@ def _validate_and_canonicalize_custom_alias(self, alias: str) -> str:
 # ── Module-level helpers ──────────────────────────────────────────────────────
 
 
-def _raise_for_status(status: UrlStatus) -> None:
-    if status == UrlStatus.BLOCKED:
+def _raise_for_status(data: UrlCacheData) -> None:
+    if data.url_status == UrlStatus.BLOCKED:
         raise BlockedUrlError("URL is blocked")
+    if data.url_status == UrlStatus.EXPIRED and data.expired_redirect_url:
```
Zingzy, line 1882: **Should-fix (transparency): the preview now shows less than the redirect serves.** An expired-with-fallback link 302s every visitor somewhere, while all three public surfaces — the preview endpoint, the legacy `/{code}+` page, public stats — show "expired" with the destination withheld. That inverts the doctrine (never reveal MORE than the redirect serves) into a cloak: a `max_clicks: 1` link becomes, after one click, a permanent redirector whose target no public surface will name. The write-time gate and the dest stamp mean SAFETY sees it, but the human checking `/{code}+` before clicking does not, and that page is the anti-phishing measure spoo points reporters at.

Fix: when status is expired and a fallback exists, the preview surfaces should present the fallback as the live destination, labeled as the post-expiry one ("this link has ended; visitors are sent to X"). That keeps the invariant bidirectional: the preview shows exactly what the redirect does.

routes/redirect_routes.py
```diff
@@ -185,6 +186,17 @@ async def redirect_url(
     except BlockedUrlError:
         log.info("url_blocked", short_code=short_code)
         return _error_page(request, "451", "THIS URL HAS BEEN BLOCKED", 451)
+    except ExpiredRedirectError as exc:
```
Zingzy, line 189: **Fine as designed, one ack:** uncounted fallback hits keep `max_clicks` honest and the log line keeps the traffic measurable in Axiom, but the owner has no in-product signal that their ended link still gets traffic. If fallback adoption is real, a separate counter (not clicks) is the future answer; not this PR's problem.

Zingzy (commented):
Mergeable cleanly (I test-merged against current main and inspected the result; the PR's diff view is misleading because the merge-base predates the #342/#343 amendments, but the branch content is current — my first read assumed staleness and was wrong). CI green, suite verified. Two should-fixes, both inline: one systemic seam gap, one transparency hole that matters given the CF-flag history.

What's right: `ExpiredRedirectError(GoneError)` is the correct shape — every existing `except GoneError` degrades safely (the password path answers its generic 400, check verified), and only the redirect route learns the subclass. The fallback runs the full destination gate, `for_link` stamps its host into BOTH `secondary_hosts` and `secondary_registrable` (verified on the merged tree), `no-store` + noindex on the 302, `max_clicks` stays `gt=0` so there is no zero-click instant redirector, INACTIVE/BLOCKED never serve it, and the flag layering matches geo/meta exactly including ungated clears. Not counting fallback hits is the right call for `max_clicks` integrity and the log keeps the traffic observable; worth remembering the owner has no in-product view of that traffic.

The live proof section (before/after the flip from both cache and DB paths, the click-cap sequence, the gate rejections) is the standard.

shared/url_utils.py
Zingzy: Fixed in b401ffd. shared.url_utils now owns one field list, SINGLE_DESTINATION_FIELDS, plus a doc-shaped link_destination_urls_for(doc_or_request). The L1 create counter, report intake (both arms) and hot screening call that instead of listing kwargs; _ALL_URLS builds its arms from the same tuple, and the backfill's filter, projection and secondary_urls read a local mirror of it that a test pins to the shared one. Pinned by tests: the repo pipeline asserts one arm per field in order, hot and report intake each get a fallback case, the L1 test asserts record_create sees the fallback, and the A/B guard text points at the tuple instead of naming fields. for_link keeps kwargs because create and update build the doc from request plus update_ops, so there is no doc to hand it yet.

services/url_service.py
Zingzy: Fixed in b401ffd. All three surfaces now name the fallback when the link is expired: the preview API carries expired_destination (split like destination), the legacy /{code}+ page renders a "Visitors now go to" block above the original destination, and public stats carry expired_redirect_url. Set only while status is expired, which is exactly when the redirect serves it; active links keep it withheld. The redirect decides before the password gate, so the preview shows it for password-protected links too. spoo-me/frontend#74 renders the matching row on the preview and stats pages.

routes/redirect_routes.py
Zingzy: Agreed, leaving as is. The route logs url_expired_fallback with the short code and fallback host, so the traffic is countable in Axiom today; an owner-facing counter is a separate outcome, not a click, and can come with fallback adoption.

Zingzy (commented):
Re-reviewed the rebase + fix commit. Both should-fixes closed, and the structural fix landed in full rather than as five patches:

- `SINGLE_DESTINATION_FIELDS` is now the one field list; `link_destination_urls_for(doc)` is the doc-shaped reader; the Mongo `_ALL_URLS` pipeline derives its arms from the same tuple; the backfill mirrors it with its own pinned copy; and all four keyword-enumerating readers (L1 counter, both report-intake arms, hot screening) now take the doc. The next destination-bearing field is a one-tuple change, which is exactly the shape that ends this class of drift. The new tests/db pipeline test pins the aggregation against a real Mongo, closing the AsyncMock blind spot from the #343 review too.
- The preview rule is now bidirectional and says so in the docstring: `expired_destination` set only while expired-with-fallback, on the preview endpoint, the legacy page and public stats. The one-click-cap cloak is gone: every public surface names where an ended link still sends people.

Suite green on the tip (3854 in a clean worktree); CI green. Mergeable. The ship-day note stands: seed the `expired_fallback` flag doc before allowlisting anyone, and this feature inherits the already-run dest backfill, so no new migration step.

PR #351, opened 2026-09-03
feat(webhooks): lag, pending depth and latency telemetry with dashboard

Zingzy:
Stacked on #348.

## What

New structured log events and fields, all from code that already runs in the click worker:

- `stream_group_stats`: one flat line per consumer group per stream (`stream`, `group`, `pending`, `lag`), emitted by the existing stream reporter next to its nested `click_stream_stats` line. The worker now also runs a reporter for the domain stream when webhooks are on, so both webhooks groups report lag.
- `webhook_pending_depth` and `webhook_endpoint_depth`: a depth reporter reads the pending counter from the endpoint documents every stats interval and logs the total, the worst endpoint, and one line per backlogged endpoint (top 20).
- `webhook_delivered` gains `latency_ms`, the time from the delivery row being created to the delivery completing. The existing `duration_ms` stays as the HTTP round trip.
- `webhook_delivery_failed`: every terminal failure now logs one line with a `reason` (`exhausted`, `gone`, `endpoint_inactive`, `event_expired`, `secret_unreadable`, `payload_over_cap`, or the single-shot error). Until now only per-attempt failures were logged, so a delivery that gave up was invisible.

Dashboard and monitors as code:

- `axiom/dashboards/spoo-webhooks.json` extended from the draft to 20 charts: consumer lag by stream, pending depth over time and a backlogged-endpoints table, created-to-completed latency percentiles, outcomes by class, plus the existing dispatch, attempt, disable and degradation panels. Deployed to Axiom as "spoo · webhooks".
- `axiom/monitors/webhooks-consumer-lag.json`: max lag on the webhooks group at or above 5000 over 10 minutes.
- `axiom/monitors/webhook-endpoints-disabled.json`: 3 or more endpoints auto-disabled in 15 minutes.
- Both monitors are live and notify the discord-ops notifier.

## Why

The flag stays off for real users until we can see the webhooks consumer group falling behind, subscribers backing up, how long a delivery takes end to end, and how deliveries end. None of that was visible from the logs the worker emitted before this change.

## How proven

Rig with the real dispatcher, executor, stream reporter and depth reporter against local Mongo and real Redis streams (consumer groups created and partly consumed so lag was real), a local receiver with instant, slow, failing and gone paths, JSON logs captured and ingested into a throwaway `spoo-rig` dataset (574 events, 0 failures). Every new panel query was then run against that dataset:

- lag by stream: 200 on the click stream, 15 on the domain stream, matching what the rig left unread
- pending depth: 120 total, worst endpoint 50, and the backlogged-endpoints table listed the five endpoints with their counts
- delivery latency percentiles over 56 deliveries: p50 1173ms, p95 4176ms, p99 6245ms
- outcomes by class: 56 delivered, 170 dropped over cap, 8 failed:endpoint_inactive, 4 failed:exhausted, 2 failed:gone
- endpoints disabled by reason: 2 gone, 2 consecutive_failures

Alert path: a temporary copy of the lag monitor pointed at `spoo-rig` with threshold 100 went to Triggered on the rig's lag of 200 within its first interval and posted to the ops notifier. It has been deleted.

Note: Axiom refuses a monitor whose query references a field the dataset has never seen, and prod has not logged `lag` yet, so one marked row (`schema_seed: true`, lag 0) was ingested into `spoo-prod` to create the lag monitor ahead of the deploy. The lag panels will show real data once the worker with this change is running.

`uv run pytest`: 3737 passed, 5 skipped.

### DIFF
--- a/workers/telemetry.py
+++ b/workers/telemetry.py
@@ -6,6 +6,9 @@ to Axiom via the existing container-logs → Vector path, and the thing to
 alert on (lag growing = worker falling behind; alert BEFORE the buffer
 fills and the sink starts falling back inline).
 
+WebhookDepthReporter emits per-endpoint pending depth from the counter the
+dispatcher maintains — the queue-side view of a subscriber falling behind.
+
 StaleConsumerJanitor deletes consumer names that are long-dead with
 nothing pending — every worker restart registers fresh
 ``{group}-{host}-{pid}`` names and Redis keeps the old ones forever,
@@ -16,9 +19,12 @@ messages would orphan its PEL entries, so only pending==0 names go.
 from __future__ import annotations
 
 import asyncio
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from infrastructure.logging import get_logger
+
+if TYPE_CHECKING:
+    from repositories.webhook_endpoint_repository import WebhookEndpointRepository
 
 log = get_logger(__name__)
 
@@ -65,6 +71,16 @@ class StreamMetricsReporter:
                 for g in groups
             ],
         )
+        # One flat line per group so lag can be charted and alerted on by
+        # group without unpacking the nested list above.
+        for g in groups:
+            log.info(
+                "stream_group_stats",
+                stream=self._stream,
+                group=g.get("name"),
+                pending=g.get("pending"),
+                lag=g.get("lag"),
+            )
 
 
 class StaleConsumerJanitor:
@@ -112,3 +128,47 @@ class StaleConsumerJanitor:
         if removed:
             log.info("stale_consumers_removed", stream=self._stream, removed=removed)
         return removed
+
+
+class WebhookDepthReporter:
+    def __init__(
+        self,
+        endpoint_repo: WebhookEndpointRepository,
+        interval_seconds: float,
+        top: int = 20,
+    ) -> None:
+        self._endpoints = endpoint_repo
+        self._interval = interval_seconds
+        self._top = top
+
+    async def run_forever(self) -> None:
+        while True:
+            try:
+                await self.report_once()
+            except asyncio.CancelledError:
+                raise
+            except Exception as exc:
+                log.warning(
+                    "webhook_pending_depth_failed",
+                    error=str(exc),
+                    error_type=type(exc).__name__,
+                )
+            await asyncio.sleep(self._interval)
+
+    async def report_once(self) -> None:
+        endpoints, total = await self._endpoints.backlog_totals()
+        backlogged = await self._endpoints.find_backlogged(limit=self._top)
+        log.info(
+            "webhook_pending_depth",
+            backlogged_endpoints=endpoints,
+            total_pending=total,
+            max_pending=backlogged[0].pending_count if backlogged else 0,
+        )
+        for ep in backlogged:
+            log.info(
+                "webhook_endpoint_depth",
+                endpoint_id=str(ep.id),
+                user_id=str(ep.user_id),
+                pending=ep.pending_count,
+                status=ep.status.value,
+            )
--- a/axiom/dashboards/spoo-webhooks.json
+++ b/axiom/dashboards/spoo-webhooks.json
@@ -1,8 +1,10 @@
 {
-  "name": "spoo · webhooks",
-  "description": "Webhook dispatch, delivery health, and endpoint lifecycle. Refresh 60s.",
+  "name": "spoo \u00b7 webhooks",
+  "description": "Webhook dispatch, consumer lag, pending depth, delivery latency and outcomes, endpoint lifecycle. Refresh 60s.",
   "owner": "X-AXIOM-EVERYONE",
-  "datasets": ["spoo-prod"],
+  "datasets": [
+    "spoo-prod"
+  ],
   "refreshTime": 60,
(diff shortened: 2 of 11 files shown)


workers/telemetry.py
```diff
@@ -112,3 +125,43 @@ async def sweep_once(self) -> int:
...
+                    error_type=type(exc).__name__,
+                )
+            await asyncio.sleep(self._interval)
+
+    async def report_once(self) -> None:
+        backlogged = await self._endpoints.find_backlogged(limit=self._top)
+        log.info(
+            "webhook_pending_depth",
+            backlogged_endpoints=len(backlogged),
+            total_pending=sum(ep.pending_count for ep in backlogged),
```
Zingzy: **Should fix.** This sums only the top 20 endpoints, and `backlogged_endpoints` above it is `len()` of the same truncated list. The dashboard tile reads this field as the total, and the time series charts `max(total_pending)` beside `worst_endpoint`. Under-reports the moment a 21st endpoint has a backlog, which is exactly the situation the tile exists for.

One aggregate next to the top list gives the honest numbers:

```python
[{"$match": {"pending_count": {"$gt": 0}}},
 {"$group": {"_id": None, "total": {"$sum": "$pending_count"}, "endpoints": {"$sum": 1}}}]
```

Keep `find_backlogged(limit=20)` for the per-endpoint lines and `max_pending`, which is correct as the top-1. The test at `test_summary_and_per_endpoint_lines` only uses two endpoints, so it cannot catch this; add a case with `top=1` and two endpoints where the summary still reports both.

workers/telemetry.py
```diff
@@ -112,3 +125,43 @@ async def sweep_once(self) -> int:
         if removed:
             log.info("stale_consumers_removed", stream=self._stream, removed=removed)
         return removed
+
+
+class WebhookDepthReporter:
+    def __init__(
+        self, endpoint_repo: Any, interval_seconds: float, top: int = 20
```
Zingzy: **Nit.** `Any` here when the two callers both pass a `WebhookEndpointRepository`. Import it under `TYPE_CHECKING` (the module already sits in `workers/`, no cycle) or type a one-method Protocol with `find_backlogged`. `StreamMetricsReporter` above gets away with an untyped redis because the client type is awkward; this one is not.

Zingzy (commented):
**Verdict: one should-fix, then mergeable** after #348 lands and this retargets to main.

**Should fix**

- `total_pending` and `backlogged_endpoints` are computed over `find_backlogged(limit=20)`, so both are capped at the top 20 endpoints. The stat tile is named "Pending deliveries queued (latest)" and the lag monitor's companion chart reads `max(total_pending)`. With 21+ backlogged endpoints the number is silently low. Detail inline.

**Checked and fine**

- APL null semantics. I probed prod: on `click_recorded` rows, which never carry `reason`, `countif(reason != "zzz")` equals `count()` (2294 of 2294), so rows lacking a field pass `!=`. The `is_test != true` filter on ts-outcomes-by-class keeps `dropped_over_cap` and `rate_limited` rows. My first read assumed Kusto semantics and was wrong.
- Same probe: `is_test` is not in the spoo-prod schema yet (`invalid field: "is_test"`), because nothing has logged it. Every chart filtering on it errors until the first `webhook_delivered` lands in prod. Self-heals on first delivery, so nothing to change, but do not read a red chart on import day as a bug.
- `latency_ms` from `created_at` on delivered and failed lines is the right end-to-end number; `duration_ms` stays the HTTP round trip.
- `_log_failed` fires outside the transition gate, so a manual retry that fails again logs a second `webhook_delivery_failed`. That is a real second failure, so I read it as correct, not a double count.
- Monitors: lag >= 5000 over 10m on the domain stream, and >= 3 endpoints disabled in 15m. Both sane for a dark feature.

**Nits inline** (typing on the reporter). One more: the depth reporter is only wired in the click worker. Embedded runtime (`WEBHOOKS_RUNTIME=embedded`) gets no `webhook_pending_depth` line and the tile stays empty. Fine if prod is always worker mode, but worth knowing.

workers/telemetry.py
Zingzy: Fixed in c38b3db. New backlog_totals() runs the $match + $group aggregate over every endpoint with pending_count > 0 and returns (endpoints, total). The summary line takes backlogged_endpoints and total_pending from it; find_backlogged(limit=20) still feeds the per-endpoint lines and max_pending is the first row of that sorted list. Added the top=1 with two endpoints test: the summary reports both while only one per-endpoint line is emitted, plus two repo tests for the pipeline shape and the empty case. On the embedded runtime note: yes, the depth reporter is wired in the click worker only. Prod runs worker mode, so I left the embedded path alone for now.

workers/telemetry.py
Zingzy: Done in c38b3db: WebhookEndpointRepository imported under TYPE_CHECKING and used as the parameter type.

Zingzy:
Re-checked c38b3db: `backlog_totals()` aggregates over every backlogged endpoint, the summary reads it, `max_pending` comes off the sorted top list, and the reporter is typed under `TYPE_CHECKING`. `test_summary_counts_beyond_the_top_list` covers the case the old test could not. CI green. Nothing open from my side; mergeable once #348 lands.

PR #352, opened 2026-09-03
feat(links): tags as a per-account registry with ids, colours and icons

Zingzy:
## What

Tags for links, as a per-account registry. Free for everyone, no flag.

- New `tags` collection: one document per tag with `name` (lowercased, trimmed, letters/digits/spaces/`-`/`_`/`.` only, 32 chars, unique per account), `color` (one of nine palette keys, the least-used one is picked when omitted) and a required `icon` from a curated set of 87 lucide keys, typed as an enum on the wire so `openapi.json` carries the full list (`tag` when omitted; `null` is a 422). At most 500 tags per account.
- `GET /api/v1/tags`, `POST /api/v1/tags`, `PATCH /api/v1/tags/{tag_id}` (rename, recolour, change the icon), `DELETE /api/v1/tags/{tag_id}` (also strips the tag from every link and reports how many).
- Links store `tag_ids`. Create and PATCH take `tag_ids` (whole-list replace; every id must be yours, 400 otherwise; anonymous creates cannot tag). Every link response embeds `tags: [{id, name, color, icon}]` in the link's order, joined with one lookup per request.
- Links list filter: `tagIds` and/or `tagNames` inside the `filter` JSON, with `tagsMatch: any | all`. Names resolve through the registry; unknown names match nothing.
- Stats and export: `tag_id` and `tag` (names) filters, comma-separated or in the `filters` JSON. Clicks never carry tags, so the service resolves them to the owner's link ids and matches on `meta.url_id`, intersecting with an explicit `url_id`. A tag nobody uses matches nothing. Filter only: `group_by=tag` is rejected because grouping clicks by a link property would need a per-click join.
- `POST /api/v1/urls/bulk/tags` with `{ids, add, remove}` of tag ids. Per item the result is current tags minus `remove` plus `add`; unchanged items are no-ops, an item that would exceed 10 tags reports `validation_error`, an `add` id you do not own rejects the whole request. One pipeline update per request (`$filter` then `$concatArrays`), so there is no window where a link has lost its old tags but not gained the new ones.
- Rate limits: 30 per minute on tag create and update, 10 per minute on tag delete, the same budget as bulk delete. Listing stays on the normal authenticated limit.
- Indexes: `(owner_id, name)` unique and `(owner_id, created_at)` on `tags`, `(owner_id, tag_ids)` on `urlsV2`. Webhook `link.*` snapshots carry `tag_ids`. Account erasure deletes the owner's tags.

Tags are not on the redirect cache projection: nothing on the hot path reads them.

## Why

Anyone with more than a handful of links needs a way to group them, filter the table, retag a selection and scope analytics to a campaign. A registry with ids (rather than plain strings on the link) makes a tag a real object: rename or recolour it once and every link follows, and the dashboard can offer a pick-from-list instead of free typing.

## How proven

- `uv run pytest` with the real-Mongo `tests/db` suite enabled: 3836 passed, 0 skipped. New tests cover the tag model and DTOs, the tag service (create with auto colour, cap, conflicts, rename, icon change, delete cascade), the repository helpers, link create/patch validation of ids, the list filter by ids and names, the stats resolution (id arm, names through the registry, empty sentinel, url_id intersection), the bulk op, the erasure cascade and every route.
- `ruff check` and `ruff format --check` clean. `openapi.json` regenerated.
- Live run against a local Mongo 8 with the app on uvicorn: created tags (auto and explicit colours, icon, 409 on a duplicate, 422 on an unknown icon), created 10 links with two tags plus one with another tag, PATCH replace and clear, list filter by ids and by names with any/all (10, 11, 0, 10, 0), tag list counts, 7 real redirects then `tag_id` and `tag` scopes agreeing (6, 6, 1, 0), the `url_id` intersection, export with a tag scope, bulk add/remove on 10 ids with a no-op re-send and an envelope 400 for a foreign id, rename plus recolour plus icon, a 409 on renaming onto an existing name, a 422 for `icon: null` and `tag` as the default icon, and a delete that reported 11 links updated and left the link that only had that tag untagged.

## Summary by CodeRabbit

* **New Features**
  * Added tag management with create, update, list, and delete operations.
  * Added tag assignment to links, including bulk add and remove actions.
  * Added tag details to link responses and webhook events.
  * Added filtering for links, statistics, and exports by tag name or ID.
  * Added tag validation, supported colors and icons, link counts, and automatic cleanup when tags are deleted.


Zingzy (commented):
One blocker (the red CI, a one-line stale assertion), three should-fixes, two should-decides, three nits, all inline. Everything below was verified against the branch, not assumed: I read the full non-test diff, traced the ownership queries, checked the cache projection and edge contract for tag leakage (clean), confirmed the registration checklist end to end (wiring, dependencies, router, indexes, all three smoke fixtures, openapi + its smoke pin), and confirmed the CI failure is exactly one assertion.

The design is right. A registry with ids instead of strings on links makes rename-once real and the dashboard pick-from-list possible; tags stay off the redirect hot path (verified: no tag_ids in UrlCacheData, the edge contract, or any redirect projection); ownership is enforced in the Mongo query at every single touchpoint; name normalisation is single-homed in shared/tags.py and handles combining marks, which most tag implementations get wrong; the empty tag scope resolving to an impossible ObjectId instead of a dropped $in is exactly the fail-closed instinct the stats path needed. The live-run evidence in the PR body is the standard every feature PR should meet.

Merge safety: merging ships nothing, but note this feature is deliberately unflagged, so the next release makes it live for every user at once. Anonymous creates are correctly refused. A frontend deployed ahead of the backend release degrades to an empty Tags page with toasted errors on create, which is acceptable transitional behavior.

services/account_erasure_service.py
```diff
@@ -210,6 +213,7 @@ async def erase(self, user_id: ObjectId) -> dict[str, int]:
         counts["custom_domains"] = await self._domain_service.delete_all_for_owner(
             user_id
         )
+        counts["tags"] = await self._tag_service.delete_all_for_owner(user_id)
```
Zingzy, line 216: **Blocker (this is the whole CI failure):** the counts dict now carries `tags`, and `tests/db/test_erasure_cascade.py:187` (`test_erase_retains_blocked_links_through_domain_cascade`) still asserts the old exact-equality dict without it. The unit-suite `EXPECTED_COUNTS` was updated but the Mongo-backed db suite was missed, which is also why the local run looked green: the db tests were among the 5 skipped. One stale assertion, verified to be the only one in the repo. Consider having the fixture create a tag so the expectation is `"tags": 1` and the cascade is actually exercised there too.

repositories/url_repository.py
```diff
@@ -386,6 +386,83 @@ async def update_by_ids_and_owner(
...
+            ids = [d["_id"] async for d in cursor]
+        except PyMongoError as exc:
+            log.error(
+                "repo_list_ids_by_tags_failed",
+                collection=self._collection_name,
+                error=str(exc),
+            )
+            raise
+        if len(ids) >= limit:
+            log.warning("tag_ids_truncated", owner_id=str(owner_id), limit=limit)
```
Zingzy: **Should-fix:** this cap fails quietly-wrong. The stats path treats the truncated list as the complete scope, so an account with a tag on more than 5000 links gets silently undercounted click totals with a 200 and only a server-side log to say so. Every other cap in this PR either rejects (500 tags, 10 per link) or fails closed (the `_NO_URL_ID` sentinel); this is the one that returns wrong data. Either raise/paginate, or propagate "truncated" to the caller so the response can say the scope was clipped.

services/tag_service.py
```diff
@@ -0,0 +1,152 @@
...
+
+    async def refs_by_id(
+        self, owner_id: ObjectId, tag_ids: list[ObjectId]
+    ) -> dict[ObjectId, TagDoc]:
+        """Owned tag docs keyed by id, for embedding on link responses."""
+        unique = list({i: None for i in tag_ids})
+        docs = await self._tags.find_by_ids_and_owner(unique, owner_id)
+        return {doc.id: doc for doc in docs}
+
+    async def assert_owned(self, owner_id: ObjectId, tag_ids: list[ObjectId]) -> None:
```
Zingzy: **Should-fix:** `assert_owned` and `ids_for_names` have zero callers (verified by grep; the only `assert_owned` hits in the repo are the custom-domain service's). The identical ownership check is rewritten inline at `url_service.py:1446` and again at `bulk_url_service.py:605`, and the two copies already disagree on the error `field` (`tag_ids` vs `add`). `TagRepository.count_by_owner` is also dead: `create` loads up to 500 full docs to count them. Either route the three services through these methods or delete them; three copies of "are these tags yours" is exactly the drift a service layer exists to prevent.

services/tag_service.py
```diff
@@ -0,0 +1,152 @@
...
+            await self._tags.update_by_id_and_owner(tag_id, owner_id, set_ops)
+        except DuplicateKeyError:
+            raise ConflictError(f"you already have a tag named '{name}'") from None
+        log.info(
+            "tag_updated",
+            user_id=str(owner_id),
+            tag_id=str(tag_id),
+            fields=list(set_ops),
+        )
+        return existing.model_copy(update=set_ops)
```
Zingzy, line 88: **Should-fix:** stale `updated_at` on the wire. The repository stamps `updated_at` into its own copy of the ops dict, so this `model_copy` never sees it: `PATCH /tags/{id}` returns `updated_at: null` on a tag's first rename while Mongo already holds the timestamp, and the next `GET /tags` contradicts the PATCH response. Add `updated_at` to the copy (or have the repo return the stamped ops).

shared/tag_icons.py
```diff
@@ -0,0 +1,99 @@
+"""Curated icon keys a tag may carry (lucide names, kebab-case).
+
+The dashboard maps these to glyphs; the API rejects anything else so a
+client can never store a key the UI cannot draw.
+"""
+
+from __future__ import annotations
+
+TAG_ICONS: frozenset[str] = frozenset(
+    {
```
Zingzy, line 13: **Should-decide:** the icon contract is unreachable from the wire. `icon` is a bare `str` in the DTOs, so OpenAPI carries only a prose hint and the frontend must hand-mirror all 87 keys with nothing to catch drift; the server 422s on any key the two lists disagree on. The house precedent is the emoji catalogue, served as data (`GET /api/v1/emoji`) precisely so the picker never hardcodes it. Either an enum/`Literal` on the field (drift becomes a generated-schema diff) or a `GET /tags/icons` endpoint. The frontend PR pins its two internal copies with a test, but the cross-repo seam has no guard.

routes/api_v1/tags.py
```diff
@@ -0,0 +1,157 @@
...
+    return TagResponse.from_doc(doc, await tag_service.link_count(user.user_id, oid))
+
+
+@router.delete(
+    "/tags/{tag_id}",
+    responses=ERROR_RESPONSES,
+    operation_id="deleteTag",
+    summary="Delete a Tag",
+)
+@limiter.limit(Limits.API_AUTHED)
```
Zingzy: **Should-decide:** all four tag verbs share the generic `Limits.API_AUTHED` (60/min), including this DELETE, which fans out an `update_many` over every link the owner has. Compare `URL_BULK_DELETE` at 5/min for 100 links. Contained to the caller's own account, so not exploitable against others, but a write with account-wide fan-out sharing the read budget is off-pattern; siblings (domains) get per-verb constants.

services/bulk_url_service.py
```diff
@@ -582,6 +586,80 @@ def _log_move_conflict(self, doc: UrlV2Doc, target: str) -> None:
...
+
+        doc_ids = [doc.id for doc, _ in changed]
+        updates: list[dict] = []
+        if remove:
+            updates.append({"$pull": {"tag_ids": {"$in": remove}}})
+        if add:
+            updates.append({"$addToSet": {"tag_ids": {"$each": add}}})
+        updates[-1]["$set"] = {"updated_at": now}
+        try:
+            for update in updates:
```
Zingzy: **Nit, two small things here.** The $pull-then-$addToSet pair leaves a crash window where tags were removed but not added; retry converges and the docstring is honest about why, so this is fine as shipped, but note Mongo does allow both in ONE update via an aggregation-pipeline update (`$set` with `$filter` + `$setUnion`), which would close the window without a transaction. Separately, the `if add and self._tag_repo is not None` guard fails OPEN when the repo is unwired (foreign ids pass), same as `_resolve_tag_ids`; the stats path fails closed for the same None. Production always wires it, but the postures should agree.

schemas/dto/requests/url.py
```diff
@@ -37,6 +38,32 @@
 
 _ALNUM_ALIAS_RE = re.compile(r"[a-zA-Z0-9_-]+\Z")
 
+_TAG_IDS_FIELD_DESC = (
+    f"Ids of tags from `GET /api/v1/tags`, at most {TAGS_MAX_PER_LINK} per link. "
+    "Every id must be one of your tags (400 otherwise). On PATCH the list "
+    "replaces the stored one; `null` or `[]` clears it."
+)
+
+_OBJECT_ID_RE = re.compile(r"^[0-9a-f]{24}$")
```
Zingzy: **Nit:** second copy of the ObjectId shape-check. This regex + the dedupe loop below duplicate what `requests/bulk.py` already has (`_OBJECT_ID_RE` + `_tag_ids_are_object_ids`), and `requests/stats.py` uses `ObjectId.is_valid` as a third variant with different accept semantics (it takes uppercase hex, the regex refuses it). One helper in `shared/` ends the fork.

routes/api_v1/tags.py
```diff
@@ -0,0 +1,157 @@
...
+from schemas.dto.responses.tag import (
+    TagDeleteResponse,
+    TagListResponse,
+    TagResponse,
+)
+
+router = APIRouter(tags=["Tags"])
+
+
+def _parse_tag_id(tag_id: str) -> ObjectId:
```
Zingzy: **Nit:** line-for-line clone of `parse_url_id` in `routes/api_v1/_helpers.py`, whose own docstring says it exists so route modules stop growing privates like this. Parameterise the message and move it there.

services/account_erasure_service.py
Zingzy: Fixed in 57f1ead. The db fixture now inserts a tag, the expectation carries tags: 1, and the test asserts the tags collection is empty after erase. Ran tests/db against local Mongo: 5 passed, none skipped.

repositories/url_repository.py
Zingzy: Removed the cap. list_ids_by_owner_and_tag_ids returns every matching id now; the owner's own link count bounds it, and a clipped scope would have undercounted silently.

services/tag_service.py
Zingzy: Routed all three through TagService. UrlService and BulkUrlService call assert_owned (with the field name), the list filter and StatsService use ids_for_names, and the inline copies are gone. create() checks the cap with count_by_owner and only loads docs when it has to pick a colour. A missing tag service now fails closed on both link paths.

services/tag_service.py
Zingzy: The service stamps updated_at into the ops it hands the repo, so the returned copy carries it and PATCH agrees with the next GET. The repo no longer stamps on its own. Covered by a test.

shared/tag_icons.py
Zingzy: Went with an enum. TagIcon is the wire type on create, update and responses, so openapi.json carries the 87 keys as an enum and any divergence between the repos is a schema diff, not a runtime 422.

routes/api_v1/tags.py
Zingzy: Added TAG_WRITE (30/min) for create and update and TAG_DELETE (10/min) for delete, matching the bulk delete budget. List stays on API_AUTHED.

services/bulk_url_service.py
Zingzy: Took the pipeline route: one update_many with $filter plus $concatArrays, order preserved, no window between remove and add. The fail-open guard is gone too; both link paths fail closed like stats.

schemas/dto/requests/url.py
Zingzy: One helper now: shared.validators.normalise_object_ids, used by the url and bulk DTOs. Left url_id on stats as it was since that predates this PR.

routes/api_v1/tags.py
Zingzy: Moved to _helpers.parse_object_id(value, message=...); parse_url_id delegates to it.

Zingzy (commented):
Re-reviewed the amended head. All nine findings are fixed at their choke points, verified in the diff and with the unit suite (2913 passed locally; CI green except the still-running bot):

- Erasure db test now creates a tag and asserts `tags: 1` plus a direct post-erase count, so the cascade is exercised, not just counted.
- The stats tag scope is uncapped with the rationale in the docstring; the silent-undercount path is gone.
- Ownership goes through `TagService.assert_owned(field=...)` everywhere, `ids_for_names` and `count_by_owner` now have callers, and both unwired-service paths fail closed (`tags are not available`) matching the stats posture.
- PATCH stamps `updated_at` in the service where `model_copy` sees it; the response can no longer contradict the next GET.
- `TagIcon` as a str enum puts all 87 keys in openapi.json, so cross-repo drift is a schema diff now. Note the side effect: explicit `""`/`null` icons on create now 422 where they previously coerced to `tag`; right call for an unshipped contract, but the frontend mock still coerces `""`, one line of drift to mirror.
- `TAG_WRITE` 30/min and `TAG_DELETE` 10/min with the fan-out rationale in the docstring.
- The two-write window is gone: the `$filter`+`$concatArrays` pipeline is order-preserving and matches the per-item semantics the service computes (`$ifNull` covers pre-field docs), one atomic write per batch.
- `normalise_object_ids` in `shared/validators.py` with both DTO callers; `parse_object_id` in `_helpers` with `parse_url_id` delegating.

One leftover, fine to leave: `requests/stats.py` still validates tag ids via `ObjectId.is_valid`, which accepts uppercase hex the shared helper refuses. Filter-side leniency, harmless.

Mergeable once CI finishes.

repositories/url_repository.py
```diff
@@ -386,6 +386,69 @@ async def update_by_ids_and_owner(
...
+        except PyMongoError as exc:
+            log.error(
+                "repo_update_many_failed",
+                collection=self._collection_name,
+                error=str(exc),
+            )
+            raise
+
+    async def list_ids_by_owner_and_tags(
+        self, owner_id: ObjectId, tags: list[str], *, limit: int = 5000
```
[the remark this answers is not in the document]
Zingzy: Stale. The id resolution has no cap any more; list_ids_by_owner_and_tag_ids returns every matching id.

services/bulk_url_service.py
```diff
@@ -582,6 +583,68 @@ def _log_move_conflict(self, doc: UrlV2Doc, target: str) -> None:
...
+                changed.append((doc, next_tags))
+        if not changed:
+            return batch.report(op="tag", user_id=owner_id)
+
+        doc_ids = [doc.id for doc, _ in changed]
+        updates: list[dict] = []
+        if remove:
+            updates.append({"$pull": {"tags": {"$in": remove}}})
+        if add:
+            updates.append({"$addToSet": {"tags": {"$each": add}}})
```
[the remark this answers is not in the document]
Zingzy: Superseded. bulk_tag is now one pipeline update_many ($filter then $concatArrays), so there is no window between the remove and the add.

shared/tags.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def normalise_tag(value: Any) -> str:
+    """One tag: trim, casefold, collapse whitespace, whitelist characters.
+
+    Raises ValueError so callers inside Pydantic validators surface a 422.
+    """
+    if not isinstance(value, str):
+        raise ValueError("tags must be strings")
+    tag = _WHITESPACE_RE.sub(" ", _CONTROL_CHARS_RE.sub("", value)).strip().casefold()
```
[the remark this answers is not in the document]
Zingzy: Fixed in d344625: whitespace collapses first, then control characters are dropped, so a tab or newline becomes a space. Tests cover both.

schemas/dto/requests/url.py
```diff
@@ -116,6 +143,31 @@ class UrlFilter(RequestBase):
...
+
+    @field_validator("tag_ids", mode="before")
+    @classmethod
+    def _norm_tag_ids(cls, v: list | None) -> list | None:
+        return _normalise_object_ids(v) or None
+
+    @field_validator("tag_names", mode="before")
+    @classmethod
+    def _norm_tag_names(cls, v: list | None) -> list | None:
+        return normalise_tags(v) or None
```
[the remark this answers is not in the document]
Zingzy: Fixed in d344625: normalise_tags takes cap, the UrlFilter validator passes cap=None, and a test filters on 11 names.

services/bulk_url_service.py
```diff
@@ -582,6 +586,80 @@ def _log_move_conflict(self, doc: UrlV2Doc, target: str) -> None:
...
+                changed.append((doc, next_tags))
+        if not changed:
+            return batch.report(op="tag", user_id=owner_id)
+
+        doc_ids = [doc.id for doc, _ in changed]
+        updates: list[dict] = []
+        if remove:
+            updates.append({"$pull": {"tag_ids": {"$in": remove}}})
+        if add:
+            updates.append({"$addToSet": {"tag_ids": {"$each": add}}})
```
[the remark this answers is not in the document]
Zingzy: Leaving as is. The cap is a UX guard on how many tags a link shows, not an invariant anything depends on. Only the owner can write their links, so the race needs the same account to fire two bulk edits within milliseconds, and the worst case is a link with a few extra tags that the next edit trims. Not worth a predicate per item.

services/stats_service.py
```diff
@@ -221,6 +228,44 @@ def _build_click_query(
...
+        if names and self._tag_repo is not None:
+            named = await self._tag_repo.find_by_names_and_owner(names, owner_oid)
+            tag_ids += [t.id for t in named if t.id not in tag_ids]
+        ids: set[str] = set()
+        if tag_ids:
+            ids = {
+                str(i)
+                for i in await self._url_repo.list_ids_by_owner_and_tag_ids(
+                    owner_oid, tag_ids
+                )
```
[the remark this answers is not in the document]
Zingzy: Fixed already: the 5,000 cap is gone and the scope carries every matching link id.

services/tag_service.py
```diff
@@ -0,0 +1,154 @@
...
+
+    async def delete(self, owner_id: ObjectId, tag_id: ObjectId) -> int:
+        """Delete the tag and strip it from every link. Returns links touched."""
+        existing = await self._tags.find_by_id_and_owner(tag_id, owner_id)
+        if existing is None:
+            raise NotFoundError("Tag not found")
+        links_updated = await self._urls.pull_tag_id_by_owner(owner_id, tag_id)
+        await self._tags.delete_by_id_and_owner(tag_id, owner_id)
```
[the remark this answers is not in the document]
Zingzy: Tolerated by design. Link responses join tag ids through the registry and skip ids it no longer has (see _tag_refs in responses/url.py), and the filter matches nothing for a deleted id, so a stray id left by this race is invisible and costs one array slot. Serialising writes per owner is more machinery than that is worth.

openapi.json
```diff
@@ -13584,6 +14400,50 @@
...
+              {
+                "type": "string",
+                "maxLength": 40
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Icon",
+            "description": "Icon key from the curated set (lucide names such as `rocket`, `megaphone`, `flag`). Unknown keys are rejected. Defaults to `tag`."
```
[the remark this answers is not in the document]
Zingzy: Fixed in d344625: UpdateTagRequest.icon is typed with SkipJsonSchema[None], so the schema is the bare TagIcon enum while omission still keeps the icon.

services/tag_service.py
```diff
@@ -0,0 +1,152 @@
...
+
+    async def create(
+        self,
+        owner_id: ObjectId,
+        name: str,
+        color: TagColor | None,
+        icon: str = "tag",
+    ) -> TagDoc:
+        existing = await self._tags.list_by_owner(owner_id)
+        if len(existing) >= TAGS_MAX_PER_OWNER:
```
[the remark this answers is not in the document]
Zingzy: Leaving as is. The check is count-based, only the owner can create their tags, and the unique (owner_id, name) index still refuses duplicates. A same-account race can land a tag or two over 500, which nothing depends on.

openapi.json
```diff
@@ -9996,6 +10542,58 @@
...
+            },
+            "type": "array",
+            "title": "Remove",
+            "description": "Tag ids to remove from every id. Tags a link does not carry are ignored.",
+            "examples": [
+              [
+                "665f0c2f9e7a4b1d2c3d4e60"
+              ]
+            ]
+          }
```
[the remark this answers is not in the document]
Zingzy: Fixed in d344625: add carries maxItems 10 in the schema and the model docstring states that add or remove must name a tag and no tag may be in both. The either-or rule has no JSON Schema shape, so it stays prose.

openapi.json
```diff
@@ -13584,6 +14489,48 @@
...
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/TagIcon"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "description": "Icon key from the curated set (lucide names). Defaults to `tag`."
+          }
```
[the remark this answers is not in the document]
Zingzy: Fixed in d344625: null is out of the UpdateTagRequest.icon schema; the field is the TagIcon enum, omitted keeps the current icon.

schemas/dto/requests/url.py
```diff
@@ -116,6 +124,31 @@ class UrlFilter(RequestBase):
...
+
+    @field_validator("tag_ids", mode="before")
+    @classmethod
+    def _norm_tag_ids(cls, v: list | None) -> list | None:
+        return normalise_object_ids(v) or None
+
+    @field_validator("tag_names", mode="before")
+    @classmethod
+    def _norm_tag_names(cls, v: list | None) -> list | None:
+        return normalise_tags(v) or None
```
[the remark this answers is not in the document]
Zingzy: Fixed in d344625, same change as the other thread on this line: the filter passes cap=None.

services/bulk_url_service.py
```diff
@@ -582,6 +586,103 @@ def _log_move_conflict(self, doc: UrlV2Doc, target: str) -> None:
...
+        if not changed:
+            return batch.report(op="tag", user_id=owner_id)
+
+        doc_ids = [doc.id for doc, _ in changed]
+        # Aggregation-pipeline update: keep what is not removed, append what
+        # is added and not already kept. Order-preserving, one write.
+        kept = {
+            "$filter": {
+                "input": {"$ifNull": ["$tag_ids", []]},
+                "cond": {"$not": {"$in": ["$$this", remove]}},
```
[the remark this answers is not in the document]
Zingzy: Not a bug. MongoDB accepts $not with a single expression operand in aggregation, and the smoke run exercised this exact pipeline against Mongo 8. Re-checked today with a scratch collection: update_many matched and modified the document and tag_ids came back as the kept ids plus the added one, in order.

PR #356, opened 2026-09-04
fix(links): normalise scheduling datetimes before comparing

Zingzy:
## What

Two small fixes to the link scheduling work that landed in #349.

- The start-before-expiry check compared an aware request datetime with the
  naive UTC datetime Mongo returns. A PATCH that changed only `starts_at` or
  only `expire_after` on a link that already had the other one set raised a
  TypeError and answered 500. Both sides are normalised to aware UTC now.
- The cache projection truncated a fractional `starts_at` to the second
  below, so a link could read as live up to 999 ms before its start. It now
  rounds up to the next whole second.

## Why

Both surfaced in review after the merge. The first is a real 500 on an
ordinary edit; the second is a sub-second inconsistency between the model's
derived status and the hot path.

## How proven

- New tests use naive stored values the way the real path does and assert
  a 400, not a TypeError, for both orderings.
- A test with a `.900` start asserts the cache rounds up and that the link
  is not live at the floor second.
- `uv run pytest` green, `ruff check` and `ruff format --check` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed URL cache scheduling so fractional start times do not become active before their intended second.
  - Improved handling of timestamps with and without timezone information during URL updates.
  - Requests with invalid start and expiration times now return a clear validation error instead of an unexpected server error.

- **Tests**
  - Added coverage for fractional start times and mixed timestamp formats.

### DIFF
--- a/infrastructure/cache/url_cache.py
+++ b/infrastructure/cache/url_cache.py
@@ -14,7 +14,7 @@ from pydantic import BaseModel, ConfigDict, Field
 
 from infrastructure.crypto import verify_password as verify_password_hash
 from infrastructure.logging import get_logger
-from shared.datetime_utils import to_unix_timestamp
+from shared.datetime_utils import to_unix_timestamp, to_unix_timestamp_ceil
 
 if TYPE_CHECKING:
     from schemas.models.url import UrlV2Doc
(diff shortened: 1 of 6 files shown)


infrastructure/cache/url_cache.py
```diff
@@ -7,21 +7,28 @@
...
-from shared.datetime_utils import to_unix_timestamp
+from shared.datetime_utils import as_aware_utc, to_unix_timestamp
 
 if TYPE_CHECKING:
     from schemas.models.url import UrlV2Doc
 
 log = get_logger(__name__)
 
 
+def _start_second(dt: datetime | None) -> int | None:
```
Zingzy: is this the right place for this function, I mean in this file? and also are we sure there is no other shared util for this thing?

Zingzy: Right on both counts. There was no shared rounding-up helper, so I added to_unix_timestamp_ceil next to to_unix_timestamp in shared/datetime_utils.py with the same naive-is-UTC rule, and the cache projection now calls that. The private helper in url_cache.py is gone. Amended into the same commit, with parametrised tests beside the other datetime helpers.

